PR #199, opened 2022-11-28
fix: update title after navigation

### LINKED ISSUE #197: Title stuck after navigation, bug report from @cheeaun@mastodon.social
From Home, go to Notifications, go back to Home, the document.title stuck at Notifications

LittleSound:
### fix #197


LittleSound:
> at a glance, this looks like an issue with keepalive, right? yes, I think it would make sense to fix in Nuxt/unhead. and if @LittleSound is happy that would be very kind ❤️

I created a PR in `unjs/unhead` that fixes this issue: https://github.com/unjs/unhead/pull/92

PR #231, opened 2022-11-29
fix: autoload new items when at top of timeline

danielroe:
this renders a golden bar at the top of the timeline (no layout shift when it loads) and it then auto-loads new items when at the top/observed


ushuz:
A potential fix for #750

PR #456, opened 2022-12-18
feat: expand mastodon links

### LINKED ISSUE #387: Expand mastodon links
When PreviewCard is detected to be a mastodon link (via powers of regex testing the author or provider or title?) we can expand the preview card in a manner similar to Twitter

From previous discussion:
1. don't implement a quote button
2. aim to have expanded mastodon links

ayoayco:
Fixes #387 

Very initial work. Some plans:
1. parse the card title to extract username, then display bigger and colored as primary
2. generate Elk links to the author's profile and post
3. apply formatting?
4. refactor to have a reusable component shared for Github and Mastodon links?
5. Remove link in the toot content

Screenshot 2022-12-18 at 10 40 46 PM

For debugging a good long quoted post: https://fosstodon.org/@shawnhooper/109547498121012073

As shown on Toot! app
image


cooperaj:
Just tested this and it appears to not render custom emoticons - which I guess is a limitation of how these are rendered but I don't know how it's possible to solve.

ayoayco:
Closing this as it has been a long while. The holdback is due to Mastodon also expressing intent to have quoted posts from their side, and personally I wanted to see how that is implemented before proceeding. If the implementation is on the server API level, then we wouldn't have to do much other than use the API. I think at this point, better close my PR, and let others explore if this will be implemented

PR #800, opened 2023-01-05
feat: add user theming support

### LINKED ISSUE #761: Custom user theming support
- defining custom primary color & background color
- other UI colors if possible? the ones defined as custom properties could be overridable
- different sets of themes for different users
- shareable links / clickables for theme sharing

ryanccn:
Closes #761

- [x] Selectable themes
- [x] Persist themes
- [x] Custom color codes (_incomplete_)
- [ ] Per-user color theme support


antfu:
I am on it

antfu:
I am not sure if we should make all of them themeable. Prefer to have only the theme color customizable, and provide a few more builtin themes for a different contrast.

zeroby0:
Yes! More built-in themes is a great idea! Custom themes is nice, but I'd definitely give up on it after trying out a few colours and finding I can't really coordinate colours.

Maybe a few built-in themes with an option to make a custom theme? Being able to tweak the built-in themes and saving them as a new custom theme would also be nice.

ryanccn:
Yep, I think we should probably provide a moderately sized selection of preset themes and give more options to "fork" these themes, akin to how [Vivaldi](https://vivaldi.com/) implements their custom themes.

PR #863, opened 2023-01-07
feat(ui): move alt text under the media preview

baptiste0928:
Move the alt text under the media preview, instead of in the top-right corner. This make the alt text vertically aligned with the image.

[Before](https://elk.zone/mstdn.social/@elk@webtoo.ls/109642445862735271):
image

After:
image


edimitchel:
Also, I think when I work on that, we shouldn't ellipse the text and make the text be wrapped.

patak-dev:
Merging, for now, the change of place as I think it is an improvement, but let's rework this later as the style doesn't fit the UI.

PR #926, opened 2023-01-09
fix: defer loading text and server links until hydration

danielroe:
Rather than rendering en-US text in the HTML (and hard-coding `mas.to` links in the local server section) we wait until the nuxt app has hydrated to show these. This does result in a larger flash of text on initial load for English speakers but might be a more consistent approach.

In addition, we now prerender all the non-server HTML (e.g. `/notifications`), except for `/settings/*` which is rendered with the default `index.html` due to complexity. This is all a shell anyway so there's only a benefit to doing so.

This also fixes an issue where on first load the `mas.to` explore/local/federated links are shown instead of the user's local server (and persist).

Thoughts welcome.

### DIFF
--- a/components/common/CommonRouteTabs.vue
+++ b/components/common/CommonRouteTabs.vue
@@ -43,7 +43,7 @@ useCommands(() => command
         exact-active-class="children:(text-secondary !border-primary !op100 !text-base)"
         @click="!preventScrollTop && $scrollToTop()"
       >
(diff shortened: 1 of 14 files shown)


sxzz:
I removed defer loading text via cafc995d. Because I refactored the settings part and made sure that the languages on both client and server are the same. See #1013

PR #948, opened 2023-01-10
feat: totally hide strict filters

ayo-run:
As discussed on the #ux channel, there is more demand to totally hide filtered posts when the FilterAction is 'hide'

This PR will
1. remove displaying "Removed by filter: phrase' from the StatusCard component
2. implement `context` for the paginator
3. implement filtering logic in the `reorderedTimeline` function
4. add tests of filtering scenarios for `reorderedTimeline` function

Changes 2-4 are needed for when the feature `virtualScroller` is true, because it adds an empty `DynamicScroller`

Also, some initial styling fix for warning filters if it is a DM:

| before | after |
| --- | --- |
|  Screenshot 2023-01-21 at 9 29 16 PM | Screenshot 2023-01-21 at 9 45 44 PM |

Handle contexts:

- [x] Home and lists
- [x] Notifications
- [x] Public timelines
- [x] Conversations
- [x] Profiles


components/notification/NotificationPaginator.vue
```diff
@@ -112,6 +112,19 @@ function groupItems(items: mastodon.v1.Notification[]): NotificationSlot[] {
...
+    (items, index) => (
+      {
+        status: items.status,
+        index,
+      }))
+    .filter(item => !!item.status?.filtered?.find(filter => filter.filter.filterAction === 'hide' && filter.filter.context.includes('notifications')))
+    .map(item => item.index)
+
+  return items.filter((_, index) => !filteredIndices.includes(index))
+}
```
patak-dev, line 126: Could this be a direct filter with the negated condition you have there? Something like
```
function removeFiltered(items: mastodon.v1.Notification[]): mastodon.v1.Notification[] {
  return items.filter(item => 
    !item.status?.filtered?.find(filter =>
      filter.filter.filterAction === 'hide' && filter.filter.context.includes('notifications')
    )
  )
}
```

ayoayco: @patak-dev yeah, that's better. Done!

patak-dev (approved):

PR #956, opened 2023-01-10
fix(i18n): fr - use `apprecier` instead of `aimer` + add misisng trads

edimitchel:


KaKi87:
Why ?
Isn't "aimer" more relevant than "apprécier" or "favori" in the semantic field of social media ?

edimitchel:
We are discussing for making some changes about this.
I did the change to be homogeneous and avoid `Mettre en favoris` or `Aimer` which sounds odd.

KaKi87:
For that specific button, `J'aime` could be used :)

PR #980, opened 2023-01-12
feat: update UI of interface settings

wheatjs:
Updated interface for the font selector and a preview post in the interface section

image


pages/settings/interface/index.vue
```diff
@@ -1,9 +1,36 @@
...
 const { t } = useI18n()
 
 useHeadFixed({
   title: () => `${t('settings.interface.label')} | ${t('nav.settings')}`,
 })
+
+const loremStatus = computed(() => {
+  if (currentUser.value) {
+    const loremStatus: mastodon.v1.Status = {
+      account: currentUser.value!.account,
```
sxzz, line 12: maybe we should also have a fake user when there's no user signed in.

pages/settings/interface/index.vue
```diff
@@ -14,6 +41,10 @@ useHeadFixed({
       </div>
     </template>
     <div p6 flex="~ col gap6">
+      <div v-if="loremStatus" border="~ base rounded" bg-active>
+        <StatusCard :status="loremStatus" />
```
sxzz, line 45: action buttons should be disabled or hidden.

components/settings/SettingsFontSize.vue
```diff
@@ -4,12 +4,61 @@ import type { FontSize } from '~/types'
 
 const sizes = ['xs', 'sm', 'md', 'lg', 'xl'] as FontSize[]
 const fontSize = useFontSizeRef()
+
+const setFontSize = (e: any) => {
+  if ('valueAsNumber' in e.target)
+    fontSize.value = sizes[e.target.valueAsNumber]
+}
```
sxzz, line 11:
```suggestion
const setFontSize = (e: Event) => {
  if (e.target && 'valueAsNumber' in e.target)
    fontSize.value = sizes[e.target.valueAsNumber as number]
}
```

pages/settings/interface/index.vue
wheatjs: Agreed that would be better, should we use the Elk account details as the fake user?

pages/settings/interface/index.vue
wheatjs: Perhaps we could add a disabled property to the status card that would disable interactions with it?

sxzz: Sure!

pages/settings/interface/index.vue
sxzz: Yes!

edimitchel:
What about don't let the slider component be affected by the font size change until the mouse is release?
Because it makes very difficult to change the value properly.
Also, I would suggest to put the slider on top of the setting page to don't be disturbed

wheatjs:
> What about don't let the slider component be affected by the font size change until the mouse is release?
> Because it makes very difficult to change the value properly.
> Also, I would suggest to put the slider on top of the setting page to don't be disturbed 

I agree with the first point. As for the second I'm not sure about putting it above the preview as it may look weird and cause a strange break in content.

wheatjs:
Going to close as I don't think brings much of an improvement to the UX of the app

PR #982, opened 2023-01-12
feat: remove link if matches preview url

### LINKED ISSUE #981: Feature: remove link from end of status content if it matches PreviewCard.url
[Experimental branch](https://github.com/elk-zone/elk/tree/feature/remove-link-if-matches-preview-URL) Checking the status content if the last child is a link and matches the PreviewCard.url

ayo-run:
Resolves #981 

1. we check if the preview card URL matches the shared link href attribute
2. we don't need to check for multiple links, because the last link will never match the preview URL in this scenario. Mastodon only generates a preview card object for the first link
3. we append the cleaned link to the preview card "provider name" so people will still find this info --> meaning, we don't actually remove anything, but instead move it into the preview card when the link is beside it

## Screenshots

| before | after |
| --- | --- |
|  Screenshot 2023-01-29 at 10 34 19 AM | Screenshot 2023-01-29 at 10 35 26 AM |

*older screenshots (https:// not yet removed)*
Screenshot 2023-01-29 at 12 41 39 AM
Screenshot 2023-01-29 at 12 42 27 AM


ushuz:
What's blocking this?

ayoayco:
> What's blocking this?

I just haven’t the time to pick it up again, planning to do this weekend. But if someone wants to, they can continue this

ayoayco:
Just need to handle when cleanSharedLink is false, should not append to the preview card providerName

composables/content-parse.ts
```diff
@@ -127,7 +129,29 @@ export function parseMastodonHTML(
...
+  const node = parse(html) as Node
+
+  if (cleanSharedLink) {
+    // filter out invisible spans
+    const filteredNode = node.children.filter((child: Node) => !!child.children)
+    const matchedIndex = lastChildLinkMatchesPreviewUrl(filteredNode, cleanSharedLink)
+
+    if (matchedIndex)
+      filteredNode[filteredNode.length - 1].children.splice(matchedIndex, 1)
+  }
```
antfu, line 141: Could we use a transform here?

components/status/StatusPreviewCardNormal.vue
```diff
@@ -19,7 +21,17 @@ const isSquare = $computed(() => (
...
   || Number(props.card.height || 0) < ogImageWidth / 2
 ))
-const providerName = $computed(() => props.card.providerName ? props.card.providerName : new URL(props.card.url).hostname)
+const providerName = $computed(() => {
+  let finalProviderName = new URL(props.card.url).hostname
+
+  if (props.card.providerName) {
+    finalProviderName = props.card.providerName
+    if (props.cleanSharedLink && finalProviderName !== props.cleanSharedLink)
+      finalProviderName = `${props.card.providerName} · ${props.cleanSharedLink.replace('https://', '')}`
```
antfu, line 30:
```suggestion
      finalProviderName = `${props.card.providerName} · ${props.cleanSharedLink.replace(/^https?:\/\//, '')}`
```

composables/content-parse.ts
ayoayco: I tried using a transform function for this but found it unnecessary because it goes through each child node. Whereas what I am going for only needs to check the very last child

patak-dev: I agree that a transform may not be needed. I think we could merge and iterate. A possibility is to avoid removing the link here but instead add some markup so we can later hide it or show it in the UI.

patak-dev:
@ayoayco should this link be removed, no? https://deploy-preview-982--elk-zone.netlify.app/universeodon.com/@antfu@webtoo.ls/109944121349743535

ayoayco:
I need to check this again, looks like something changed and links at the end are no longer cleaned up

Shinigami92:
@ayoayco can we put this under a user/experimental flag that can be toggled on or off in https://main.elk.zone/settings/preferences ?

ayoayco:
@Shinigami92 hey shini, yes I think this would be best behind a setting. However I have not worked on this in a while; I'm not sure how big the work is to make it work.

Shinigami92:
> @Shinigami92 hey shini, yes I think this would be best behind a setting. However I have not worked on this in a while; I'm not sure how big the work is to make it work.

So first you should rebase the PR to resolve all current conflicts
Then it should be really easy and straight forward, because we already have the infrastructure for adding new experimental flags
Just orientate yourself on existing ones
We e.g. have the possibility to toggle on/off counters next to the like or repost icons for posts

ayo-run:
Closing as I do not see interest for this recently

PR #992, opened 2023-01-12
feat: add additional profile icons

cooperaj:
Included some icons for GPG/PGP fields on the profile. Also a generic photos one.

I've also added a keyoxide one - which is the same as GPG essentially. Should I also add a Keybase? and if so should I use their own icons by preference?


composables/masto/icons.ts
```diff
@@ -11,17 +11,22 @@ export const accountFieldIcons: Record<string, string> = Object.fromEntries(Obje
   Facebook: 'i-ri:facebook-line',
   GitHub: 'i-ri:github-line',
   GitLab: 'i-ri:gitlab-line',
+  Gpg: 'i-ri:key-2-line',
```
sxzz, line 14: I think `GPG` (all uppercase) will be better

cooperaj: I did think of that but they're all lowercased later anyway

danielroe (approved):

danielroe:
feel free to add keybase, and yes if we can match their iconography to the minimal line-icon style

Issue #917, opened 2023-01-09
Posting code is bugged


cyberalien:
Example of post text (link instead of code because GitHub converts Markdown): https://gist.github.com/cyberalien/3bf1871bc10236acb5c09c394d8d5422

When posting that message (copy/paste from Gist), editor bugs out, treating whole message as code block:

![posting-bug](https://user-images.githubusercontent.com/822287/211390191-46e9b4fb-7ff5-497c-9a77-cb1a3509975c.png)

This is result of that post in Mastodon and Elk (I'll open a different bug report for incorrectly rendering it):

![posting-result-mastodon](https://user-images.githubusercontent.com/822287/211391046-56fd3b7b-1235-495a-bbf0-5a1b2a2f443c.png)

![post-bug-result](https://user-images.githubusercontent.com/822287/211391073-2a60f3e7-613b-4679-af67-4c3189a8050b.png)

cyberalien:
> This should now be resolved. Let me know if not and we can reopen.

Still bugged. I think you closed wrong issue. Other issues were about rendering that code, which was also bugged. This issue is about editor.

danielroe:
I was able to successfully post that code 👀 Are you still not able to do so?

cyberalien:
No. Tried it on local from latest `main` branch, tried it on `elk.zone`. In both cases editor shows exactly the same bugged sample as in first screenshot.

Result is a bit different, now rendered correctly for what editor generated, but still bugged because editor generates wrong code by adding ``` before and after pasted message.

cyberalien:
Result demo: https://elk.zone/mas.to/@cyberalien/109675703583251205

cyberalien:
Did that, still bugged.

But something weird just happened. I tried recording demo with a clean browser (purged stuff, logged in, started demo) and this happened:
https://user-images.githubusercontent.com/822287/212043051-30059d35-9294-4a65-a057-dfec7ed232fb.mov

I'm not sure what's going on here. After posting message, Elk went in infinite loop. Restarted vite, reloaded page, purged all site data and cache, as soon as I log in to that account, it freezes again, just like at the end of this video.

However, it was enough to show bug. In your demo when you pasted code, editor rendered it correctly. In my demo, as you can see from 0:05, it pasted text incorrectly.

cyberalien:
Figured it out. It wasn't infinite loop, it was an uncaught exception that prevented rendering.

[...]

PR #994, opened 2023-01-12
fix: catch exceptions in HTML parser

cyberalien:
Fixes Elk freezing when rendering post that fails to render.

Demo of bug in video in this post: https://github.com/elk-zone/elk/issues/917#issuecomment-1380132738

This doesn't actually fix linked issue, but at least prevents Elk from freezing, so Elk can continue functioning and renders bad content as an empty string. It also logs error to console.


danielroe (approved):

011295292:
Long long

PR #1006, opened 2023-01-12
perf: mock semver import

danielroe:
this cuts 30kB of unused `semver` from the client-side bundle


sxzz:
I'll refactor this part in masto.js

danielroe:
What do you mean? I think we'll still need to mock `semver` to prevent it from being bundled.

sxzz:
Never mind. Let's get this merged first!

Issue #1004, opened 2023-01-12
Cannot log into certain instances (tested blorbo.social, fanglitch.space, retro.pizza)


biasto-bias:
### Pre-Checks

- [X] Check that there isn't [already an issue](https://github.com/elk-zone/elk/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/elk-zone/elk/discussions) or join our [Discord Chat Server](https://chat.elk.zone).
- [X] Providing a screenshot or video to reproduce the issue or show visually what was meant.
- [ ] I am willing to provide a PR.

### Describe the bug

Once I managed to open the login prompt, I tried typing in my instance (blorbo.social). It came up with several other instances ending in ".social", but not my instance.

I tested some popular instances (mas.to, c.im, mastodon.social) and they came up fine. Then I tested retro.pizza, aethy.com, and fanglitch.space, and only aethy.com worked.

I've attached a screenshot below of what came up when I typed in blorbo.social.

Thank you!

### Reproduction video or screenshot

<img width="456" alt="image" src="https://user-images.githubusercontent.com/118267931/212069813-9c4d5e05-33ef-40fe-95e3-f243d9edd265.png">

### Additional Context

Browser: Firefox on desktop.

danielroe:
you don't have to select from the dropdown - that's just for ease of use. What happens if you type your instance in full and then click sign in?

tom-sherman:
I can see how the design is confusing. It requires the user to click outside of the input or press enter to dismiss the dropdown list.

I think a super simple fix here is to hide the dropdown when the user enters a valid URL.

biasto-bias:
Thank you so much! I tried clicking outside the dropdown box to make it disappear so I could get to the sign-in button, and it worked fine. I feel pretty silly - sorry about that!

PR #1009, opened 2023-01-12
feat: hide server autocomplete when you enter a valid URL

### LINKED ISSUE #1005: When entering the name of a server that's not on the list pressing enter changes the name to the first fuzzy search result
It should not do this imo.

tom-sherman:
There is confusing UX right now as seen by #1004

closes #1005

This PR attempts to improve the UX by hiding the dropdown (and revealing the signin button) when you enter a valid URL into the server box.


tom-sherman:
Hmmm this doesn't actually work because `https://foo` is a valid URL, I guess I need to find a different heuristic.

tom-sherman:
I've improved the heuristic slightly in 092007e73008e29165234f646b20344d1cda0e5a, but now it hides the dropdown too early.

https://user-images.githubusercontent.com/9257001/212090281-cc053312-c3df-444b-a5a0-7118ed23c6a8.mov

In the above video, the autocomplete shouldn't hide when I've typed a valid URL that matches one in the list exactly.

tom-sherman:
Ok I improved the heuristic to fix that edge case, I think this is good enough now!

components/user/UserSignIn.vue
```diff
@@ -47,14 +47,6 @@ async function oauth() {
   }
 }
 
-async function handleInput() {
```
tom-sherman, line 50: I had to move the `handleInput` definition below `filteredServers` to fix an eslint "used before defined" error, I think this was a false positive tho.

PR #1013, opened 2023-01-12
feat(settings): respect settings from server

sxzz:


composables/users.ts
```diff
@@ -271,13 +271,7 @@ export function checkLogin() {
...
  */
 export function useUserLocalStorage<T extends object>(key: string, initial: () => T) {
-  // @ts-expect-error bind value to the function
-  const storages = useUserLocalStorage._ = useUserLocalStorage._ || new Map<string, Ref<Record<string, any>>>()
-
-  if (!storages.has(key))
-    storages.set(key, useLocalStorage(key, {}, { deep: true }))
-  const all = storages.get(key) as Ref<Record<string, T>>
-
+  const all = useLocalStorage<Record<string, T>>(key, {}, { deep: true })
```
sxzz, line 274: Caused Vue error here. Improve that later

antfu:
What's the reason for removing the `._` map? It will cause every component to create a `useStorage` ref, could be less optimal

sxzz:
@antfu It will cause some Vue warnings if I reverted my changes. I have no idea why did it happen.
image

Maybe we should resolve circular references first.

PR #1023, opened 2023-01-12
docs: fix elk screenshot on mobile

userquin:
Current:

elk-fixed-height-screenshot

Fix:

elk-fixed-width-screenshot


sxzz:
image

but it'll be stretched on desktop devices

userquin:
we should create a new one square

danielroe:
What if you set both height and width?

userquin:
upps, I'll try it, thx Daniel

PR #1025, opened 2023-01-12
feat: add security headers (with nuxt-security)

jviide:
This pull request adds the [nuxt-security](https://nuxt-security.vercel.app/) module to the setup, and using it to add several recommended security headers like [Strict-Transport-Security](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Strict-Transport-Security) and [Referrer-Policy](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Referrer-Policy).

Personally the greatest interest lies in the [Content-Security-Policy](https://nuxt-security.vercel.app/middlewares/headers#content-security-policy) (CSP) header that can act as a good last line of defense against script injections and such. Unfortunately I couldn't find a current way to run Nuxt v3 without allowing `'unsafe-inline'` in the [script-src](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/script-src) and [style-src](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/style-src) directives. In the ideal case hashes of inline scripts and styles (or per-request random nonces added to those inline elements) could be collected to get rid of the `'unsafe-inline'`s, realizing the full benefits of CSP. Maybe in the future!

Despite this I believe this could be a good starting point: the other headers and CSP directives still have their benefits.

In addition to the modified CSP headers, the following nuxt-security defaults have been disabled:
 * [Cross-Origin-Embedder-Policy](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cross-Origin-Embedder-Policy) headers are disabled to allow cross-origin embeds.
 * Disabled rate limiter ~~& request size limiter~~, as I don't have enough information for possible proper values for them 🙂


Baroshem:
Hey @jviide 

I am glad that you liked the module enough to recommend implementing it here!

Baroshem:
@jviide 

Sounds good! Keep in mind that the rate limiter here is using memory-cache to store the IP's so for big application handling many many requests this may not be efficient. 

At this point maybe it would be reasonable to disable this middleware and implement the same funtionality as a part of the infrastructure, not application itself by using a tool like Fail2Ban for example.

The built in rate limiter is aimed for smaller projects with less traffic. 

What we can also do here is that we could try to implement another source of storage for the rate limiter so that the values are not stored in memory but somewhere else (maybe Redis?).

Let me know what you think about it 🙂

Baroshem:
@jviide 

Agree. If you will have some spare time, could you do a little research and create an issue in my repo for potential module improvements? I am always open for any feedback on what can be done better 😃 

I would really appreciate it!

danielroe (approved):

osmaa:
I'm probably an idiot, but `pnpm run dev` after this commit on Mac OS will fail due to all requests after the first to http://localhost:5314/ being upgraded to https. Disabled nuxt-security and localhost dev started to work again.

osmaa:
I haven't. The problem reproduces on a fresh clone of the repo and with both node 18 and 19 from Homebrew.

osmaa:
hmm, apparently relatedly, also elk.zone is presenting me with security errors I haven't seen from previous versions. Clearly not the same issue as I have with dev, but also clearly to do with the new directives from nuxt-security... Browser: Safari 16.2 (17614.3.7.1.7, 17614)

image

osmaa:
yes it does! thank you. As for the CORS error, that indeed is unrelated and instead caused by https://github.com/elk-zone/elk/commit/2e7979817a539496925bb0b897fdee1165dfd133 which tries to fetch /nodeinfo/2.0, a GoToSocial-specific API that doesn't exist in Mastodon.

PR #1034, opened 2023-01-12
feat: collapse mentions

antfu:
Initial try. I personally would prefer to have an even less conspicuous design, like a smaller font size and gray text etc.

---

Before: 

image

After:

image

### DIFF
--- a/components/content/ContentMentionGroup.vue
+++ b/components/content/ContentMentionGroup.vue
@@ -1,5 +1,5 @@
 <template>
-  <p flex="~ gap-1" items-center text-sm class="zen-hide">
+  <p flex="~ gap-1" items-center text-sm class="zen-none">
     <span i-ri-arrow-right-line ml--1 text-secondary-light /><slot />
   </p>
 </template>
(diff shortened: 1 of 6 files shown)


antfu:
Later for the editor, I am thinking of having a separate line out of the editor for the mentions.

ksesong:
How often are people reading the mentions? How often does the detailed listing matter? The vertical line that connects the avatars, which is in my opinion not visible enough, suggests in a way the individuals that are part of the conversation.

If we somehow managed to put it aside, and remove the boilerplate, I feel this would help readability, and appear more as a chat message in a long conversation than a fax cover sheet.

One idea would be to remove all ending mentions, and have the information shown aside, like on the following mockup (before/after) as a counter on the top right corner of a post "@3". This could also be an inline list of avatars, if we want to be more descriptive.

Screenshot 2023-01-14 at 21 07 20

This would allow multiple interactions. We could highlight the counter if the current user were to be mentioned. Clicking on the "@3" could be a toggle, to show the mentions in their raw form, or a tooltip that would show the mentions in a vertical list, like a chat member list, or an inline list of avatars.

PR #1044, opened 2023-01-13
fix: shiki fallback language

webfansplz:
https://user-images.githubusercontent.com/22515951/212225502-5a68371a-3a5d-4d8c-b7d5-5378760f2b1c.mov

`Shiki` doesn't seem to provide highlighting for the text [language](https://github.com/shikijs/shiki/blob/main/docs/languages.md), we might be use markdown instead.

### DIFF
--- a/composables/shiki.ts
+++ b/composables/shiki.ts
@@ -32,10 +32,11 @@ export function useHightlighter(lang: Lang) {
       .then(() => {
         registeredLang.value.set(lang, true)
       })
-      .catch((e) => {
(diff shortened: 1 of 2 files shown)


sxzz (approved):

antfu:
I guess instead, we should try catch with the language loading and fallback to the plain one. As the language id could be arbitrary from users' input.

sxzz (approved):

sxzz (approved):

PR #1051, opened 2023-01-13
feat(i18n): Update French translations

KaKi87:


locales/fr-FR.json
```diff
@@ -120,6 +120,7 @@
     "title": "Elk est en mode Aperçu !"
   },
   "language": {
+    "none": "Aucun",
```
edimitchel, line 123: I saw that this translation was missing for all languages and isn't use anywhere

locales/fr-FR.json
```diff
@@ -161,7 +162,7 @@
     "blocked_domains": "Domaines bloqués",
     "blocked_users": "Utilisateur·ice·s bloqué·e·s",
     "bookmarks": "Marque-pages",
-    "built_at": "Dernier build {0}",
+    "built_at": "Dernière build {0}",
```
edimitchel, line 165: For me build is masculine.
Le build and not la build.

edimitchel (changes requested):

locales/fr-FR.json
KaKi87: Alright, I therefore removed it from all language files.

locales/fr-FR.json
KaKi87: Although I can't find an explicit definition anywhere, there are more occurences of "la build" than "le build" on the web in the context of computing.

I'd argue that since "build" can be translated with "compilation" or "construction" which are both feminine words then it sounds more natural for "build" to be feminine as well.

Thinking about it, I instead replaced "build" with "compilation" in order to simplify that decision.

edimitchel (approved):

locales/fr-FR.json
```diff
@@ -266,10 +273,7 @@
     "notifications": {
       "label": "Notifications",
       "notifications": {
-        "label": "Paramètres des notifications",
-        "notifications": {
-          "label": "sqds"
```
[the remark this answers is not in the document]
KaKi87: Thanks !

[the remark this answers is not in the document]
KaKi87: I'm using [nested key diffing](https://media.discordapp.net/attachments/1044887051155292203/1060255664460529664/screenshot-text-compare.com-2023.01.04-18_56_40.png) between EN & FR files to spot all missing or unexpected properties.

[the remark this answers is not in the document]
edimitchel: Interesting! I use the VSCode extension i18n for making sure no translation is missed but yes for extra translation your solution is good!

PR #1052, opened 2023-01-13
feat: add docker support

Candinya:
So it would be easier to deploy

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:lts-alpine AS base
+FROM docker.io/library/node:lts-alpine AS base
 
 # Prepare work directory
 WORKDIR /elk
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,7 +16,7 @@ RUN apk add git --no-cache
 COPY . ./
 
 # Build
-RUN pnpm i
+RUN pnpm i --frozen-lockfile
 RUN pnpm build
 
 FROM base AS runner
--- a/Dockerfile
+++ b/Dockerfile
@@ -29,4 +29,10 @@ EXPOSE 5314/tcp
 
 ENV PORT=5314
 
+# Specify container only environment variables ( can be overwritten by runtime env )
+ENV NUXT_STORAGE_FS_BASE='/elk/data'
+
+# Persistent storage data
+VOLUME [ "/elk/data" ]
+
 CMD ["node", ".output/server/index.mjs"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,16 +3,17 @@ FROM node:lts-alpine AS base
 # Prepare work directory
 WORKDIR /elk
 
-# Prepare pnpm ( refer to https://pnpm.io/installation#on-alpine-linux )
-RUN wget -qO /bin/pnpm "https://github.com/pnpm/pnpm/releases/latest/download/pnpm-linuxstatic-x64" && chmod +x /bin/pnpm
-
 FROM base AS builder
 
-COPY . ./
+# Prepare pnpm ( refer to https://pnpm.io/installation#on-alpine-linux )
+RUN wget -qO /bin/pnpm "https://github.com/pnpm/pnpm/releases/latest/download/pnpm-linuxstatic-x64" && chmod +x /bin/pnpm
 
 # Prepare deps
 RUN apk update
 RUN apk add git --no-cache
+
+# Copy all files
+COPY . ./
 
 # Build
 RUN pnpm i
@@ -27,4 +28,5 @@ COPY --from=builder /elk/.output ./.output
 EXPOSE 5314/tcp
 
 ENV PORT=5314
+
 CMD ["node", ".output/server/index.mjs"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,16 +3,16 @@ FROM node:lts-alpine AS base
 # Prepare work directory
 WORKDIR /elk
 
-# Prepare deps
-RUN apk update
-RUN apk add git --no-cache
-
 # Prepare pnpm ( refer to https://pnpm.io/installation#on-alpine-linux )
 RUN wget -qO /bin/pnpm "https://github.com/pnpm/pnpm/releases/latest/download/pnpm-linuxstatic-x64" && chmod +x /bin/pnpm
 
 FROM base AS builder
 
 COPY . ./
+
+# Prepare deps
+RUN apk update
+RUN apk add git --no-cache
 
 # Build
 RUN pnpm i
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,34 @@
+FROM node:lts-alpine AS base
+
+# Prepare work directory
+WORKDIR /elk
+
+# Prepare deps
+RUN apk update
+RUN apk add git --no-cache
+
+# Prepare pnpm ( refer to https://pnpm.io/installation#on-alpine-linux )
+RUN wget -qO /bin/pnpm "https://github.com/pnpm/pnpm/releases/latest/download/pnpm-linuxstatic-x64" && chmod +x /bin/pnpm
+
+FROM base AS builder
+
+COPY . ./
+
+# Build
+RUN pnpm i
+RUN pnpm build
+
+# Remove dev deps ( skip postinstall script after removing dev deps or it will fail )
+RUN pnpm i -P --ignore-scripts
+
+FROM base AS runner
+
+ENV NODE_ENV=production
+
+COPY ./package.json ./
+COPY --from=builder /elk/node_modules ./node_modules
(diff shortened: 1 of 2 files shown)


sxzz (approved):
Wow looks great!

sxzz:
I'm not very familiar with docker. Is there a way to pass a `.env` file?

Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+
+# Remove dev deps ( skip postinstall script after removing dev deps or it will fail )
+RUN pnpm i -P --ignore-scripts
+
+FROM base AS runner
+
+ENV NODE_ENV=production
+
+COPY ./package.json ./
+COPY --from=builder /elk/node_modules ./node_modules
```
danielroe, line 29: you shouldn't need to include anything but `.output`

```suggestion
```

Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+COPY --from=builder /elk/.output ./.output
+
+EXPOSE 5314/tcp
+
+CMD ["pnpm", "start"]
```
danielroe, line 34:
```suggestion
CMD ["node", ".output/server/index.mjs"]
```

Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+FROM base AS runner
+
+ENV NODE_ENV=production
+
+COPY ./package.json ./
+COPY --from=builder /elk/node_modules ./node_modules
+COPY --from=builder /elk/.output ./.output
+
+EXPOSE 5314/tcp
+
```
danielroe, line 33:
```suggestion
EXPOSE 5314/tcp

ENV PORT=5314

```

Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+RUN apk update
+RUN apk add git --no-cache
+
+# Build
+RUN pnpm i
+RUN pnpm build
+
+# Remove dev deps ( skip postinstall script after removing dev deps or it will fail )
+RUN pnpm i -P --ignore-scripts
+
```
danielroe, line 23:
```suggestion
```

Candinya:
> I'm not very familiar with docker. Is there a way to pass a `.env` file?

If for production running, yes, use volume map like `./.env:/elk/.env:ro` should be OK

And the .env file (if any) seems to be copied automatically during build process by  `COPY . ./` , so it should be fine

sxzz:
Also we could publish docker image to github packages using github actions when releasing a new version.

cyberalien:
> And the .env file (if any) seems to be copied automatically during build process by `COPY . ./` , so it should be fine

Is that safe though? If docker image is published, it will include private information.

Also `COPY . ./` would copy `node_modules` tree, which is then overridden by `pnpm I`.

Also, `pnpm i` installs new dependencies, which might introduce bugs. Should command be `pnpm i --frozen-lockfile` instead, installing last known working versions of all dependencies, which is pnpm version of `npm ci`?

Dockerfile
```diff
@@ -0,0 +1,32 @@
...
+ENV NODE_ENV=production
+
+COPY --from=builder /elk/.output ./.output
+
+EXPOSE 5314/tcp
+
+ENV PORT=5314
+
+CMD ["node", ".output/server/index.mjs"]
```
sxzz, line 32: Should we use `ENTRYPOINT` instead? I'm not sure.

Also, why did I press `Ctrl - C` on `docker run --rm -p 5314:5314 -it elk` but nothing happened? Only `docker stop` can terminate the process.

cyberalien:
> Docker Hub is charged for the organizations.

Only for private packages. It is free for open source stuff.

Dockerfile
cyberalien: `ENTRYPOINT` cannot be overridden, but `CMD` can. It allows adding custom node flags if needed.

sxzz:
@cyberalien But I didn't see a free option [here](https://hub.docker.com/choose-plan?type=organization)
image

cyberalien:
> @cyberalien But I didn't see a free option

Weird. I have one for Iconify that has "Docker Free Team" subscription. Looks like they've changed sign up form.

Found this though: https://www.docker.com/community/open-source/application/

sxzz:
@cyberalien Yeah I think they've changed policy before.

I didn't see any particularly benefit of using Docker Hub, so I would prefer to use GH Packages directly than apply a "sponsored" Docker Hub with a very detailed form.

danielroe:
A suggestion for this PR - maybe we can move the Dockerfile to the docs rather than directly in the repo? There will also be other ways of deploying elk and it could get cluttered.

Candinya:
Just created a free docker hub org ( `elkzone` ), please give me a username to send invitation? Thanks.

sxzz:
@Candinya `sanxiaozhizi`

Can we change the organization id to `elk-zone`? (as the same as GH org)

Candinya:
> @Candinya `sanxiaozhizi`
> 
> Can we change the organization id to `elk-zone`? (as the same as GH org)

Invitation sent ;)

Sadly docker hub doesn't support dash in username, so I'm afraid we cannot use the same org name 🥲

Candinya:
> Is that safe though? If docker image is published, it will include private information.

Yes, this will have security issues, so for official repository we won't recommend including .env file (which always contain private keys etc.). And this repo ignores this, so it's OK.

But users can customize their docker image freely with their own CI settings.

> Also `COPY . ./` would copy `node_modules` tree, which is then overridden by `pnpm I`.

Yes, we need an .dockerignore file to prevent such ignored files, I'd like to add one.

> Also, `pnpm i` installs new dependencies, which might introduce bugs. Should command be `pnpm i --frozen-lockfile` instead, installing last known working versions of all dependencies, which is pnpm version of `npm ci`?

Good idea! I'll add this too.

Candinya:
> A suggestion for this PR - maybe we can move the Dockerfile to the docs rather than directly in the repo? There will also be other ways of deploying elk and it could get cluttered.

It's a good idea! Move files to function-oriented directories can keep the repo tree clean.

But on many occations, place the `Dockerfile` in the repo root can bring benefits to simplify workflow settings (the default Dockerfile usage, also easy for docker-compose bu simply specifying `build: .` ). and is also good for others who wants to fork their own version ( they can easily find this file and modify as they want ).

If there's any updates (like decide to place Dockerfile in subdirectory (like `deploy`) or something else), don't hesitate to let me know. :D

sxzz:
@Candinya I think so. Let's leave it in the root directory first.

More opinions are welcome.

danielroe:
I would prefer a page in the docs, as we also will need `nginx` configuration, for example, and possibly `pm2`. (And we can link through to Nitro deployment information also.)

sxzz:
Hmmm, how about putting it to `/docker/Dockerfile`? I don't think it's a good idea for putting it on docs only.

Candinya:
BTW do we need to persistent fs storage?

cooperaj:
I think general best practice for Dockerfile's is to have them at the root of the project (if the project produces a single artefact).

Personally I'd prefer an officially produced image that can be relied upon, though I appreciate the administration overhead that will add, as just having it as a documented thing with no official builds means Dockerhub will be awash with unofficial version. Like mine, should that not happen 😄

danielroe:
Then let's add it at root of project _and_ generate a docker image - wdyt @elk-zone/team?

Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+
+# Prepare deps
+RUN apk update
+RUN apk add git --no-cache
+
+# Copy all files
+COPY . ./
+
+# Build
+RUN pnpm i --frozen-lockfile
```
cooperaj, line 19: It might be better to first copy the lock files and run the install. Then you copy in the remaining files and build. This improves the caching performance on any potential CI as it should not have to download all the dependencies for every run (just ones in which the lock files change)

sxzz:
Let's have this first. We can also continue to discuss it in the issue.

Candinya:
Split the pnpm install before & after file copy (let's call them `Pre` and `Post`), but with some compromise, here's the reason:

- A post-install script is specified in package.json, which requires some project files to work, and simply include those involved files in `Pre` might break the cache optimize idea if any of them changes.
- Meanwhile pnpm cannot skip one specific post-install script (like just ignore the package.json specified one, and run all others in deps), so they can only be skipped together, with --ignore-scripts option.

And I'm not sure why with --ignore-scripts option pnpm somewhat installed just 1302 packs, it's a little smaller than the ordinary deps number (like 1471)

But this still workls because the `Post` reused 1287 packages and only downloaded 165 more, which should save a lot time with cache option.

Here's the log of this command if you'd like to analyze it 😄 
[elk-build-logs-8c8e434e.log](https://github.com/elk-zone/elk/files/10430713/elk-build-logs-8c8e434e.log)

PR #1053, opened 2023-01-13
chore: reduce bundle size for Tauri target

JonasKruckenberg:
This PR:
1. Disables the PWA module for Tauri builds (since it's not needed anyway)
2. Cleans up files copied from the `public` folder that aren't needed in an app setting either


danielroe (approved):

userquin:
@danielroe without PWA you don't have service worker, and so, you don't have push notifications on Elk Tauri Apps

JonasKruckenberg:
> @danielroe without PWA you don't have service worker, and so, you don't have push notifications on Elk Tauri Apps

true actually, I'll have a look at the notification code to see if we can re-enable them without the other pwa cruft

PR #1084, opened 2023-01-13
feat: use animate like icon

Innei:
https://user-images.githubusercontent.com/41265413/212352598-a80cee8f-39db-4eed-841b-666df18613d6.mov


sxzz:
I think it's too similar with Twitter...

Innei:
But now UI looks like Twitter too, right?

edimitchel:
I see some improvement gap there
heart

edimitchel:
About @sxzz feedback, why not using a different style: for example, doing a double beating with a solid wave leaving the heart in the background?

boehs:
The icon should not be solid when unliked

PR #1096, opened 2023-01-13
feat: first pass at resolving server host from well-known

cooperaj:
My server lives on a different domain to my user handle. This is a supported setup via mastodon and means my handle `@pieceofthepie@n8e.dev` can be automatically found and used when Mastodon itself lives at `social.n8e.dev`. This uses the magic of WebFinger for the most part but there are other items living under the `.well-known` url that let you, for instance, discover what server lives at what domain.

In this case it's possible to use `https://n8e.dev/.well-known/nodeinfo`, which has been setup to 301 redirect to the Mastodon install to find the host. 

This PR adds an additional resolution check when attempting to sign in that checks this endpoint to see if Elk needs to go somewhere else before attempting the OAuth flow. 

It is naive. For instance, it currently caches the host for future use as the resolved one, which may confuse users. Re-logging in. It is hopefully resilient to failure and will fallback to whatever the user has entered. It may not be the best way to do it etc etc.


cooperaj:
An alternative implementation might use the full user handle and the `.well-known/webfinger` endpoint to discover the users server. 

In theory that means any webfingered user could find a server with ease. 

I have one (`https://acpr.dev/.well-known/webfinger?resource=acct:adam@acpr.dev`) that means you can search for my email on most ActivityPub services and end up at my Mastodon profile.

Issue #256, opened 2022-11-30
support more html in statuses


danielroe:
glitch and maybe other activitypub servers support more features in statuses. See, for example:

https://elk.zone/toot.cat/@woozle/109433157214171127

PR #1102, opened 2023-01-13
feat: support additional html elements

danielroe:
This only has effect if you are on a glitch (or other) instance that supports more html elements. (You can see this when logged out at http://localhost:5314/toot.cat/@woozle/109433157214171127.)

(I chose not to render `h1` larger as I thought it would mess with the aesthetic more than ol/ul.

CleanShot 2023-01-13 at 21 56 31@2x

resolves https://github.com/elk-zone/elk/issues/256

### DIFF
--- a/styles/global.css
+++ b/styles/global.css
@@ -141,7 +141,24 @@ em-emoji-picker {
   pre code {
     --at-apply: bg-transparent px0 py0 rounded-none leading-1.6em;
   }
-
+  ol {
+    --at-apply: list-decimal my-3 pl-6 ml-2;
+  }
+  ul {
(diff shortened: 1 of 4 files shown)


antfu (approved):

styles/global.css
```diff
@@ -141,7 +141,24 @@ em-emoji-picker {
...
+  ul {
+    --at-apply: list-disc my-3 pl-6 ml-2;
+  }
+  li {
+    --at-apply: mt-1 mb-1;
+    &:empty {
+      --at-apply: hidden;
+    }
+  }
+  'ol > li' {
```
antfu, line 156: @danielroe is this intentional? I got this
image

PR #1123, opened 2023-01-14
fix: should hide tooltip when click outside

### LINKED ISSUE #1062: Toolip not hide when click elsewhere.
### Pre-Checks

### Describe the bug

We click emoji icon, tooltip shows, that's good. But when click somewhere else, tooltip did't hide.

https://user-images.githubusercontent.com/49969959/212295135-d8b0eff3-7388-4c1a-8b16-d19e6d4d361b.mp4

### Additional Context

_No response_

webfansplz:
Closes #1062

### DIFF
--- a/components/common/CommonTooltip.vue
+++ b/components/common/CommonTooltip.vue
@@ -9,7 +9,7 @@ defineProps<{
 <template>
   <VTooltip
     v-bind="$attrs"
-    :auto-hide="true"
+    auto-hide
   >
     <slot />
     <template #popper>
--- a/components/common/CommonTooltip.vue
+++ b/components/common/CommonTooltip.vue
@@ -4,11 +4,20 @@ import type { Popper as VTooltipType } from 'floating-vue/dist'
 defineProps<{
   content?: string
 } & Partial<typeof VTooltipType>>()
+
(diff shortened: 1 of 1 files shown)


edimitchel:
This should normally not be needed.
I think we need some @Akryum advice or confirmation it's a bug?

webfansplz:
> By default tooltips close on mouse out or blur, you shouldn't have to add this logic. Also if you really need it, just put `autoHide`

Thanks.

webfansplz:
> This should normally not be needed. I think we need some @Akryum advice or confirmation it's a bug?

For #1062, it might make sense to add `autoHide`.

antfu (approved):

edimitchel (approved):

components/common/CommonTooltip.vue
```diff
@@ -9,6 +9,7 @@ defineProps<{
 <template>
   <VTooltip
     v-bind="$attrs"
+    :auto-hide="true"
```
edimitchel, line 12:
```suggestion
    auto-hide
```

edimitchel (changes requested):

PR #1126, opened 2023-01-14
feat: add confirm for account handle

webfansplz:


components/account/AccountMoreButton.vue
```diff
@@ -9,9 +9,16 @@ let relationship = $(useRelationship(account))
 
 const isSelf = $(useSelfAccount(() => account))
 
+const { t } = useI18n()
 const masto = useMasto()
-const toggleMute = async () => {
-  // TODO: Add confirmation
+
+const isConfirmed = async (title: string) => {
+  return await openConfirmDialog(t('common.confirm_dialog.title', [title])) === 'confirm'
```
sxzz, line 16:
```suggestion
const isConfirmed = (title: string) => {
  return openConfirmDialog(t('common.confirm_dialog.title', [title])) === 'confirm'
```

webfansplz: We might need to it, because it's return a promise. See: https://github.com/elk-zone/elk/blob/896681d5b919f76877237f796ec204069be73981/composables/dialog.ts#L40

sxzz: Oh I didn't see `=== 'confirm'`😹

sxzz (approved):

PR #1127, opened 2023-01-14
feat: enhancements to profile account notifications

### LINKED ISSUE #1124: Account Notifications - Add tooltip to icon and include the action in the "More" menu.

Building on the changes I made in #985 I would like to create a new pull request to make further enhancements to the Notification feature on the Account profile view to provide consistency to the patterns found in the Elk codebase and give users more clarification on what action would occur when clicking the icon. 

**What are "Notifications"?**
For an end user, enabling notifications on a followed account means that they will receive a notification anytime this account posts something new.

### Suggested solution

I would like to make the following changes specifically:

* Adding a tooltip to the notification icon on the AccountHeader view
* Use the term "Notifications" throughout code and ui for consistency.
* Add the notifications to the AccountMoreButton component in addition to the icon.

The reason for adding the action in multiple places is that if someone is unsure what "Notifications" means via the tooltip, the More command of "Enable notifications for {user.x}" should make that clear.  If you toggle the bell the menu item is affected, and vice versa.

Here's a demonstration of these new updates:

Demonstration of suggested Notifications updates

### Alternative

If the notification bell is going to be included in a release, the CommonTooltip should be added if nothing else.  The "More" dropdown item could be removed if duplicating the action is considered against the UI/UX design patterns in Elk.

I'm also open to suggestions on the placement of the bell icon and the icon used for it.  Elk uses a LOT of icons so here are a few alternatives I thought might work that do not conflict with others already in use. You can also peruse available icon options on [Remix](https://remixicon.com/). 

Alternative notification icon options

### Additional context

This is my first opensource contribution, other than opening issues and having someone else do it, so any words of advice or recommendations are greatly appreciated. I want to make sure this request benefits the community and makes me proud of my contribution. Thanks!

jeffsikes:
This PR adds a tooltip to the Account Header's new notify button.  The tooltip text changes as the notify button is toggled from on to off.

Additionally, the icon has changed from a "jingle bell" to a "notification bell" icon which is consistent with the concept of "notifications" in Elk (the same type of icon is used for global notifications as well).  This was a discussion that occurred on Jan 15th in the UI Discord channel.

Fixes #1124

Huge thanks to @rshigg for working on these changes with me, as we both worked on similar parts of the code in separate PRs! 🙌🏼 

The videos below show all 4 states of the notification button:

1. inactive
2. inactive hover
3. active hover
4. active

## Before
https://user-images.githubusercontent.com/6627582/212523186-26ff5732-5e20-4355-a03f-5c2b227a50df.mov

## After
https://user-images.githubusercontent.com/6627582/212569147-b4140825-542e-48c6-981e-c90456a7b582.mov

## Tooltip Screenshots
Screenshot 2023-01-15 at 3 52 34 PM
Screenshot 2023-01-15 at 3 52 41 PM


components/account/AccountMoreButton.vue
```diff
@@ -62,6 +68,22 @@ const toggleReblogs = async () => {
 
       <template v-if="currentUser">
         <template v-if="!isSelf">
+          <template v-if="relationship!.following">
```
patak-dev, line 71: I'm getting an error when testing here, the relationship could be `undefined`
```suggestion
          <template v-if="relationship?.following">
```

jeffsikes: @patak-dev The menu item was removed here, so the code is gone now.

components/account/AccountHeader.vue
```diff
@@ -100,17 +104,18 @@ const isNotifiedOnPost = $computed(() => !!relationship?.notifying)
...
-            :class="isNotifiedOnPost ? 'text-primary border-primary hover:bg-red/20 hover:text-red hover:border-red' : 'border-base hover:text-primary'"
-            @click="toggleNotifications"
-          >
-            <span v-if="isNotifiedOnPost" i-ri:bell-fill block text-current />
-            <span v-else i-ri-bell-line block text-current />
-          </button>
+          <CommonTooltip v-if="!isSelf && relationship?.following" :content="getNotificationIconTitle()">
+            <button
+              :aria-pressed="isNotifiedOnPost"
+              :aria-label="getNotificationIconTitle()"
```
rshigg, line 110: The reason that I kept the label on the button as just `account.notify_on_post` is because of the way `aria-pressed` works. Screenreaders announce the label and then if the button is "pressed" it will let the user know, which is exactly what we want here. If we change the labels on click then it will no longer do that.

Here's a great article that explains this behavior: https://joshcollinsworth.com/blog/accessible-toggle-buttons#understanding-aria-pressed

locales/en-US.json
```diff
@@ -28,7 +28,8 @@
     "muted_users": "Muted users",
     "muting": "Muted",
     "mutuals": "Mutuals",
-    "notify_on_post": "Notify me when {username} posts",
+    "notifications_on_post_disable": "Stop notifying me when {0} posts",
+    "notifications_on_post_enable": "Notify me when {0} posts",
```
rshigg, line 32: I do prefer going with the more explicit `{username}` interpolation but it does seem that list interpolation is the current pattern. I'll leave it up to @patak-dev to decide.

components/account/AccountHeader.vue
jeffsikes: Thanks for the explanation, makes perfect sense. I can change it back to notify_on_post.

What if the aria-pressed stayed as account.notify_on_post, but the CommonTooltip continued to display the altered text based on current state of the button?

rshigg: The tooltip definitely should display both labels, sorry forgot to mention that. Also just to clarify, I think your naming for the i18n keys is great lets keep them as is.

jeffsikes: Speaker of naming conventions, to keep all three versions grouped, what do you think about renaming "notify_on_post" as "notifications_on_post_aria"?

rshigg: I was under the impression that the wording for `notifications_on_post_enable` and `notify_on_post` were the exact same. In which case we can get rid of `notify_on_post` entirely.

jeffsikes: As I pull up the previous code I see that. Ok, leaving it at the two then. Less is more!

jeffsikes: @rshigg this change was made, however if we decide to add back in the {username} interpolation based on @patak-dev decision, it will change again.

rshigg (approved):
Great work on this!

components/account/AccountHeader.vue
jeffsikes: We did add the {username} interpolation back in as discussed.

locales/en-US.json
jeffsikes: {username} interpolation was used to make it clear what value is expected to be passed to the localization file.

PR #1129, opened 2023-01-14
feat: refetch home page when click elk logo

Mini-ghost:
In the current version, clicking the Elk Logo in the navigation bar will reload the entire page.

https://user-images.githubusercontent.com/39984251/212484144-4ff97168-12cf-4da5-99b4-5f1bcff26d6b.mov

According #86, this is to expect that when the Logo is clicked, page will jump to top and refresh timeline to latest.

I try to refactor it with nuxt hook to avoid refetching the whole page.

https://user-images.githubusercontent.com/39984251/212484627-8c1c3bb6-9397-4a36-8723-c5175bfb8bd8.mov


components/nav/NavTitle.vue
```diff
@@ -2,6 +2,14 @@
 const { env } = useBuildInfo()
 const router = useRouter()
 const back = ref<any>('')
+
+const nuxtApp = useNuxtApp()
+
+const onClickLogo = () => {
+  // @ts-expect-error missing types
+  nuxtApp.hooks.callHook('elk-logo:click')
```
edimitchel, line 10: You could probably add an ambiant type to this no?

edimitchel: Also, I'm not pretty sure using a nuxt hook is the best move..

edimitchel: That's said: you can extends `RuntimeNuxtHooks` interface to add the new one and remove ts ignore ;)

Mini-ghost: I think adding an ambient type is better than using ts-expect-error , but I'm not sure how to extend `RuntimeNuxtHooks`

I did this but it doesn't seem to work.

```ts
// shims.d.ts

declare module '@nuxt/schema' {
  interface RuntimeNuxtHooks {
    'elk-logo:click': () => void
  }
}

export {}
```

Mini-ghost: I had a similar functional requirement in a company project, currently I solved it with hookable, but that was a vue project not nuxt.

I'm also not sure if the nuxt core team expects hookable to be used in this way, but so far it seems to work.

edimitchel: Ask this in the #nuxt discord channel, Daniel is usually answer quickly.

components/nav/NavTitle.vue
```diff
@@ -11,15 +19,14 @@ router.afterEach(() => {
...
-  <!-- Use external to force refresh page and jump to top of timeline -->
   <div flex justify-between>
     <NuxtLink
       flex items-end gap-4
       py2 px-5
       text-2xl
       focus-visible:ring="2 current"
-      to="/"
-      external
+      to="/home"
```
antfu, line 28: Probably better to be `/`, as the direction will be different because on whether it's logged. Or assigning the value conditionally.

antfu (commented):
I think the full page reload should be fixed, but not sure if it makes sense to serve the logo as a refresh button instead of "Go to home page"?

frankgx97:
Would it make more sense to use the home button rather than the elk logo? Mobile devices cannot benefit from this feature since there's no elk logo at all, also considering currently there's no way to refresh timeline in PWA other than killing and relaunching the app.

Mini-ghost:
> I think the full page reload should be fixed, but not sure if it makes sense to serve the logo as a refresh button instead of "Go to home page"?

In this PR, the link behavior is "Go to homepage" if the user is not on the homepage, otherwise "Refetch timeline data" if the user is on the homepage, is this PR title inaccurate ?

patak-dev:
I now get a blank timeline if I open the preview, scroll down a bit, and then press the logo.

components/nav/NavTitle.vue
Mini-ghost: It is resolved at c78b4312a51a2d13b576e0e70fee9c290c7cb2af

edimitchel (approved):

Mini-ghost:
> I now get a blank timeline if I open the preview, scroll down a bit, and then press the logo.

This solved!

middleware/auth.ts
```diff
@@ -4,6 +4,11 @@ export default defineNuxtRouteMiddleware((to) => {
   if (to.path === '/signin/callback')
     return
 
+  if (isHydrated.value) {
+    if (to.path === '/')
+      return navigateTo('/home')
+  }
+
```
patak-dev, line 11: Is this needed? Maybe the logo could go directly to /home?

Mini-ghost: I think the logo can go directly to `/home`, currently just follow the original design

PR #1138, opened 2023-01-14
fix: Improve styling and add accessible text to notify button

rshigg:
This PR makes it more obvious when notifications for a user are enabled. It also adds an `aria-label` to the notify button since currently it does not have any accessible text.

Huge shoutout to @jeffsikes for helping to get this PR ready!

The UI in the videos below is going through the following states in order:
1. inactive
2. inactive hover
3. active hover
4. active

## Before

https://user-images.githubusercontent.com/90143161/212495938-114b2c7a-a296-48b7-9586-34d095cb5775.mov

## After

https://user-images.githubusercontent.com/90143161/212495800-4f84ac9f-33b6-428c-80fd-cd0244f0a1ca.mov

### Active state screenshots
localhost_5314_universeodon com_@michae1webb@toot wales (3)
localhost_5314_universeodon com_@michae1webb@toot wales (2)


jeffsikes:
Hey @rshigg just this morning I also opened a pull request #1127 with similar changes.  Great minds think alike!  I utilized the CommonTooltip and the locales for the hover text to match similar functionality elsewhere.  I also added the same action into the More menu, for similar reasons - making it more accessible for text only viewers.  One thing mine doesn't cover directly tho is that aria label, tho I did think about it. I didn't see that anywhere else in the code patterns so didn't add it in.

rshigg:
@jeffsikes lets discuss this over at discord? I've been posting about this change in the ui channel.

components/account/AccountHeader.vue
```diff
@@ -100,8 +100,8 @@ const isSelf = $(useSelfAccount(() => account))
         <div absolute top-18 inset-ie-0 flex gap-2 items-center>
           <AccountMoreButton :account="account" :command="command" />
 
-          <button v-if="!isSelf && relationship?.following" flex gap-1 items-center w-full rounded op75 hover="op100 text-pink" group @click="toggleNotify()">
-            <div rounded-full p2 group-hover="bg-pink/10">
+          <button v-if="!isSelf && relationship?.following" flex gap-1 items-center w-full rounded op75 hover="op100" group :aria-label="`${relationship?.notifying ? 'Stop notifying' : 'Notify'} me when @${account.username} posts`" @click="toggleNotify()">
+            <div rounded-full p2 border-1 :class="relationship?.notifying ? 'text-primary border-primary' : 'border-base'" :group-hover="relationship?.notifying ? 'text-red bg-red/10 border-red' : 'text-primary'">
```
userquin, line 104: can you replace all div inside button with span? it is not valid html, add  `block` attribute

components/account/AccountHeader.vue
```diff
@@ -100,8 +100,8 @@ const isSelf = $(useSelfAccount(() => account))
         <div absolute top-18 inset-ie-0 flex gap-2 items-center>
           <AccountMoreButton :account="account" :command="command" />
 
-          <button v-if="!isSelf && relationship?.following" flex gap-1 items-center w-full rounded op75 hover="op100 text-pink" group @click="toggleNotify()">
-            <div rounded-full p2 group-hover="bg-pink/10">
+          <button v-if="!isSelf && relationship?.following" flex gap-1 items-center w-full rounded op75 hover="op100" group :aria-label="`${relationship?.notifying ? 'Stop notifying' : 'Notify'} me when @${account.username} posts`" @click="toggleNotify()">
```
edimitchel, line 103: Please think i18n.
Create new translation key :)
VSCode ext exist to ease that :) (i18n by Intlify)

components/account/AccountHeader.vue
rshigg: great catch! the changes from this PR are going to be added to #1127, so we'll make that change there 😄

rshigg:
@userquin @edimitchel addressed both of your comments

edimitchel (approved):

patak-dev:
Nice work! Lets merge this one first, and then we can continue in #1127 
Leaving a note here that later on it would be great to avoid the red hover effect after switching on the bell (this is something we should do everywhere though)

Issue #939, opened 2023-01-10
Switching a user doesn't refreshes boosts and favorites for posts


patak-dev:
Reproduction: Go to the detailed view of a post you favorited, and change the user to one that didn't favorited.

patak-dev:
Fixed by https://github.com/elk-zone/elk/pull/1139

PR #1139, opened 2023-01-14
fix: use user id in cache key

### LINKED ISSUE #1134: Mixed profile data between accounts
this is marked as fixed but I seem to still be having this issue. 

the account @twilight@autonomous.zone is logged but being looked at from another account that is logged in, @wakest@mastodon.social.  its showing a post I have no recollection of ever seeing and that was not boosted by that account ever. also the server that is logged into: mastodon.social: has a history of posts already on display from this account
image

_Originally posted by @wakest in https://github.com/elk-zone/elk/issues/366#issuecomment-1382876749_

danielroe:
we should not cache statuses by server alone - cache might include whether a user has liked the status which is user-specific. Saw a report in Discord to this effect.

Related issue but not resolved by this: https://github.com/elk-zone/elk/issues/1134.

### DIFF
--- a/composables/cache.ts
+++ b/composables/cache.ts
@@ -19,7 +19,8 @@ function removeCached(key: string) {
 
 export function fetchStatus(id: string, force = false): Promise<mastodon.v1.Status> {
   const server = currentServer.value
-  const key = `${server}:status:${id}`
+  const userId = currentUser.value?.account.id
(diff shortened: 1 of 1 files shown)


patak-dev (approved):
Thanks 🙏🏼 

This also fixes:
- https://github.com/elk-zone/elk/issues/939

Issue #153, opened 2022-11-27
Implement social images and expansions for elk.zone links


patak-dev:
Expansion when sharing links on Twitter, Slack, Discord, etc

holtwick:
That would be great. For better context I attached a screenshot of the same Mastodon post shared by using the link from the browser's location field via Apple Messages app. The one from mastodon.social provides a preview.

![20230113-103848-capture-holtwick@2x](https://user-images.githubusercontent.com/49557/212287966-2039a7d7-b41c-4254-9732-a086b03ade23.png)

PR #1144, opened 2023-01-14
feat: implement social previews by redirecting to origin server

danielroe:
Rather than return a response containing any information about a status/user/server, we can simply redirect to the correct URL of the mastodon instance involved. This instance's `robots.txt`, privacy policy, etc. will then govern the response. However, we still will get the benefit of whatever meta tags that instance provides.

**Discord**

CleanShot 2023-01-14 at 23 42 14@2x

**Twitter**

CleanShot 2023-01-14 at 23 44 40@2x

**Facebook**

CleanShot 2023-01-14 at 23 43 12@2x

**Elk**

CleanShot 2023-01-14 at 23 43 46@2x

**Mastodon UI**

CleanShot 2023-01-14 at 23 44 10@2x

resolves https://github.com/elk-zone/elk/issues/153

### DIFF
--- a/plugins/social.server.ts
+++ b/plugins/social.server.ts
@@ -1,0 +1,16 @@
+import { sendRedirect } from 'h3'
+
+export default defineNuxtPlugin(async (nuxtApp) => {
+  const route = useRoute()
+  if (!route.params.server)
+    return
+
+  const req = nuxtApp.ssrContext!.event.node.req
+  const userAgent = req.headers['user-agent']!
+  if (!userAgent)
+    return
+
+  const isOpenGraphCrawler = /twitterbot|discordbot|facebookexternalhit|googlebot|msnbot|baidu|ahrefsbot|duckduckgo/i.test(userAgent)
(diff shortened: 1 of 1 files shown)


patak-dev (approved):
This is so awesome

plugins/social.server.ts
```diff
@@ -0,0 +1,16 @@
+import { sendRedirect } from 'h3'
+
+export default defineNuxtPlugin(async (nuxtApp) => {
+  const route = useRoute()
+  if (!route.params.server)
+    return
+
+  const req = nuxtApp.ssrContext!.event.node.req
```
Atinux, line 8: Why not using useRequestEvent() here @danielroe ?
https://nuxt.com/docs/api/composables/use-request-event

Atinux: Or even https://nuxt.com/docs/api/composables/use-request-headers ?

danielroe: much better!

PR #1145, opened 2023-01-15
fix: ignore `sensitive` flag when there is no summary or media

### LINKED ISSUE #1072: Sensitive flag should be ignored if summary == null
### Pre-Checks

### Describe the bug

When viewing a post where the `sensitive` flag is `true`, the "show more" prompt should not be shown where `summary == null`.

Here is a post where the `sensitive` flag is `true`, but `summary` is null.

`curl -v -H "Accept: application/activity+json" https://hackers.town/@thegibson/109681977901723288 | jq .`

Elk renders it like this, the "Show more" toggle does nothing:

Screenshot 2023-01-13 at 13 16 03

Mastodon renders it like this (ignoring the `sensitive` flag):

Screenshot 2023-01-13 at 13 14 22

_No response_

### Additional Context

_No response_

danielroe:
Mastodon uses sensitive flag + empty spoiler text to indicate sensitive media.

resolves #1072

### DIFF
--- a/components/status/StatusContent.vue
+++ b/components/status/StatusContent.vue
@@ -26,7 +26,7 @@ const isFiltered = $computed(() => filterPhrase && (context && context !== 'deta
     }"
   >
     <StatusBody v-if="!isFiltered && status.sensitive && !status.spoilerText" :status="status" :with-action="!isDetails" :class="isDetails ? 'text-xl' : ''" />
-    <StatusSpoiler :enabled="status.sensitive || isFiltered" :filter="isFiltered">
+    <StatusSpoiler :enabled="(status.sensitive && (!!status.spoilerText || !!status.mediaAttachments?.length)) || isFiltered" :filter="isFiltered">
       <template v-if="filterPhrase" #spoiler>
         <p>{{ `${$t('status.filter_hidden_phrase')}: ${filterPhrase}` }}</p>
       </template>


rshigg:
Are `spoilerText` and `mediaAttachments` the only two things that can appear inside the hidden area?

danielroe:
You can also have a poll, but I don't believe it's possible to mark a poll as sensitive.

rshigg:
Sounds good. I think since the condition has gotten quite long it should be stored in a variable to help with readability and to make it easier to modify in the future if needed.

TheEnbyperor:
I don’t think this should be merged. The activitypub spec allows a post to be marked as sensitive without a summary nor media attachments.

Mastodon may never itself generate a post like this, but a federated posts may have this. 

Elk should instead have a way of displaying posts as sensitive (i.e. hiding behind some filler CW) regardless of a CW or media existing.

rshigg:
That's a great point! I guess the real solution here is to update the verbiage to be less confusing when this specific edge case happens.

danielroe:
The tricky thing is that Mastodon uses this: `sensitive: true` with no spoiler text to refer _specifically_ to displaying the status but not the images. This may be a bug or failure to comply with the protocol but it is how the Mastodon web client currently behaves.

ayoayco:
Hi all! I think we can close this now as PRs #2065 and #2072 will address this, and we will have further improvement on #2068 

cc @danielroe @patak-dev

PR #1146, opened 2023-01-15
fix: fall back to votes count if voters count is not supplied

### LINKED ISSUE #1125: Poll result percentage wrong
### Pre-Checks

### Describe the bug

I looked at [this](https://front-end.social/@leaverou/109688267189424003) toot in [Elk](https://elk.zone/front-end.social/@leaverou/109688267189424003) and submitted the poll. This led to the results being shown. The percentage calculation on the graph was wrong though, including "2,400%", "0%", "400%", and "300%" (see screenshot).

image

### Additional Context

Firefox Nightly, Windows 11

danielroe:
resolves #1125


components/status/StatusPoll.vue
```diff
@@ -27,13 +27,18 @@ async function vote(e: Event) {
   }
   poll.voted = true
   poll.votesCount++
-  poll.votersCount = (poll.votersCount || 0) + 1
+
+  if (!poll.votersCount && poll.votesCount)
```
sxzz, line 31: `poll.votesCount` is always a number, not including `null`

danielroe: Is there a change you would suggest here?

antfu: Relavent to this? https://elk.zone/m.webtoo.ls/@cheeaun@mastodon.social/109688468147731159

danielroe: I think this PR is good to go. It doesn't matter if votersCount (or votesCount) is null or 0 - we fall back to zero in that case.

antfu (approved):

PR #1147, opened 2023-01-15
fix: disable selecting elements in navbar

### LINKED ISSUE #1132: UI elements should be unselectable
### Pre-Checks

- [X] I am willing to provide a PR.

### Describe the bug

Currently, UI elements like buttons and placeholders are selectable, which is weird. IMO those UI elements are expected to be unselectable.

image

### Additional Context

_No response_

sechi747:
resolved #1132


danielroe (approved):
This looks good to me - thanks!

I think we probably also need to apply `select-none` to other parts of the UI as well, like the user switcher.

Issue #123, opened 2022-11-26
Preserving state between route changes


LittleSound:
Preserve the page/component state when going back to the previous page instead of refreshing it

https://user-images.githubusercontent.com/19204772/204080323-d547ace1-c557-48b5-9700-c8eccb889d2b.mp4

PR #1163, opened 2023-01-15
fix: show status actions on DM

### LINKED ISSUE #988: don't hide status actions on DM
https://github.com/elk-zone/elk/blob/22e6b131517343d4c6571a171c914fe1890b95d3/pages/[[server]]/@[account]/[status].vue#L81

Even if it's a DM, we can still favorite and bookmark it.

webfansplz:
### Description

Closes #988 

### Additional context

 /cc @patak-dev, WDYT?

---

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Translations update
- [ ] Other

### DIFF
--- a/pages/[[server]]/@[account]/[status].vue
+++ b/pages/[[server]]/@[account]/[status].vue
@@ -78,7 +78,6 @@ onReactivated(() => {
           :status="status"
           command
           style="scroll-margin-top: 60px"
-          :actions="status.visibility !== 'direct'"
         />
         <PublishWidget
           v-if="currentUser"


patak-dev:
The idea to avoid showing these actions for DMs is that they aren't used as much as in regular posts. You don't normally bookmark a DM, you can't boost a DM, so we are only left with favorite and reply. You also don't care about favorite counts for a DM. (all these actions should be in the three dots though)

I think we should show a simplified reply and favorite action buttons directly on the message bubble itself, maybe in the bottom right of it, like over the border with position absolute. The idea is to try to keep their height small.

patak-dev:
Another option for the favorite and reply actions placement is left of the bubble below the avatar. But I'm worried about one-line messages. Maybe that is no longer an issue now that the list of mentions is on a separate line.

Or maybe on the top right next to the visibility `@` label

webfansplz:
> Another option for the favorite and reply actions placement is left of the bubble below the avatar. But I'm worried about one-line messages. Maybe that is no longer an issue now that the list of mentions is on a separate line.
> 
> Or maybe on the top right next to the visibility `@` label

How about this:


https://user-images.githubusercontent.com/22515951/212539935-169e958e-bfcc-4e2c-8f6d-8c3446da174e.mov

rshigg:
Maybe the long-term solution is to add DM swipe-events? The way messaging apps allow you to perform actions on messages when you swipe sideways.

danielroe:
I do like the idea of favouriting a message in a DM to communicate approval or appreciation - not sure how helpful it would be to reply to a previous message in a chain though. 🤔

sxzz:
@patak-dev Actually we shouldn't decide for users, and I think Elk should implement all features of the original Mastodon website. (it's a bad experience that I have to go back to the mastodon website and do it on that)

> You also don't care about favorite counts for a DM.

Users can mention multiple people (maybe hundreds) so I think we should keep it. But for single DM, we can improve that surely.

patak-dev:
@sxzz I implemented the removal of actions halfway, so if you would like, let's add it back until we have a full implementation for this idea. It was only intended to be in conversations, and most of the time you are having them with a single person where the count doesn't makes sense. But we should take all these cases into account before moving forward. You can see how much difference it makes in the new zen mode. I'll try to do this better again in the future. @webfansplz maybe you could leave this PR as a new base and create a new one with your original commit that just added them back?

PR #1165, opened 2023-01-15
fix: prevent HTML injections to code blocks

jviide:
### Description

This pull request fixes a raw HTML injection vector. A Mastodon post payload that contains HTML that looks enough like a code block can bypass HTML sanitation. This is fixed by:
 * not directly injecting unescaped HTML content in the ContentCode.vue component
 * escaping code conten in shiki.ts when the requested language is not found or has not been loaded yet

Visiting the page https://elk.zone/bob.pwniverse.io/@jviide/109692653165280776 demonstrates how a message crafted on a normal, non-attacker-controlled instance (running the [Mastodon Glitch Edition](https://glitch-soc.github.io/docs/)) can paint over the whole Elk UI and run arbitrary code in e.g. click handlers.

### Additional context

There are two points of code that must be protected. The first one is the ContentCode.vue component. If the language is not given, then the component injects the code content as-is with the v-html directive. This is fixed by adding the code content to DOM as normal text in such situations.

The other place is usually available for a much shorter time, but should be addressed anyway in `~/composables/shiki.ts`. When the shiki module & language information has not yet been bootstrapped, the `useHighlighter` function returns `undefined`. In such cases the `highlightCode` function (used by the ContentCode component) returns the content as-is, which in turn is injected as-is with v-html by the ContentCode component. This is addressed by escaping the unformatted content in `highlightCode`.

---

I added three tests for these cases into the `content-rich.test.ts` test file. To facilitate those tests I removed the mock for ContentCode component, and added a small mock for the shiki-es module that disables the CDN functionality, making testing it a bit easier. I also removed the reactivity transform use from ContentCode to get the tests work after removing the component mock.

These changes required updating the test snapshots for the said file.

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Translations update
- [ ] Other

### DIFF
--- a/components/content/ContentCode.vue
+++ b/components/content/ContentCode.vue
@@ -4,7 +4,7 @@ const props = defineProps<{
   lang?: string
 }>()
 
-const raw = computed(() => decodeURIComponent(props.code).replace(/&#39;/g, '\''))
+const raw = $computed(() => decodeURIComponent(props.code).replace(/&#39;/g, '\''))
 
 const langMap: Record<string, string> = {
   js: 'javascript',
@@ -13,7 +13,7 @@ const langMap: Record<string, string> = {
 }
 
 const highlighted = computed(() => {
-  return props.lang ? highlightCode(raw.value, (langMap[props.lang] || props.lang) as any) : raw.value
+  return props.lang ? highlightCode(raw, (langMap[props.lang] || props.lang) as any) : raw
 })
 </script>
 
--- a/components/content/ContentCode.vue
+++ b/components/content/ContentCode.vue
@@ -4,7 +4,7 @@ const props = defineProps<{
   lang?: string
 }>()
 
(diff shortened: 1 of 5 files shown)


components/content/ContentCode.vue
```diff
@@ -4,7 +4,7 @@ const props = defineProps<{
   lang?: string
 }>()
 
-const raw = $computed(() => decodeURIComponent(props.code).replace(/&#39;/g, '\''))
+const raw = computed(() => decodeURIComponent(props.code).replace(/&#39;/g, '\''))
```
patak-dev, line 7: Thanks a lot for this PR 🙏🏼 

I think we could keep the `$computed` here, did you have a reason for the change?

patak-dev: Ah, I see, @danielroe has our back there

patak-dev (approved):
Tested it a works well for me

PR #1172, opened 2023-01-15
feat(i18n): include country variants utility

userquin:
### Description

This PR includes:
- an utility to avoid adding duplicate entries for language country variants.
- add spanish country variants.

On draft, I need to add some docs.

### Additional context

The approach is the following:
- add generic one, for example `es`, with: `{ code: 'es', file: 'es.json', name: 'Español' }`
- rename `es-ES.json` to `es.json`
- exclude `es` from i18n locales and add the country variants, overriding entries from default
- add `countryLocaleVariants` with each country: if some variant requires override entries, just add overrideFile: true to the entry
- add nuxt module with `i18n:extend-messages` hook: here I load the default locale messages, and then for each country variant I do a merge

If I omit on `i18n:extend-messages` hook adding default entries there is no translation, seems messages are missing.

This way we can just add a few entries per locale: just reuired country variants messages

---

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Translations update
- [ ] Other


userquin:
Problems found with this approach:
- dev server will not refresh changes made to locale file: i18n hook not being called (HMR is there)
- the hook requires to do a merge: `en-US`, `es` and custom variant file
- where is the code for additional messages? it seems we are breaking modules per locale

Maybe it will be better to add a new script to generate locale variants: reuse the code on nuxt module asking for contry variant to regenerate all country variants (will work also on dev, but will require run another script).

Last and optimal solution, include full chain resolution in `vue-i18n`:
- `xx-YY_variant` > `xx_YY` > `xx` > `default`
- `xx-YY` > `xx` > `default`

OFC, with this option we need to include all entries on `i18n` configuration: we'll need to change here (Elk) show generic ones, maybe adding a flag or adding some logic.

userquin:
Closing this, bundle size will be huge, vue-i18n will provide merge multiple files per locale,  and so, we can reuse base locale.

I'll send a new PR when ready for spanish and docs.

Issue #1122, opened 2023-01-14
Grayscale support for digital wellbeing


ljrk0:
### Clear and concise description of the problem

Colorful images and UI interfaces encourage interaction, leading to addiction-like behaviors.

### Suggested solution

As part of Elk's digital wellbeing configurations, adding a greyscale or greyscale-optimized mode could tone down the effect of such imagery and interfaces.

### Alternative

Alternatively, some systems such as iOS support a device-wide greyscale mode.  While this is definitely nice to reduce time spent on the phone in general, this filter may work against us by also decreasing contrast of certain UI elements.  Ideally, Elk could detect (how?) that this mode is active and adjust certain UI elements if needed.

However, a system-wide greyscale mode is no replacement for a per-app setting.  Especially on the desktop I need the colors for work but would like to have my social media client to be less distracting.  Also, some systems don't have system-wide greyscale in general.

### Additional context

The now unmaintained client pinafore.social had the option to turn on a greyscale mode (theme) based in recommendations of humantech.com.

Maybe that's something Elk could implement as well, what do you think?

Finally, I just wanted to thank you for the work, I really appreciate all the effort into not "only" developing the most polished client but also caring for r2l languages, digital wellbeing and secure technology (totally hyped about tauri native!).

JonasKruckenberg:
Interesting, do you by any chance have some studies handy or other data to back this up (If it's just something you discovered about your own behaviour that's ofc totally valid too) because it's the first time I heard about this being a strategy to mitigate causes of addiction. I would really like to know more

patak-dev:
Also interested if there are studies to justify this.

We can keep talking here, closing to properly track things now that https://github.com/elk-zone/elk/pull/1177 has been merged. Later on we can create new issues to improve on it.

boehs:
Here is one such study

- https://www.tandfonline.com/doi/abs/10.1080/03623319.2020.1737461?journalCode=ussj20

> In addition, total screen time was significantly less for these individuals by an average of 37.90 min each day. The influence of grayscale did not account for improvement in anxiety or depression

And another, this one a bit broader in scope but publicly accessible https://psyarxiv.com/tjynk

ljrk0:
> Interesting, do you by any chance have some studies handy or other data to back this up (If it's just something you discovered about your own behaviour that's ofc totally valid too) because it's the first time I heard about this being a strategy to mitigate causes of addiction. I would really like to know more

[...]

PR #1177, opened 2023-01-15
feat: add grayscale mode to user preferences

rshigg:
### Description

Implements #1122

https://user-images.githubusercontent.com/90143161/212548857-34f4baa1-b4e5-4107-87f0-81ba9c0aa99f.mov

---

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Translations update
- [ ] Other


layouts/default.vue
```diff
@@ -10,10 +10,12 @@ const showUserPicker = logicAnd(
   usePreferences('experimentalUserPicker'),
   () => useUsers().value.length > 1,
 )
+
+const isGrayscale = computed(() => getPreferences(userSettings.value, 'grayscaleMode'))
```
sxzz, line 14:
```suggestion
const isGrayscale = usePreferences('grayscaleMode')
```

layouts/default.vue
```diff
@@ -10,10 +10,12 @@ const showUserPicker = logicAnd(
   usePreferences('experimentalUserPicker'),
   () => useUsers().value.length > 1,
 )
+
+const isGrayscale = usePreferences('grayscaleMode')
 </script>
 
 <template>
-  <div h-full>
+  <div h-full :class="{ 'filter-grayscale': isGrayscale, 'filter-none': !isHydrated }">
```
danielroe, line 18: What about?

```suggestion
  <div h-full :class="{ 'filter-grayscale': isHydrated && isGrayscale }">
```

patak-dev:
I thought that this mode would apply the grayscale filter only to images and avatars. But leave the rest of the UI with the primary color. I think that would be better, as more people will be able to use this wellness mode from an a11y perspective.

And it would also be interesting if when you hover or focus an avatar or image, they get to be full color so you experience grayscale while browsing but you can see the content as intended if you would like to do so without disabling this wellness option.

layouts/default.vue
rshigg: That was actually my initial implementation 😅 I'll switch back.

rshigg:
@patak-dev something like this? 

https://user-images.githubusercontent.com/90143161/212552880-e6b0707f-a05b-442c-a332-361954bf6452.mov

ljrk0:
Wow, first off, thanks to @rshigg for this quick PR!

> I thought that this mode would apply the grayscale filter only to images and avatars. But leave the rest of the UI with the primary color. I think that would be better, as more people will be able to use this wellness mode from an a11y perspective.

Can you elaborate a bit more why you think this would improve a11y?  Current implementations, both iOS grayscale mode as well as Pinafore's apply this to the whole interface.  This was also what I expected and also part of what I enjoyed, as it drove down the whole interaction with the sometimes dangerous pull of social media.  Only grayscaling content would turn down interaction on image content but I'd personally prefer to go all in here.

> 
> And it would also be interesting if when you hover or focus an avatar or image, they get to be full color so you experience grayscale while browsing but you can see the content as intended if you would like to do so without disabling this wellness option.

I *love* this idea.  I'm not sure whether hover may lead to too many distractions but it looks great in @rshigg 's demo.

patak-dev:
> > I thought that this mode would apply the grayscale filter only to images and avatars. But leave the rest of the UI with the primary color. I think that would be better, as more people will be able to use this wellness mode from an a11y perspective.
> 
> Can you elaborate a bit more on why you think this would improve a11y? Current implementations, both iOS grayscale mode as well as Pinafore's apply this to the whole interface. This was also what I expected and also part of what I enjoyed, as it drove down the whole interaction with the sometimes dangerous pull of social media. Only grayscaling content would turn down interaction on image content but I'd personally prefer to go all in here.

I don't know to be honest. This is just what I thought when I first saw your issue. My thinking was that we use primary color to help people find important actions, what is enabled, etc. Our use of primary color is not driven to engage (we even removed the bright Compose button and people are fighting back to get it again). We use it to make the interface more accessible and easier to understand. If we go full grayscale, I think we need to do a lot of work to rework our use of color and change active states to always include a shape difference (like line icons to fill icons). I think we aren't that bad in that regard.

In summary, my thought was that we wanted to tone down only the content that could drive engagement.

There is also an interesting discussion about code blocks, my ideal would be to have syntax highlighting on as this is also not to drive engagement but to help the reader see the structure.

> I _love_ this idea. I'm not sure whether hover may lead to too many distractions but it looks great in @rshigg 's demo.

I also was afraid of this, but I think it should be fine as we users can scroll out of the timeline and only interact with the post when we find something interesting. If this ends up being an issue, we could add a new toggle button on images so you need to click it to turn full color for it, or we could add a delay when hovering so you can still cross an image with the mouse without triggering the change (and only if you stay for a bit it changes, maybe this could be done with a CSS transition that is slow to go into full color but reverts quickly to avoid the need of JS state).

ljrk0:
> Another way to see it is that the contrast may suffer quite a bit in our default themes if we apply a gray filter everywhere. So if people would really like to have that experience I think it could be decoupled from this wellness feature:

Absolutely, this (I think) I touched upon in the original issue as well when it came to the "detect iOS grayscale mode" part.  A system wide grayscale mode can effectively remove UI hints.  Pinafore addressed this by adding shape differences just like you proposed.

> * The wellness preference: Grayscale content (apply to images, avatars, expansions?, polls?)
> * A grayscale theme once we have theme support that uses only gray colors but is intentional in that, so it may need to use a darker background to improve contrast, and maybe use full white for primary color so that the things we want to highlight from a functional POV are still visually different

That'd be a neat solution, I think.  Of course, one could always have like "make an additional option" but this easily leads to clutter.  Splitting the logic into "theme" and "grayscale content" makes semantically more sense.

I do like a toned down theme for the reasons stated above (drive back engagement in general, not only to attention grabbing content).  This helps me to concentrate much better.

Interesting you brought up syntax highlighting, as this I'd probably disable as well if I could.  My rule of thumb when I prepare reports is to have it disabled when in-line (usually small code examples) but on for the appendix (full code listings).

rshigg:
I have read all of your feedback and I agree that the best long term solution is to have a grayscale "theme". For now I have restricted grayscale mode to images, video, code blocks, and the status actions (reply, etc.). I'm going to update the demo video in the description.

Thank you everyone for your perspectives 🧡

patak-dev:
I think the transition works very well for avatars. I see it isn't applied to attachments. I think it would also be good for them.

One thing that calls my attention a lot with this feature enabled is the green color of the retweet. I think that it should be in normal color for this wellness option (not grayscale).

patak-dev:
@rshigg it is in the top comment from netlify: https://deploy-preview-1177--elk-zone.netlify.app/home

patak-dev:
Oh, looks like it is canceled since yesterday. @danielroe do you know what happened with the netlify deploys?

danielroe:
We have to manually initiate each one @patak-dev from the Netlify deploys panel, for security reasons.

rshigg:
Thank you Daniel!

patak-dev:
@ljrk0 let's merge this one so we can get some feedback outside this group. Please create a new issues with suggestions and let's iterate to improve this feature

ljrk0:
> @ljrk0 let's merge this one so we can get some feedback outside this group. Please create a new issues with suggestions and let's iterate to improve this feature

Sorry for being a bit late!  This looks awesome already and works really well for me.  Thanks @rshigg for the PR and @patak-dev for the helpful thoughts.  I think splitting theme & desaturating content was very much the right call.

Issue #1179, opened 2023-01-15
Custom Font


sxzz:
### Clear and concise description of the problem

customize font family on preference 

### Suggested solution

add an option `fontFamily` to preferences

### Additional context

relate #1170

antfu:
https://github.com/elk-zone/elk/pull/1182#issuecomment-1383224123

I think we could pre-select a few sets of fonts, that match styles across sans, serif, mono. That user could change the font by only clicking.

For example:

- DM Sans / DM Mono / DM Serif
- Fira Sans / Fira Code / ?
- Noto Sans / Noto Mono / Noto Serif

PR #1182, opened 2023-01-15
feat: change default font

### LINKED ISSUE #1170: Clearer font please!

`AI` looks like `Al` on elk.zone:

Screenshot 2023-01-15 at 11 23 36

### Suggested solution

New font

### Alternative

Maybe we could have a font where capital-i and lowercase-L don't look identical! :)

### Additional context

Would be super useful for visually impaired users, too.

Shinigami92:
### Description

closes #1170 

### Additional context

Good comparison status:
- https://elk.zone/mas.to/@ayo@ayco.io/109682917903570914
- https://deploy-preview-1182--elk-zone.netlify.app/universeodon.com/@ayo@ayco.io/109682917998896114

---

- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Translations update
- [x] Other


Shinigami92:
I tried installing
- https://www.npmjs.com/package/@fontsource/fira-code
- https://www.npmjs.com/package/@fontsource/firago

And then use these via unocss `presetWebFonts`, but I had no luck and didn't found good docus about how to do that

So I switch to what is currently done in the project and just downloaded the ttf files from
- https://www.fontmirror.com/fira-code
- https://github.com/bBoxType/FiraGO/tree/master/Fonts/FiraGO_TTF_1001/Roman

---

For now I just switched the default 400/regular fonts, but somehow magically also the bold font is working :thinking: or I just inspect it wrongly
So maybe we also need FiraGO for **bold** and _italic_

Shinigami92:
|Previously|Changed|
|-|-|
|image|image|

styles/global.css
```diff
@@ -2,6 +2,7 @@ html {
   font-size: var(--font-size, 15px);
   width: 100%;
   width: 100vw;
+  font-family: var(--font-family, 'FiraGO');
```
Shinigami92, line 5: @sxzz I'm not fully sure how these things works, but I think this is already a prepare for https://github.com/elk-zone/elk/issues/1179

I assume then in a follow-up PR we can dynamically set the `--font-family` variable and swap it

sxzz:
Like I mentioned before, we shouldn't make decision for users. Let user to set their browser font. 
image


And it may also break non-English characters on Windows, like Japanese, Chinese. (I'm not sure).

Shinigami92:
@sxzz oh you mean on system layer :thinking: 
I'm not sure about that, e.g. I even don't know if I can switch my font on my mobile Opera browser :thinking:, can I?

ushuz:
I agree with @sxzz that font setting should be left to users to decide, and there're existing mechanisms to do so.

> I'm not sure about that, e.g. I even don't know if I can switch my font on my mobile Opera browser 🤔, can I?

@Shinigami92 You could try changing font in your phone's system settings, if you really want. But in any case, switching Elk's default font for **ALL USERS**  is not a fix for not being able to configure fonts on certain mobile browsers for some users.

Shinigami92:
> I agree with @sxzz that font setting should be left to users to decide, and there're existing mechanisms to do so.
> 
> > I'm not sure about that, e.g. I even don't know if I can switch my font on my mobile Opera browser thinking, can I?
> 
> @Shinigami92 You could try changing font in your phone's system settings, if you really want. But in any case, switching Elk's default font for **ALL USERS** is not a fix for not being able to configure fonts on certain mobile browsers for some users.

No, you do not understand the root cause.
Right now on main `DM Sans` is used as default
https://github.com/elk-zone/elk/blob/3513f091482b2dd4ebb39a102bd0b0f0e39e5d4f/unocss.config.ts#L81-L86
https://github.com/elk-zone/elk/blob/3513f091482b2dd4ebb39a102bd0b0f0e39e5d4f/styles/global.css#L7-L34

So this PR is just changing these fonts to use `Fira` instead of `DM`
So all of the given arguments about using the system font and/or browser override does not apply as then it is already an "issue" in the released version of Elk

I'm totally okay that someone provide the possibility for what #1179 is requesting, but it doesn't prevent this PR which fixes the issue declared by #1170

ushuz:
@Shinigami92

ushuz:
> No, you do not understand the root cause.

@Shinigami92 I feel like it's the other way around.

Based on what I can see in the devtools, correct me if I'm wrong, `font-family` actually defaults to user's system ui font, not DM fonts. So supposedly, if someone is unsatisfied with the font, they can always change the font settings on their end.

image

However, your PR will change `font-family` to `Fira` for **ALL USERS** on **ALL DEVICES**, which is at a totally different level.

Shinigami92:
The Discord user `box#7898` pointed out that Noto Sans supports 806 different languages, but Fira Sans/GO currently only 328
This is a point and I will change the font to Noto Sans/Mono

Shinigami92:
> The Discord user `box#7898` pointed out that Noto Sans supports 806 different languages, but Fira Sans/GO currently only 328 This is a point and I will change the font to Noto Sans/Mono

I tried `Noto Sans`, but this is worse and has the issue that lower case `l` is not distinquisable to capital `I`

image

antfu:
I would say can we provide three sets of fonts with different styles for users to choose from in their preferences.

I honestly never have issues reading `I` and `l` since most of the time you have the context. Worth mentioning is that on many site they just look alike (Twitter, GitHub, etc. on macOS). But I do understand some might prefer to have more distinguishable fonts. I think we should follow System's default by default, and let the users to chocies the fonts we selected.

rshigg:
@antfu would the team prefer that the default UI font is the system sans fonts? We're discussing this in the ui channel on discord.

antfu:
Actually, it is the default font right now (see https://github.com/elk-zone/elk/pull/1190). And I personally prefer to keep it as-is, while making the possibility for users to config.

rshigg:
We were wondering about letting the user change it in interface settings.

danielroe:
I'd agree - default system font looks nice (IMO) and is more performant. Allowing user configuration seems like a nice idea.

Shinigami92:
Okay, I start to understand
IMO the current config setup via unocss is a bit confusing and miss-interpretable
So this PR is not the right way right now and for the other stuff I do not have the time as well as knowlegde
So I will close it for now and would be happy to review an upcoming PR that solves it the right way

rshigg:
I am in agreement that system fonts should be the default but I would like to at least have an option to use a dyslexia-friendly font. To me this seems like something we can pretty easily add to the interface settings.

PR #1183, opened 2023-01-15
fix(i18n): 登鹿 for zh-CN & zh-TW

yenche123:
### Description

Is this a pun intentionally made by the developers?

sign_in_elk

---

- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [X] Translations update
- [ ] Other

- [X] Read the [Contributing Guidelines](https://github.com/elk-zone/elk/blob/main/CONTRIBUTING.md).
- [X] Check that there isn't already a PR that solves the problem the same way to avoid creating a duplicate.
- [X] Provide related snapshots or videos.
- [X] Provide a description in this PR that addresses **what** the PR is solving, or reference the issue that it solves (e.g. `fixes #123`).

### DIFF
--- a/locales/zh-CN.json
+++ b/locales/zh-CN.json
@@ -61,7 +61,7 @@
     "reset": "重置",
     "save": "保存",
     "save_changes": "保存更改",
-    "sign_in": "登鹿",
+    "sign_in": "登录",
     "switch_account": "切换帐号",
     "vote": "投票"
   },
--- a/locales/zh-TW.json
+++ b/locales/zh-TW.json
@@ -64,7 +64,7 @@
     "reset": "重設",
     "save": "儲存",
     "save_changes": "儲存更改",
-    "sign_in": "登鹿",
+    "sign_in": "登入",
     "switch_account": "切換帳號",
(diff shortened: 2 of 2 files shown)


locales/zh-CN.json
```diff
@@ -61,7 +61,7 @@
     "reset": "重置",
     "save": "保存",
     "save_changes": "保存更改",
-    "sign_in": "登鹿",
+    "sign_in": "登录",
```
sxzz, line 64: It's by design.

---

这是故意的。因为叫做「鹿鸣」，所以改「登录」为「登鹿」

ushuz: We could leave some comments there if JSONC is supported.

antfu:
I would suggest we construct a terms table as README in the repo for each language. JSON5 is not a good option as we are using i18n Ally to write back this file (so the comment can't be persisted), and it could be hard to read in a large file. Closing this PR for now.

locales/zh-CN.json
yenche123: 好哦，很有趣

PR #1190, opened 2023-01-15
fix: Actually use "DM sans" font in the UI

Azurency:
### Additional context

---

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Translations update
- [ ] Other

---
Right now `DM sans` is referenced in the source code but never applied. This PR adds "font-sans" in the default layout.

Before (on macos) : 
CleanShot 2023-01-15 at 19 28 26@2x

After (with DM sans) : 
CleanShot 2023-01-15 at 19 28 16@2x

(Another solution is the remove any reference to DM sans and use the os default ui font.)

Related to  #1170 #1179 #1182

### DIFF
--- a/layouts/default.vue
+++ b/layouts/default.vue
@@ -13,7 +13,7 @@ const showUserPicker = logicAnd(
 </script>
 
 <template>
-  <div h-full>
+  <div h-full font-sans>
     <main flex w-full mxa lg:max-w-80rem>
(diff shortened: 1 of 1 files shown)


antfu:
Thanks for bringing this up. I actually feel lucky that we made this mistake so the current Elk could follow the system's default. Let's discuss at https://github.com/elk-zone/elk/pull/1182 first. Than you.

PR #1192, opened 2023-01-15
refactor: remove unused DM Sans

antfu:
### Description

#1190
https://github.com/elk-zone/elk/pull/1182#issuecomment-1383225179

### Additional context

---

- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Translations update
- [ ] Other

### DIFF
--- a/styles/global.css
+++ b/styles/global.css
@@ -13,13 +13,6 @@ html {
 }
 @font-face {
   font-display: swap;
-  font-family: 'DM Sans';
-  font-style: normal;
-  font-weight: 400;
-  src: url(/fonts/DM-sans-v11.ttf) format('truetype');
-}
-@font-face {
-  font-display: swap;
   font-family: 'DM Serif Display';
   font-style: normal;
   font-weight: 400;
--- a/docs/app.vue
+++ b/docs/app.vue
@@ -3,13 +3,3 @@
     <NuxtPage />
   </AppLayout>
 </template>
-
-<style>
-@font-face {
-  font-display: swap;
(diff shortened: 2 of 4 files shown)


rshigg (approved):
Looks good 👍

styles/global.css
```diff
@@ -11,13 +11,6 @@ html {
   font-weight: 400;
   src: url(/fonts/DM-mono-v10.ttf) format('truetype');
```
Shinigami92, line 12: why only the sans font, but not the mono?

also: what is this `DM Serif Display`?

styles/global.css
```diff
@@ -11,13 +11,6 @@ html {
...
 }
-@font-face {
-  font-display: swap;
-  font-family: 'DM Sans';
-  font-style: normal;
-  font-weight: 400;
-  src: url(/fonts/DM-sans-v11.ttf) format('truetype');
```
Shinigami92, line 19: I think you can also delete the `/public/fonts/DM-sans-v11.ttf` file

antfu: I prefer to keep it in case we need it in the future. With this change it does not download to the client side so it's good enough.

styles/global.css
antfu: Because they are in use right now. We could change them later but it's not the goal of this PR>

styles/global.css
Shinigami92: Is it not possible to revert it at all time via git if needed?

PR #1195, opened 2023-01-15
feat: theme colors

### LINKED ISSUE #761: Custom user theming support

antfu:
### Description

rework #800, close #761

This PR introduces the theme color system:

image

The colors are generated from Elk's theme color, with the same saturation and brightness (via HCL).

### Additional context

---

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Translations update
- [ ] Other


rshigg:
@antfu this is a great opportunity to fix a particular issue I've been coming across. Currently because each custom color is passed in as a hex value unocss doesn't know how to change the alpha channel. I think if we pass these colors as RGB we'll be able to append the opacity css var as well, which will let us do stuff like `bg-primary/20` for example.

antfu:
@rshigg what's the specific issue you are trying to solve?

constants/themes.json
```diff
@@ -0,0 +1,119 @@
+[
+  [
+    "#cc7d24",
+    {
+      "--c-primary": "#cc7d24",
+      "--c-primary-active": "#b16605",
```
boehs, line 6: Why not 
```json
{
  "#cc7d24": {
    "--c-primary": "#cc7d24",
    "--c-primary-active": "#b16605"
    ...
  }
}
```

boehs (commented):
I quite like this as a first step. Some thoughts

- Could there be a general RGB picker as well as the predefined options? (seems very possible due to the modular themes and the generation script)
- Could an "advanced" option be added that allows arbitrary variables to be changed?

rshigg:
Issue was probably the wrong word to use but for example when I was styling the notify button yesterday I wanted to set the background to the primary color but with low opacity. Currently that's not possible because custom colors don't get passed the opacity variable.

constants/themes.json
antfu: It doesn't really matter tbh.

antfu:
@boehs About color pickers or so, I would prefer not to do them because:

- It's hard for the user to pick a good color that has good contrast in all designed places
- CSS variables are already relatively easy to change; there are many browser extensions that could override the styles with much more controls. I would consider this a niche case, and someone really needed it and did have the solution already. It's not worth it for the complexity on our side.

antfu:
@rshigg you may use `bg-primary-fade`. `bg-primary/20` won't be possible as we are using CSS variables that are not known at runtime. Indeed there are solutions to store rgb numbers in the variables, but I feel the cases are more like we should limit the variations we used for the primary color.

antfu:
For reference, interestingly, we have this: https://github.com/catppuccin/elk

boehs:
@antfu 

> It's hard for the user to pick a good color that has good contrast in all designed places

Isn't that a responsibility the user is willingly taking on by using the color picker? I don't imagine it would come as a surprise

ryanccn:
> Indeed there are solutions to store rgb numbers in the variables

I think it would be a good idea to use RGB values in the CSS variables; currently we already have `--c-bg-base-rgb` replicating `--c-bg-base` because we need a translucent background for the sticky navbar. Having all of the colors defined as RGB in variables would be a more flexible solution IMO

antfu:
I have added the rgb version of the primary color. Still not very sure if it's worth it to change very colors to it, as it make the value hard to maintain. But now we are having a convention, which should be able to make future extensions easier.

antfu:
> Isn't that a responsibility the user is willingly taking on by using the color picker? I don't imagine it would come as a surprise

@boehs Yes, but too many options are not always a good thing. It's a bit hard to explain the philosophy here, but I think my points are:

- Keep it simple for regular users
- For power users, you will never be able to fulfill their needs. While this is a web app and also open source, it's pretty hackable in many aspects.

PR #1198, opened 2023-01-15
fix: publish widget mentions

patak-dev:
### Description

https://github.com/elk-zone/elk/commit/9571d7338a824865e3ddad7be52587c657b806ea introduced mentions as separate lines. This PR fixes some issues with the new mentions scheme for the PublishWidget:

- wrap mentions (they were getting out of the widget)
- use short handle (like in the messages)
- click to remove (with basic feedback using red color)

Adding a new mention can be done as usual by including it in the message. We can later improve the UI, but I think we could merge this PR to let users delete mentions until we do.

---

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Translations update
- [ ] Other

### DIFF
--- a/composables/masto/account.ts
+++ b/composables/masto/account.ts
@@ -7,10 +7,14 @@ export function getDisplayName(account: mastodon.v1.Account, options?: { rich?: 
   return displayName.replace(/:([\w-]+?):/g, '')
 }
 
+export function acctToShortHandle(acct: string) {
+  return `@${acct.includes('@') ? acct.split('@')[0] : acct}`
+}
+
 export function getShortHandle({ acct }: mastodon.v1.Account) {
   if (!acct)
     return ''
-  return `@${acct.includes('@') ? acct.split('@')[0] : acct}`
+  return toShortHandle(acct)
 }
 
 export function getServerName(account: mastodon.v1.Account) {
--- a/composables/masto/account.ts
+++ b/composables/masto/account.ts
@@ -14,7 +14,7 @@ export function acctToShortHandle(acct: string) {
 export function getShortHandle({ acct }: mastodon.v1.Account) {
   if (!acct)
     return ''
-  return toShortHandle(acct)
+  return acctToShortHandle(acct)
 }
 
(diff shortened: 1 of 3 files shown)


composables/masto/account.ts
```diff
@@ -7,10 +7,14 @@ export function getDisplayName(account: mastodon.v1.Account, options?: { rich?:
...
 
+export function acctToShortHandle(acct: string) {
+  return `@${acct.includes('@') ? acct.split('@')[0] : acct}`
+}
+
 export function getShortHandle({ acct }: mastodon.v1.Account) {
   if (!acct)
     return ''
-  return `@${acct.includes('@') ? acct.split('@')[0] : acct}`
+  return toShortHandle(acct)
```
ushuz, line 17: Maybe it's `acctToShortHandle()` here?

patak-dev: Good catch! Thanks for the review 🙏🏼 
Fixed at [`e32b483` (#1198)](https://github.com/elk-zone/elk/pull/1198/commits/e32b483a8afaa1da14258ce8d4d55d078de3fd8f)

composables/masto/account.ts
```diff
@@ -7,10 +7,14 @@ export function getDisplayName(account: mastodon.v1.Account, options?: { rich?:
...
+export function acctToShortHandle(acct: string) {
+  return `@${acct.includes('@') ? acct.split('@')[0] : acct}`
+}
+
 export function getShortHandle({ acct }: mastodon.v1.Account) {
   if (!acct)
     return ''
-  return `@${acct.includes('@') ? acct.split('@')[0] : acct}`
+  return acctToShortHandle(acct)
 }
```
danielroe, line 18: I'm not quite sure why these two were split out - aren't they now doing the same thing?

danielroe (approved):

composables/masto/account.ts
patak-dev: Mostly, but one receieves an account object, and the other one a well defined acct string. In the mentions where I used this, there was no account object

Issue #1199, opened 2023-01-15
Empty draft kept in local storage and no way for removing it


edimitchel:
### Pre-Checks

- [X] Check that there isn't [already an issue](https://github.com/elk-zone/elk/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/elk-zone/elk/discussions) or join our [Discord Chat Server](https://chat.elk.zone).
- [X] Providing a screenshot or video to reproduce the issue or show visually what was meant.
- [ ] I am willing to provide a PR.

### Describe the bug

Working on https://github.com/elk-zone/elk/issues/1196, I observe that when we compose a post without touching anything, a new draft is created with an empty status content.
On leaving the editor by changing the page, the draft is kept in locale storage. Coming back, we could get cached information like _writing language_ of the post.

This is potentially a feature as we probably want to keep such information, but at this moment, even touching anything the draft is kept in memory.
It should only persisted (or not removed) when something changed from default (warning text, language, content).

Also, on missing thing is to be able to reset gracefully the editor.

### Reproduction video or screenshot

_Had issue for recording video._

To reproduce the issue, do it in incognito mode
Open your dev tools and show local storage data. Focus on elk-draft value.
By default, an empty draft is created with status empty. Go on notifications page. elk-draft is still there even nothing has been changed. >> No need to keep it.

Changing anything like language, privacy, warning message or status should make the draft be kept in memory.

Issue #774, opened 2023-01-04
Implement Preferences API


sxzz:
Docs: https://docs.joinmastodon.org/methods/preferences/

https://discord.com/channels/1044887051155292200/1060212767585488996/1060274770056319046

shuuji3:
https://github.com/elk-zone/elk/commit/16561845 implemented `GET /api/v1/preferences` but `PUT /api/v1/preferences` is not available yet.

PR #1200, opened 2023-01-15
feat: add writing language preference

edimitchel:
### Description

Add the option of writing language to use.
image

When a new message is composed, it will apply the writing language as language of the toot.

An issue could make the language not well applied when the composer has been already shown before the writing language change. See #1199 

### Additional context
Not sure if the way of passing vModel.

---

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Translations update
- [ ] Other

### DIFF
--- a/components/settings/SettingsLanguage.vue
+++ b/components/settings/SettingsLanguage.vue
@@ -2,14 +2,17 @@
 import type { ComputedRef } from 'vue'
 import type { LocaleObject } from '#i18n'
 
-const userSettings = useUserSettings()
+const props = defineProps<{
+  language?: string
+}>()
+const model = useVModel(props, 'language')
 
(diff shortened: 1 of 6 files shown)


antfu:
I guess we should read and sync with preference from Mastodon?

components/settings/SettingsLanguage.vue
```diff
@@ -2,14 +2,17 @@
 import type { ComputedRef } from 'vue'
 import type { LocaleObject } from '#i18n'
 
-const userSettings = useUserSettings()
+const props = defineProps<{
```
sxzz, line 5: we can use “defineModel” instead

sxzz:
relate issue of Preferences API #774

PR #1201, opened 2023-01-15
perf: improve perf on rendering long list of replies

danielroe:
### Description

Fetching a lstatus with many replies can take a very long time to render - in my testing, https://elk.zone/mastodon.social/@MarkRuffalo/109638532333776039 took over 10s to fetch from the Mastodon API and then at least as long to render all of the 700+ replies in the DOM.

This PR makes a couple of improvements (including moving to `shallowRef` from `ref` within `useAsyncData`) but the most significant is the switch to using virtual scroller for post replies. It also renders a skeleton when loading the context. In view of past regressions (e.g. on notifications) when adopting virtual scroller, it would be good to have some review and testing before merging.

### Additional context

---

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Translations update
- [ ] Other

### DIFF
--- a/pages/[[server]]/@[account]/[status].vue
+++ b/pages/[[server]]/@[account]/[status].vue
@@ -23,7 +23,7 @@ const { client } = $(useMasto())
 const { data: context, pending: pendingContext, refresh: refreshContext } = useAsyncData(
   `context:${id}`,
   async () => client.v1.statuses.fetchContext(id),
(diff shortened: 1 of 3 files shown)


antfu (approved):

antfu:
Awesome! I was just thinking about it yesterday

PR #1212, opened 2023-01-16
feat: add app initializing animation

alexzhang1030:
### Description

When have bad networks, app's hydrating is a LONG time...

I add  a animation to makes loading have more interesting

https://user-images.githubusercontent.com/49969959/212635138-d53aec6f-3dce-47c6-8262-a7a3cbb88d79.mp4

Use `v-show` , does not effect `SEO`( perhaps? )

### Additional context

---

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Translations update
- [ ] Other

### DIFF
--- a/layouts/default.vue
+++ b/layouts/default.vue
@@ -14,7 +14,10 @@ const showUserPicker = logicAnd(
 
 <template>
   <div h-full>
-    <main flex w-full mxa lg:max-w-80rem>
+    <main v-show="!isHydrated" w-full h-100vh flex="~ center" aria-label="elk is initializing">
+      <img src="/logo.svg" shrink-0 aspect="1/1" sm:h-16 xl:h-32 class="rtl-flip" animate-pulse :alt="$t('app_logo')">
+    </main>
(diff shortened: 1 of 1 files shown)


patak-dev:
I think we should show this only after 1 or 2 secs, and maybe we should show the elk in the position where it will appear later so there is one less layout shift. cc @danielroe, in case something else may be going on here... I would expect that the browser would be able to show the pre-generated HTML right away.

danielroe:
Is the issue here that we just show much less on mobile when first loading? the browser does show the pre-generated html but maybe we're not generating enough for a proper 'app shell' on mobile.

alexzhang1030:
Yes, it does show prerender HTML, we can't do anything, like:

https://user-images.githubusercontent.com/49969959/213066834-13ec3a23-b906-42d9-8a49-a1cb798af3bd.mp4

But actually this is an extreme case, I will update waiting hydration delay later.

boehs:
I feel like the big thing I'm learning from this is that much more of Elk could be server rendered

PR #1223, opened 2023-01-16
fix(editor): mention handling on post editing

antfu:


smallnine9:
cool, I also want to fix this, but I spent too much time to read the code :(

PR #1232, opened 2023-01-16
feat: Add tooltip to account notification bell

tom-sherman:
Honestly didn't know what this button did without looking at the HTML 😅


tom-sherman:
A UX detail that may not be favourable: The tooltip stays open after pressing the button.

https://user-images.githubusercontent.com/9257001/212735006-ef889920-3d1c-4f87-b019-d8fe72771050.mov

jeffsikes:
This is already being implemented here: https://github.com/elk-zone/elk/pull/1127

patak-dev:
Thanks for the PR @tom-sherman, let's focus on #1127, you could leave a review there if you have some time.

PR #1233, opened 2023-01-16
feat: support all HTML elements allowed by Glitch

### LINKED ISSUE #1230: HTML from Mastodon is sanitized to plain text
### Pre-Checks

- [x] I am willing to provide a PR.

### Describe the bug

glitch-soc instances support Markdown toots, and the content clients receive are in the form of HTML. However, Elk's sanitization seems to resolve all of these HTML tags into plain text, including text formatting ones like `<strong>` and `<b>`.

### Additional Context

_No response_

jviide:
### Description

This pull request adds support for the rest of the HTML elements that glitch-soc allows. Some of the elements (like `<em>`) were already supported. This pull request fills up the roster with the rest of the elements based on https://github.com/glitch-soc/mastodon/blob/13227e1dafd308dfe1a3effc3379b766274809b3/lib/sanitize_ext/sanitize_config.rb.

The post pictured in issue #1230 looks like this after applying this change:

Screenshot 2023-01-16 at 19 33 37

Some elements, like `<blockquote>`, are now kept in the DOM, but their are essentially unstyled. This is probably best solved by someone who - unlike me - has a some styling skills 🙂

The test snapshot for one test needed to be updated, as `<strong>` elements are now allowed.

Closes #1230.

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Translations update
- [ ] Other

### DIFF
--- a/composables/content-parse.ts
+++ b/composables/content-parse.ts
@@ -35,12 +35,35 @@ const sanitizer = sanitize({
   code: {
     class: filterClasses(/^language-\w+$/),
   },
-  // other elements supported in glitch
-  h1: {},
-  ol: {},
-  ul: {},
-  li: {},
+  // Other elements supported in glitch, as seen in
+  // https://github.com/glitch-soc/mastodon/blob/13227e1dafd308dfe1a3effc3379b766274809b3/lib/sanitize_ext/sanitize_config.rb#L75
+  abbr: {
+    title: keep,
+  },
+  del: {},
+  blockquote: {
+    cite: filterHref(),
+  },
+  b: {},
+  strong: {},
+  u: {},
+  sub: {},
+  sup: {},
+  i: {},
   em: {},
+  h1: {},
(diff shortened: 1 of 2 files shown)


composables/content-parse.ts
```diff
@@ -35,12 +35,35 @@ const sanitizer = sanitize({
...
+  h2: {},
+  h3: {},
+  h4: {},
+  h5: {},
   ul: {},
-  li: {},
-  em: {},
+  ol: {
+    start: keep,
+    reversed: keep,
```
sxzz, line 62: I wonder why we use the function `keep` here? Instead of `undefined` directly?

sxzz: Sorry, it's okay, I read it wrong.

sxzz (approved):

danielroe (approved):

PR #1240, opened 2023-01-16
feat: StackBlitz code block expansion

patak-dev:
### Description

StackBlitz added the code in the meta description when sharing a URL with `L1-L3` (highlight code in a playground)

https://stackblitz.com/edit/node?file=index.js%3AL1-L3

This would allow us to render the code using shiki instead of using a social image and have a snippet where you can click to see it working in a playground.

The design is still in progress, sending the PR to test and start playing

image

image

---

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Translations update
- [ ] Other

### DIFF
--- a/components/status/StatusContent.vue
+++ b/components/status/StatusContent.vue
@@ -6,7 +6,7 @@ const { status, context } = defineProps<{
   context?: mastodon.v2.FilterContext | 'details'
 }>()
 
-const isDM = $computed(() => status.visibility === 'direct')
+const isDM = $computed(() => false) // status.visibility === 'direct')
 const isDetails = $computed(() => context === 'details')
 
 // Content Filter logic
--- a/components/status/StatusContent.vue
+++ b/components/status/StatusContent.vue
@@ -6,7 +6,7 @@ const { status, context } = defineProps<{
   context?: mastodon.v2.FilterContext | 'details'
 }>()
 
-const isDM = $computed(() => false) // status.visibility === 'direct')
+const isDM = $computed(() => status.visibility === 'direct')
 const isDetails = $computed(() => context === 'details')
 
 // Content Filter logic
--- a/components/status/StatusPreviewCard.vue
+++ b/components/status/StatusPreviewCard.vue
@@ -16,6 +16,6 @@ const gitHubCards = $(usePreferences('experimentalGitHubCards'))
 
 <template>
   <StatusPreviewGitHub v-if="gitHubCards && providerName === 'GitHub'" :card="card" />
-  <StatusPreviewStackBlitz v-else-if="gitHubCards && providerName === 'stackblitz.com'" :card="card" :small-picture-only="smallPictureOnly" :root="root" />
+  <LazyStatusPreviewStackBlitz v-else-if="gitHubCards && providerName === 'stackblitz.com'" :card="card" :small-picture-only="smallPictureOnly" :root="root" />
   <StatusPreviewCardNormal v-else :card="card" :small-picture-only="smallPictureOnly" :root="root" />
 </template>
--- a/components/status/StatusPreviewCard.vue
+++ b/components/status/StatusPreviewCard.vue
@@ -15,7 +15,7 @@ const gitHubCards = $(usePreferences('experimentalGitHubCards'))
 </script>
 
 <template>
-  <StatusPreviewGitHub v-if="gitHubCards && providerName === 'GitHub'" :card="card" />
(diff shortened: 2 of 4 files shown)


antfu (approved):
LGTM. /cc @danielroe do you think we should add `<Lazy` to those components so they won't be included in the initial payload?

edimitchel:
I would suggest to lower a bit the open text, it looks a bit too high

patak-dev:
@edimitchel I changed the order so it looks more like a card and integrated both with the rounded borders 
image

@antfu I think as we keep adding more of these card types, we should lazy load them. There may be a lot of users that won't need these components because their communities don't use GitHub, StackBlitz, or other domains that will add later. We could do that in another PR. I wonder if we should have a general On/Off switch for all the custom cards or individual ones. Leaning towards having only one to avoid too many options.

components/status/StatusContent.vue
```diff
@@ -6,7 +6,7 @@ const { status, context } = defineProps<{
   context?: mastodon.v2.FilterContext | 'details'
 }>()
 
-const isDM = $computed(() => status.visibility === 'direct')
+const isDM = $computed(() => false) // status.visibility === 'direct')
```
edimitchel, line 9: debugging ?

danielroe:
> do you think we should add <Lazy to those components so they won't be included in the initial payload?

Yes, we should. And probably also with the GitHub card too. 👍

components/status/StatusPreviewCard.vue
```diff
@@ -9,74 +9,13 @@ const props = defineProps<{
...
-      min-w-24 w-24 h-24 sm="min-w-32 w-32 h-32" bg="slate-500/10" flex justify-center items-center
-      :class="[
-        root ? 'rounded-lg' : '',
-      ]"
-    >
-      <div :class="cardTypeIconMap[card.type]" w="30%" h="30%" text-secondary />
-    </div>
-    <StatusPreviewCardInfo :p="isSquare ? 'x-4' : '4'" :root="root" :card="card" :provider="providerName" />
-  </NuxtLink>
+  <StatusPreviewStackBlitz v-else-if="gitHubCards && providerName === 'stackblitz.com'" :card="card" :small-picture-only="smallPictureOnly" :root="root" />
```
danielroe, line 19:
```suggestion
  <LazyStatusPreviewStackBlitz v-else-if="gitHubCards && providerName === 'stackblitz.com'" :card="card" :small-picture-only="smallPictureOnly" :root="root" />
```

components/status/StatusPreviewCard.vue
```diff
@@ -9,74 +9,13 @@ const props = defineProps<{
...
-const cardTypeIconMap: Record<mastodon.v1.PreviewCardType, string> = {
-  link: 'i-ri:profile-line',
-  photo: 'i-ri:image-line',
-  video: 'i-ri:play-line',
-  rich: 'i-ri:profile-line',
-}
 </script>
 
 <template>
   <StatusPreviewGitHub v-if="gitHubCards && providerName === 'GitHub'" :card="card" />
```
danielroe, line 18:
```suggestion
  <LazyStatusPreviewGitHub v-if="gitHubCards && providerName === 'GitHub'" :card="card" />
```

components/status/StatusPreviewCard.vue
patak-dev: That is all that is needed?! Nuxt never stops surprising me

patak-dev:
Ok, I'm happy with this style for now, I think we can iterate and improve it in further PRs.

image

edimitchel:
Yes iterate probably on limiting lines of code and let expand when needed.

edimitchel (approved):

components/status/StatusPreviewStackBlitz.vue
```diff
@@ -0,0 +1,93 @@
...
+      display-block of-hidden
+      bg-card
+      w-full
+      justify-center
+      p-3
+      pb-4
+    >
+      <div flex justify-between>
+        <p flex gap-1>
+          <span>Code Snippet from</span><span>{{ meta.file }}</span><span text-secondary>{{ `- Lines ${meta.lines}` }}</span>
```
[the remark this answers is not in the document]
patak-dev: Ill do it in another PR, I want to play with main.elk.zone to test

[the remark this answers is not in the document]
patak-dev: done

PR #1250, opened 2023-01-17
fix: show replying user in zen mode editor

### LINKED ISSUE #1222: Zen Mode editor should show who you are replying to
Zen Mode is now hiding the tagged users in a post. This also affects the editor when replying.

Situation right now:

| Not Zen Mode | Zen Mode |
| --- | --- |
| Screenshot 2023-01-16 at 1 07 37 PM | Screenshot 2023-01-16 at 1 07 47 PM |

### Suggested solution

We want to always display the tagged users in the editor when writing a reply, even on Zen Mode, to allow modifying recipients (adding/removing) 

This is important because a thread could go long and at some point you will not know who's involved and might forget you are in Zen Mode (as zen mode users tend to do)

webfansplz:
### Description

https://user-images.githubusercontent.com/22515951/212854474-aaff890b-5feb-48ee-8096-14a9996cf051.mov

Closes #1222 

### Additional context

---

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Translations update
- [ ] Other

### DIFF
--- a/components/content/ContentMentionGroup.vue
+++ b/components/content/ContentMentionGroup.vue
@@ -1,5 +1,11 @@
+<script setup lang="ts">
+defineProps<{
+  showInZenMode?: boolean
+}>()
+</script>
+
 <template>
-  <p flex="~ gap-1 wrap" items-center text-sm class="zen-none">
+  <p flex="~ gap-1 wrap" items-center text-sm :class="{ 'zen-none': !showInZenMode }">
     <span i-ri-arrow-right-line ml--1 text-secondary-light /><slot />
   </p>
 </template>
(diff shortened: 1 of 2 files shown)


components/content/ContentMentionGroup.vue
```diff
@@ -1,5 +1,11 @@
+<script setup lang="ts">
+defineProps<{
+  showInZenMode?: boolean
```
patak-dev, line 3: I think we should name this `replying`, we may need to apply later other modifications based on this context.

patak-dev (approved):

PR #1251, opened 2023-01-17
feat: add current user following hashtags page

### LINKED ISSUE #1194: provide a list of followed hashtags

There are a few different ways to discover a hashtag, and I see where I can follow a given hashtag with the ⭐.  However, there currently is no way to get a list of the hashtags I am following.

Shinigami mentioned in Discord that the underlying Mastodon API wasn't there before, but is implemented now:
https://docs.joinmastodon.org/methods/followed_tags/

### Suggested solution

I defer to the experts on design matters...  One place to show this is in the user's profile, i.e. either at the top of my profile

100 Posts, 55 Following, 21 Followers, and **5 Hashtags**

where you probably want a simple gerund for "following a hashtag" but I can't think of one.

Or, you could modify the "Following" view to include both people and hashtags in some way.

### Alternative

_No response_

### Additional context

Related Mastodon issue:
https://github.com/mastodon/mastodon/issues/20763

and PR:
https://github.com/mastodon/mastodon/pull/21773

alexzhang1030:
### Description

closes #1194 

Add current user following hashtags page:

PC: 
image

Mobile:

image

### Additional context

If servers are not support this feature, will return `404` error with this message:

image

Should we solve this specific case ?

---

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Translations update
- [ ] Other


Shinigami92:
Have not reviewed the code and somehow there is no netlify preview
1. Are the hashtags clickable so you can jump to the hashtag site itself?
2. Is it possible to show a possibility to unfollow a hashtag from this site directly without jumping into the specific site?

alexzhang1030:
> Have not reviewed the code and somehow there is no netlify preview
> 
> 1. Are the hashtags clickable so you can jump to the hashtag site itself?
> 2. Is it possible to show a possibility to unfollow a hashtag from this site directly without jumping into the specific site?

It uses the same component with `explore/tags`

image
 
So:

1. Elk can show tags contents, and this is clickable.
2. We can follow/unfollow by here ↓

image

Shinigami92:
Interesting, I didn't know this tab/view/component

We could add my idea in another PR later 👌

alexzhang1030:
> Interesting, I didn't know this tab/view/component
> 
> We could add my idea in another PR later 👌
> 
> btw: I implemented the star for follow/unfollow a tag 😉

For my personal thoughts, the color of the star is a little unclear.

Maybe we could add a conspicuous color ? 🤔

sxzz:
maybe we can move it to more action menu. it looks too obtrusive on mobile devices.
image

alexzhang1030:
> 

Agree, Let's try it out.

alexzhang1030:
# Update

Now it is placed at account more:

image

PC:

image

Mobile:

image

patak-dev:
I think users may not need to follow/unfollow tags too many times, so it may be better to hide this screen a bit more. The official Mastodon app has it in Settings, maybe we should move it there too? 
image

alexzhang1030:
> I think users may not need to follow/unfollow tags too many times, so it may be better to hide this screen a bit more. The official Mastodon app has it in Settings, maybe we should move it there too? image

I guess this is an anthor feature, the [related issue](https://github.com/mastodon/mastodon/issues/20763) in `mastodon` is still open

ronilaukkarinen:
As this is currently on the official Mastodon as a feature, would be great to see it in Elk as well!

Sma11X:
This is helpful for unfollowing tags, otherwise I might only unfollow tags on mastodon 🙏

shuuji3:
#1251 implemented the hashtags page as an independent page with the star follow button.

Issue #1267, opened 2023-01-17
Clickable links in image Alt


zeroby0:
### Clear and concise description of the problem

The Alt seems like a nice place to put image sources. 

- If there's multiple images, there's no confusion which source belongs to which image because the link is in the image alt, not in the toot body
- The links don't take up limited toot characters

The only downside to having links in images I see is poorly written screen-readers that might read out the link letter-by-letter.

If the links in Alt can be identified and made clickable in Elk, that would be pretty neat.

### Suggested solution

In the Alt popup, search for URLs in the image alt and replace them with anchor tags for the links.

https://github.com/sindresorhus/get-urls can find links in strings.

### Alternative

Show a list of links in the description, below the description.

### Additional context

I've tried search+replace links with `<a href=''>link</a>` in image Alt, but it's inserted as-is as a string, instead of being parsed as HTML. 

Naively letting all text in Alt be parsed as HTML is dangerous so we would need a good way to insert anchor tags where the URLs are.

ZMYaro:
> The Alt seems like a nice place to put image sources.
Alt text is absolutely not the place for that.  Alt text is only supposed to be an alternative text substitute for the image if the user is unable to see the image (due to the image failing to load, the user being visually impaired, the browser not showing the image, etc.).  It is _not_ supposed to be used as a caption or source link.

zeroby0:
Okay then

PR #1270, opened 2023-01-17
feat: show list of links in image description

zeroby0:
### Description

Fixes https://github.com/elk-zone/elk/issues/1267

If there are URLs in image description, the ALT popup show them as a list of clickable links.

A new i18n key `desc_urls` is added. Translators, please add translations if merged 🙏🏻 

#### Preview

https://deploy-preview-1270--elk-zone.netlify.app/universeodon.com/@elkdev/109705079934697907

Screenshot from 2023-01-18 00-04-44

### Additional context

It would've been ideal to replace URLs inline with anchor links. I saw StatusBody.vue doing something like that with `contentToVNode`, which I tried but couldn't get working. If you know how to use it properly, it might be better to do that instead of what this PR is doing.

> e.g. is there anything you'd like reviewers to focus on?

I've never used typescript before and am not very experienced with Vue. I hope this doesn't break something because I haven't added any types.

The pnpm.lock file has changed and some of the dependencies had their `dev` key removed automatically. If something mysteriously breaks in your system, please see if reverting this PR fixes it.

---

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Translations update
- [ ] Other


zeroby0:
Doing this the vnode way is probably much better. No added dependencies and one less i18n key for translators to keep track of. If i figure out how to do that, I'll open a new PR.

patak-dev:
Thanks for the PR @zeroby0! I think this makes sense, but as you described, we should have the links replaced in place. Let's leave this open to see if others jump in to help you with the implementation.

zeroby0:
Closing in favour of #1285

PR #1274, opened 2023-01-17
feat: explore and search unified

### LINKED ISSUE #894: Make the sidenav responsive on height
_[According to Discord thread](https://discord.com/channels/1044887051155292200/1061775169254854778)_

I have a concern about the sidenav layout and the consequence on wide (or not tall enough) screens.
See the attached image of the side nav: the account part is cropped (the viewport size is 1536x731 on my desktop)

image

It should never get scroll bar for UX purpose!

For making sure scroll never comes, we should add a responsiveness mechanism to hide items using usability order.
Great principales written by @patak-dev :

- Explore, local, Bookmarks, favourites could be removed
- We need an order of what is more important and remove the rest if the screen is not tall enough
- The account should always be visible

My opinion is agreeing previous point but I would like to extends (it's my own):

- Removing Home icon in favor of the Elk icon as home icon
- Put Parameters closer to account (remove padding)
- Put Compose icon button at the same place to the back icon (or aside to)

### LINKED ISSUE #1258: Use explore page as search page

A bit related with #1120.

As a user I want to have less nav item in the side nav and thinking that searching is quite like explore to something I also know what it is.

### Suggested solution

As 🐦 did, exploring is searching for something or just for curiousity.
My suggestion would be to use the explore page also as the search page. The impact would be mainly on the side nav or the bottom nav for mobile.

The behaviour would be :
On desktop: search field will have placeholder Explore and search. On focus, the explore page will be displayed.
The search field could be an "Explore" button which became an input on click.

On mobile, the search icon would show the explore page but with the search field focused to let user search. When result appears, explore page would be replaced by.

### Alternative

_No response_

### Additional context

This will probably fix the issue about side nav height (#894)

edimitchel:
### Description

Merge usage of search and explore in one.
Use search and explore button which will redirect to the explore page but put the focus on the search field which takes same place than the search and explore just clicked. 

Resolves #1258
Resolves #894 

explore-search

### Additional context

This has been made to optimize the number of nav items without loosing any.

---

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Translations update
- [ ] Other


patak-dev:
One issue with this approach is that you can no longer search while looking at the context around. You are forced to go to the search route. I evaluated doing this for search instead of having the results in a dropdown but I didn't move forward for this reason. I think a search input that doesn't take you out of the current route is important to keep.

edimitchel:
Updated @patak-dev @ZMYaro @baptiste0928

patak-dev:
I think we should leave Search on the left sidebar, I like that it is stable as you resize. And I don't mind it disappearing for smaller screens instead of taking height space. I would prefer to keep the current interface and review once we have more content on the right side to see how everything fits together.

baptiste0928:
The search pane on mobile devices is actually empty, what do you think about merging search and explore but for only on mobile?

Current search pane:
search-mobile-current

With this PR:
search-mobile-new

edimitchel:
I'll do it in a separate PR
I have also few UI optimization ;)

PR #1282, opened 2023-01-18
fix: prevent default event when hide dropdown

webfansplz:
### Additional context

---

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Translations update
- [ ] Other

### DIFF
--- a/components/common/dropdown/Dropdown.vue
+++ b/components/common/dropdown/Dropdown.vue
@@ -8,6 +8,7 @@ defineProps<{
 
 const dropdown = $ref<any>()
 const colorMode = useColorMode()
+const mask = useMask()
 
 const hide = () => dropdown.hide()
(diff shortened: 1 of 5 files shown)


patak-dev:
@Akryum this seems to be a feature everyone would need, are we missing something in the way we use VDropdown? @webfansplz I wonder if we should better upstream this feature if this is not yet supported

webfansplz:
> @Akryum this seems to be a feature everyone would need, are we missing something in the way we use VDropdown? @webfansplz I wonder if we should better upstream this feature if this is not yet supported

I've been thinking about it, but I'm not sure if it's within the scope of `VDropdown`. /cc @Akryum

Akryum:
It could totally be an option in floating-vue.

Issue #1283, opened 2023-01-18
Image deletion is *really* hard to see


dilinger:
### Pre-Checks

- [X] Check that there isn't [already an issue](https://github.com/elk-zone/elk/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/elk-zone/elk/discussions) or join our [Discord Chat Server](https://chat.elk.zone).
- [X] Providing a screenshot or video to reproduce the issue or show visually what was meant.
- [ ] I am willing to provide a PR.

### Describe the bug

It was really difficult for me to discover how to delete an image from a post that's being drafted. I was actually going to file a bug about how you can't do it, until I read https://github.com/elk-zone/elk/issues/860#issuecomment-1381546859 and realized that you can, it's just really hard to see because it's a tiny light-grey 'X' on top of a grey background.

It should be easier for users to see how to delete an image. Either the X should always remain (like the ALT and Edit buttons), or it should have a different color background (again, like the ALT and Edit buttons), or it should be a color that is more likely to be visible.

### Reproduction video or screenshot

Here's what it looks like without a mouse-over

![Screenshot-2023-01-18,01-57-08](https://user-images.githubusercontent.com/11581730/213104851-52687644-f959-451f-9180-9e86d19063f8.png)

Here's what it looks with with a mouse over the first image. It's basically invisible in this image.

![Screenshot from 2023-01-18 01-58-58](https://user-images.githubusercontent.com/11581730/213105408-f75212bc-5b35-4916-ad0c-46697e3f32b4.png)

Here's the mouse over the second image. It's a tiny bit more visible here, but still very easy to miss/hard to see.

![Screenshot from 2023-01-18 02-00-42](https://user-images.githubusercontent.com/11581730/213105510-0bfae1de-cd45-4607-a4c4-e3c5013e2f42.png)

### Additional Context

I want to be clear what the failure mode for this kind of thing is. Elk is amazing (thank you for all your hard work!), but after the 0.5 update I was stuck with an in-progress post where I couldn't delete images. Because of that, I couldn't get rid of the post itself, and so I couldn't create a new post at all. I had to switch away from Elk in order to create a new post. Deleting images should be immediately obvious.

dilinger:
Oops, I see that this is a duplicate of #1067 ,  sorry.

PR #1286, opened 2023-01-18
fix: make attachement removal button static & more visible #1067

### LINKED ISSUE #1067: Cross for clossing images is almost invisible in white images
### Pre-Checks

- [X] I am willing to provide a PR.

### Describe the bug

When adding images to a new publication a cross appears when hovering over the image, but this cross is almost invisible in white images.

Screenshot 2023-01-13 at 12-13-23 Elk
The cross is almost invisible in this white background screenshot
elk-screenshot
Hovering over the cross

### Additional Context

The cross should be visible no matter the colour of the image.

dilinger:
The black background makes it a lot more obvious, and let's always leave it there rather than requiring a mouseover. Added the translated text as an aria-label for screenreaders/a11y, because "X" isn't super descriptive.

### Description

This takes a chainsaw to the existing image deletion button in the PublishAttachement component, but.. It's currently really hard to find/see, as described in #1067 and #1283.  It also makes it stylishly consistent w/ the Edit button.

This fixes #1067. Tested on both my laptop (firefox 102) and my phone (android 11).

### Additional context

---

- [X] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Translations update
- [ ] Other

### DIFF
--- a/components/publish/PublishAttachment.vue
+++ b/components/publish/PublishAttachment.vue
@@ -31,7 +31,7 @@ const toggleApply = () => {
         v-if="removable"
         :aria-label="$t('attachment.remove_label')"
         class="bg-black/75 hover:bg-red/75"
-        text-white px2 py2 rounded-5 cursor-pointer
+        text-white px2 py2 rounded-full cursor-pointer
         @click="$emit('remove')"
       >
         <div i-ri:close-line text-3 :class="[isHydrated && isSmallScreen ? 'text-6' : 'text-3']" />
--- a/components/publish/PublishAttachment.vue
+++ b/components/publish/PublishAttachment.vue
@@ -30,7 +30,8 @@ const toggleApply = () => {
       <div
         v-if="removable"
         :aria-label="$t('attachment.remove_label')"
-        class="bg-black/75 text-white px2 py2 rounded-5 cursor-pointer"
+        class="bg-black/75 hover:bg-red/75"
+        text-white px2 py2 rounded-5 cursor-pointer
         @click="$emit('remove')"
       >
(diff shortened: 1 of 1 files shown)


dilinger:
Oh yeah, here's what it looks like on firefox:

Screenshot-2023-01-18,04-24-09

antfu:
Can we keep using the icon instead of the char `X`? And also would be better to make the button square

edimitchel (changes requested):
Please make it round

Capture d’écran 2023-01-18 à 16 42 57

Capture d’écran 2023-01-18 à 16 46 51


Use also [i-ri:close-line icon](https://icones.js.org/collection/ri?s=close-line)

dilinger:
Um, square or round? I personally like it square so that it matches both the image itself as well as the Edit button.

Servelan:
Can't see the image delete icon (using the web version). This is an accessibility issue, btw.

edimitchel:
We should probably use an invert filter to be sure the contrast is respected.

dilinger:
Alright, how's this? Here's what it looks like on my end:

Screenshot-2023-01-21,03-05-32

patak-dev (approved):
Thanks! Let's merge it and keep iterating. We'll need to extract a close button component that we can reuse in other places in the app later.

Servelan:
Accessibility would suggest a bigger x (50% bigger).  Good background, 
good contrast, however.

patak-dev:
@Servelan PR or issue welcome about the 50% bigger, thanks for your feedback

PR #1293, opened 2023-01-18
feat: hide redudant mentions

patak-dev:
### Description

Hides redundant mentions in status when these are already in context (in the replying to badge or connected to the previous status)


antfu:
Can we have a few tests to freeze its behaviour if you'd like to?

patak-dev:
Good call @antfu, I found an edge case to fix thanks to the tests (we weren't using the path in the app, but better to have it working already). Let's merge it so we can test it with the canary folks

Issue #597, opened 2022-12-28
Explore tab is missing on mobile


TodePond:
On mobile, the explore tab is gone!
It was recently replaced with the search tab.

I **don't** like how twitter does this:
On twitter, the 'explore' and 'search' tabs are the same thing. (Presumably because they want to try to hook you with viral posts at every opportunity?)

However, some users (eg: me) don't want to be bombarded with trending posts/topics every time they want to do a simple search. I think that the explore tab should be added back in to the '...' menu.

![image](https://user-images.githubusercontent.com/15892272/209806398-1c514b1d-8e95-4235-bcb9-a99dee446e7c.png)

PR #1301, opened 2023-01-18
feat: use explorer page as search for mobile only

### LINKED ISSUE #894: Make the sidenav responsive on height

### LINKED ISSUE #1258: Use explore page as search page

### LINKED ISSUE #1560: UX: Improve user's path to explore on mobile
Right now I always press search, because default Mastodon interface has Search + Explore combined. Because of this I always end up pressing search (as it's sticky on the bottom nav) and end up on this empty view:

In my opinion the current search feature on Mastodon is rarely used. The views are actually great as combined, because the search occupies very little space. 

Suggestion 1:
Combine these views.

Suggestion 2:
Add explore icon on bottom to the mobile in place of search.

edimitchel:
### Description

Merge usage of search and explore in one only for mobile or medium screen (could have both search and explore icon active for medium screen).

Resolves #1258
Resolves #894 
Resolves #1560 

explore-search-2

---

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Translations update
- [ ] Other


edimitchel:
> As I said on #1274, I think it would make more sense to follow Twitter's example and have the Search icon _replace_ the Explore icon in medium size windows instead of being a redundant link to the same page.

Applied.

antfu (approved):
Looks nice!

boehs (changes requested):
I think that "Explore" should be hidden in the mobile submenu, because they are the same page. Otherwise, LGTM

patak-dev:
Let's merge this one so we avoid the need to resolve conflicts and iterate. I think we'll need to review how this will look if we show search results similar to the timeline more like a dialog for mobile later.

TodePond:
Hey just to chime in on this - I feel like there should be a setting to toggle this as desired by the user. I really like having 'explore' hidden away, as to not get in the way when just trying to do a simple search. On twitter, I understood why they might want to shove trending posts in my face, but it has been very refreshing to not have that on mastodon.

I think many other people appreciate that about mastodon VS twitter as well :)
Explained a bit further back here: https://github.com/elk-zone/elk/issues/597

Anyone else relate?

patak-dev:
I agree here, I was also hesitant about making this change but it is good that we tried to get some more data. There were a few messages from folks confused by it on Mastodon. I think we should revert and have to explore and search separately again and let people explicitly go to explore when they want to discover content (I don't use explore at all for example, so I also prefer not to see it).

About having an option to merge them, I think it is better to avoid it for now as we already have too many options, and it isn't clear to me that a lot of people are asking for this.

This PR introduces others changes too that are great (sticky user in the sidebar, etc), @edimitchel let me know if you'd like to send a revert only for the merging part, if not I'll get to it later today or tomorrow

edimitchel:
I can do a revert but I would like to suggest another thing: on mobile, when search is clicked only show the search input and focus to it.
When focus is lost or scroll is made more than 100px, the search input would be hidden.
Also, I think the search result should not be a popper but should take place on the body.
I'll revert the merge now then work on that feature if it sounds good for you.

patak-dev:
Yes, I think we can revert then keep exploring how to improve things.

TodePond:
That sounds like a good plan!

Either way, thanks for putting in time to *explore* these options @edimitchel, I appreciate it! It's useful to see features like this actually implemented, to get a better understanding of them 🙌

PR #1303, opened 2023-01-18
feat: added autoplay setting

### LINKED ISSUE #1235: Autoplay setting

As a user I want to be able to disable autoplay on media, especially video, so that I can save on bandwidth when on mobile connection 

### Suggested solution

Add a setting to enable/disable autoplay. My preferred default option would be „off“, but that’s discussable.

### Alternative

Disable autoplay on mobile connections.

### Additional context

_No response_

I think I could work on a PR, if the preferred solution is discussed 😊

niklaswolf:
### Description
Added an autoplay setting and functionality to prevent autoplaying videos when the setting is disabled.
The default value is set like the current behaviour, so autoplaying is enabled.

closes #1235 

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Translations update
- [ ] Other

### DIFF
--- a/components/status/StatusAttachment.vue
+++ b/components/status/StatusAttachment.vue
@@ -65,9 +65,11 @@ const video = ref<HTMLVideoElement | undefined>()
 const prefersReducedMotion = usePreferredReducedMotion()
 const isAudio = $computed(() => attachment.type === 'audio')
 
+const enableAutoplay = usePreferences('enableAutoplay')
+
 useIntersectionObserver(video, (entries) => {
   const ready = video.value?.dataset.ready === 'true'
-  if (prefersReducedMotion.value === 'reduce') {
+  if (prefersReducedMotion.value === 'reduce' || !enableAutoplay.value) {
     if (ready && !video.value?.paused)
       video.value?.pause()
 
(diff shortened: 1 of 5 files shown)


userquin (approved):

patak-dev:
Auto-play is a preference that already exists in Mastodon. @sxzz should we add this one under Elk's settings until we start reading user preferences? I think we should start reading from the correct API directly. 
Here is [the API for reference](https://docs.joinmastodon.org/methods/preferences/) that doesn't document the [auto play of gifs field](https://github.com/mastodon/mastodon/blob/main/app/serializers/rest/preferences_serializer.rb#L33)

sxzz:
@patak-dev Seems that Mastodon API didn't expose that configuration item. So we can have it first.

Store it in a filter content (discussed in Discord before) when we implemented preference store.

antfu:
May we also use https://www.w3schools.com/tags/att_video_autoplay.asp for browsers that supports it?

rshigg:
@antfu great point! All it would take is just adding `:autoplay="enableAutoplay"` to the video element in `StatusAttachment.vue`.

niklaswolf:
@antfu yep this feature also has very good support. I did not add it to the video elements because I thought there was a reason why this was done the way it is right now (intersection observer).
But one way or the other a setting to opt out of autoplay is needed 😊

antfu:
Let's have this for now. Opened https://github.com/elk-zone/elk/issues/1324 for tracking in another PR

PR #1310, opened 2023-01-19
fix: respect the length seen

Mini-ghost:
### Description

Currently, when we enter an emoji, the number of characters is not equal to 1

before

On Mastodon, this emoji has 1 character count.

Because, `'👩‍👩‍👧‍👧'.length` is `11`. And in this particular case we can't directly use `Array.from(str).length` to get the number of characters.

```ts
const length = Array.from('👩‍👩‍👧‍👧').length // length is 7
```

In modern browsers we can use `Intl.Segmenter` to get the correct length.

```ts
const length = [...new Intl.Segmenter().segment('👩‍👩‍👧‍👦')].length // length is 1
```

But that is not supported by FireFox. So I chose `lodash.toarray`.

```ts
import toArray from 'lodash.toarray'
const length = toArray('👩‍👩‍👧‍👦').length // length is 1
```
after

References:

- [Intl.Segmenter](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Segmenter)
- [How to count the correct length of a string with emojis in javascript?](https://stackoverflow.com/a/54369605/17048127)
- [Count character length of emoji?](https://stackoverflow.com/a/46085147/17048127)

**UPDATE**

Thank for @Brawaru sharing, [string-length](https://github.com/sindresorhus/string-length) is smaller than `lodash.toarray`

| lbrary | size |
| :--- | :--- |
| lodash.toarray | Bundle size is 4.62KB -> 2.06KB (gzip) |
| string-length| Bundle size is 1.53KB -> 821B (gzip) |

```ts
import stringLength from 'string-length'
const length = stringLength'👩‍👩‍👧‍👦')// length is 1
```

---

- [x] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Translations update
- [ ] Other


Brawaru:
[`string-length`](https://github.com/sindresorhus/string-length) seems to achieve the same goal while being smaller than `lodash.toarray`:

[bundlejs](https://bundlejs.com) for [`lodash.toarray`](https://bundlejs.com/?q=%28import%29lodash.toarray&treeshake=%5BtoArray%5D&text=%22toArray%28%27%F0%9F%91%A9%E2%80%8D%F0%9F%91%A9%E2%80%8D%F0%9F%91%A7%E2%80%8D%F0%9F%91%A7e%CC%88%27%29.length%22&config=%7B%22analysis%22%3A%22treemap%22%7D) (2.06KB gzip) vs [`string-length`](https://bundlejs.com/?q=%28import%29string-length&treeshake=%5BstringLength%5D&text=%22stringLength%28%27%F0%9F%91%A9%E2%80%8D%F0%9F%91%A9%E2%80%8D%F0%9F%91%A7%E2%80%8D%F0%9F%91%A7e%CC%88%27%29%22&config=%7B%22analysis%22%3A%22treemap%22%7D) (833B gzip).

Mini-ghost:
@Brawaru Thanks for sharing, it's really awesome 🧡

danielroe (approved):

PR #1311, opened 2023-01-19
chore: preview header only when it's already set

webfansplz:
### Additional context

e.g. Twitter

#### Before

https://user-images.githubusercontent.com/22515951/213364829-1feac29e-df88-4770-97dd-0fc6b97a5a49.mov

#### After

https://user-images.githubusercontent.com/22515951/213365146-25525178-cc68-4e3f-b756-ecbf70ba11a5.mov

---

- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Translations update
- [x] Other

### DIFF
--- a/components/account/AccountHeader.vue
+++ b/components/account/AccountHeader.vue
@@ -20,6 +20,7 @@ const relationship = $(useRelationship(account))
 
 const namedFields = ref<mastodon.v1.AccountField[]>([])
 const iconFields = ref<mastodon.v1.AccountField[]>([])
+const hasHeader = !account.header.endsWith('/original/missing.png')
 
 function getFieldIconTitle(fieldName: string) {
   return fieldName === 'Joined' ? t('account.joined') : fieldName
@@ -30,6 +31,8 @@ function getNotificationIconTitle() {
 }
 
 function previewHeader() {
+  if (!hasHeader)
+    return
   openMediaPreview([{
     id: `${account.acct}:header`,
     type: 'image',
@@ -85,7 +88,7 @@ const isNotifiedOnPost = $computed(() => !!relationship?.notifying)
 
 <template>
   <div flex flex-col>
-    <button border="b base" z-1>
+    <button border="b base" z-1 :class="!hasHeader && 'cursor-auto'">
       <img h-50 height="200" w-full object-cover :src="account.header" :alt="t('account.profile_description', [account.username])" @click="previewHeader">
     </button>
     <div p4 mt--18 flex flex-col gap-4>
--- a/components/account/AccountHeader.vue
+++ b/components/account/AccountHeader.vue
@@ -88,9 +88,9 @@ const isNotifiedOnPost = $computed(() => !!relationship?.notifying)
 
 <template>
   <div flex flex-col>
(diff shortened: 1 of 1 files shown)


components/account/AccountHeader.vue
```diff
@@ -20,6 +20,7 @@ const relationship = $(useRelationship(account))
 
 const namedFields = ref<mastodon.v1.AccountField[]>([])
 const iconFields = ref<mastodon.v1.AccountField[]>([])
+const hasHeader = !account.header.endsWith('/original/missing.png')
```
Mini-ghost, line 23:
```suggestion
const hasHeader = $computed(() => !account.header.endsWith('/original/missing.png'))
```

How about using `$computed()` here?

components/account/AccountHeader.vue
```diff
@@ -85,7 +88,7 @@ const isNotifiedOnPost = $computed(() => !!relationship?.notifying)
 
 <template>
   <div flex flex-col>
-    <button border="b base" z-1>
+    <button border="b base" z-1 :class="!hasHeader && 'cursor-auto'">
       <img h-50 height="200" w-full object-cover :src="account.header" :alt="t('account.profile_description', [account.username])" @click="previewHeader">
     </button>
```
danielroe, line 93: Probably better not to output a button at all if it's not clickable.

```suggestion
    <component :is="hasHeader ? 'button' : 'div'" border="b base" z-1>
      <img h-50 height="200" w-full object-cover :src="account.header" :alt="t('account.profile_description', [account.username])" @click="previewHeader">
    </component>
```

components/account/AccountHeader.vue
```diff
@@ -30,6 +31,8 @@ function getNotificationIconTitle() {
 }
 
 function previewHeader() {
+  if (!hasHeader)
+    return
```
danielroe, line 35: If we don't add a click handler at all, we can avoid the early return here too.

```suggestion
```

components/account/AccountHeader.vue
Mini-ghost: What about using a disabled button here?

danielroe (approved):

components/account/AccountHeader.vue
danielroe: I think that would also be a good solution - do you think it has any benefits over a div in this case?

Mini-ghost: When using a disabled button, the accessibility semantics are: there was a button here, but now it's disabled; with a div it doesn't have any semantics.

In addition, when using the disabled button to encounter an update, in Vue's diff algorithm, only update the attribute instead of creating a new DOM node

danielroe: Yes, that's true - but I'm not sure whether that's _better_ when it comes to the header of a page. Probably it is? I don't expect there will ever be any update to `hasHeader`.

Mini-ghost: I checked where this component is used. It seems that there will be no state update, so naturally there will be no such problem.

Sorry, I only looked at this component and did not pay special attention to the page it uses.

PR #1316, opened 2023-01-19
feat: use colorized logo

edimitchel:
### Description

Let the logo use the newest introduced color theme preference.

logo-colorized

### Additional context

We could probably add an opt-out in order to avoid the logo to be affected by the color preference

---

- [ ] Bug fix
- [x] New Feature
- [ ] Documentation update
- [ ] Translations update
- [ ] Other


patak-dev (approved):

antfu:
I am not sure we should do this: https://elk.zone/m.webtoo.ls/@antfu/109703346039698907

We are using the icon to show different builds (production, canary, preview, etc.). With this change, it would make it ambiguous and less recognizable (e.g. if one changed to theme color to red, a long while later, they might think Elk's logo IS red).

PR #1319, opened 2023-01-19
refactor: reduce unnecessary html structures

Mini-ghost:
### Description

This PR has adjusted two parts:

~**1. Add type attribute to all buttons**~

~Currently some buttons have a type attribute, but most do not. so I add the corresponding type attribute based on its function.~

~Also, can we add [vue/html-button-has-type](https://eslint.vuejs.org/rules/html-button-has-type.html) eslint rule?~

**1. Rreduce unnecessary html structures**

According to the [HTML5 Spec Document from W3C](http://w3c.github.io/html/textlevel-semantics.html#the-a-element)

> Content model:
>
> Transparent, but there must be no interactive content descendant, a element descendant, or descendant with the tabindex attribute specified.

Therefore `<button>` is not recommended to be nested inside `<a>` elements.

**2. Remoce unused component**

I tried searching for `<PublishButton>` and `<LazyPublishButton>` on the project, it probably won't be used, so I removed it.

---

- [ ] Bug fix
- [ ] New Feature
- [ ] Documentation update
- [ ] Translations update
- [x] Other


components/notification/NotificationPreferences.client.vue
```diff
@@ -143,6 +143,7 @@ onActivated(() => (busy = false))
               </fieldset>
               <div flex="~ col" gap-y-4 gap-x-2 py-1 sm="~ justify-between flex-row">
                 <button
+                  type="submit"
```
userquin, line 146: there is no need to add it, it is the default

components/notification/NotificationPreferences.client.vue
```diff
@@ -168,6 +169,7 @@ onActivated(() => (busy = false))
             <form flex="~ col" mt-4 @submit.prevent="removeSubscription">
               <span border="b base 2px" class="bg-$c-text-secondary" />
               <button
+                type="submit"
```
userquin, line 172: there is no need to add it, it is the default

Mini-ghost:
> Currently some buttons have a type attribute, but most do not. so I add the corresponding type attribute based on its function.

@antfu  Understood, should I revert this commit, or close it ?

components/notification/NotificationPreferences.client.vue
Mini-ghost: According @antfu suggestion, I drop it commit.

antfu (approved):

PR #1321, opened 2023-01-19
feat: normal text color for nav logo

patak-dev:
### Description

See @antfu's comment here https://github.com/elk-zone/elk/pull/1316#issuecomment-1397079724, it is now a bit confusing to know in what env we are.

This PR tests the idea of using normal color for the nav logo (same as many other apps, like GitHub, Twitter, etc). The brand isn't content for the user, and a big primary color blob in the corner is a bit distracting in my opinion.

The PR also switches to directly change `Elk` by `Dev` or `Main` instead of the superinduce so the layout doesn't change much.

image

image

Dev:

image

### DIFF
--- a/components/nav/NavLogo.vue
+++ b/components/nav/NavLogo.vue
@@ -29,9 +29,9 @@
 
 <style scoped>
 svg path.wood {
-    fill: var(--c-dark-primary-light);
+    fill: var(--c-text-secondary);
 }
 svg path.body {
-    fill: var(--c-primary);
+    fill: var(--c-text-base);
 }
 </style>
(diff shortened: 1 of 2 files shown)


wakest:
this looks much better tbh

patak-dev:
@ksesong the screenshot point is very good feedback, and you also bring something that was bothering me a lot when doing screenshots for social. The dev green logo looks quite strange, I actually did the last post cropping it to avoid showing the logo.

I reverted the label change, so now it looks like before:
image

antfu (approved):

antfu:
I think we could have it, and release 0.6.1, maybe to see how they react to this?

patak-dev:
Yes, let's do it. Better to change it before everyone updates and starts to like playing with the Elk color. Worst case this could be a very niche preference :)

PR #1349, opened 2023-01-20
feat(native): add pwa to Tauri

userquin:
Since Tauri is using custom protocol scheme (`tauri://`), native WebView will handle it as non secured, and so, we cannot register the service worker (at least `macOS`): check discord thread https://discord.com/channels/1044887051155292200/1065996405493202944/1066040488005746739


userquin:
@JonasKruckenberg I'll set this on hold since we cannot register yet service worker using custom `tauri://` scheme

Issue #1350, opened 2023-01-20
Custom tall emojis aren't rendered correctly


patak-dev:
https://elk.zone/m.codeflow.club/@stackblitz/109721669842498046

<img width="642" alt="image" src="https://user-images.githubusercontent.com/583075/213744093-7fb2f048-e5f2-4383-82d9-332da0a4cb7a.png">

patak-dev:
This is the emoji
![stackblitz](https://user-images.githubusercontent.com/583075/213744550-bcf3bafb-ae25-4a32-a257-b41a602df934.png)

piotrski:
Haha, they edited out the toot
![Uploading image.png…]()

patak-dev:
> Haha, they edited out the toot

I edited the toot 😅

PR #1352, opened 2023-01-20
fix: rendering of high emoji

piotrski:
fix: #1350

### DIFF
--- a/styles/global.css
+++ b/styles/global.css
@@ -62,10 +62,13 @@ body {
 
 .custom-emoji {
   display: inline-block;
-  overflow: hidden;
+  vertical-align: text-bottom;
+}
+
+.custom-emoji img {
   max-height: 1.3em;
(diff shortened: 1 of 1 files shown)


styles/global.css
```diff
@@ -12,32 +12,6 @@ html {
...
-}
-
-::-webkit-scrollbar-thumb {
-  background: #8885;
-  border-radius: 1px;
-}
-
-::-webkit-scrollbar-thumb:hover {
-  background: #8886;
-}
```
antfu, line 39: Please revert this. If it's a reason, you maybe bring it in another issue/PR instead.

antfu (changes requested):

styles/global.css
piotrski: Sorry, I committed this by mistake. I was testing the styles for the native application. I've already reverted it.

PR #1358, opened 2023-01-20
feat: add user preference to hide translation button

### LINKED ISSUE #1342: Add an option to disable the translate button
I would like to suggest the option to disable the translate button.

niklaswolf:
Simply adds the preference to hide the translation button as requested from some users :)
closes #1342

### DIFF
--- a/components/status/StatusTranslation.vue
+++ b/components/status/StatusTranslation.vue
@@ -10,6 +10,9 @@ const {
   translation,
   enabled: isTranslationEnabled,
 } = useTranslation(status, getLanguageCode())
+const preferenceHideTranslation = usePreferences('hideTranslation')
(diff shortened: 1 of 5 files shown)


edimitchel:
I would prefer to let select which languages to not translate instead of hiding all translation.

patak-dev (approved):
I think we could have both @edimitchel. And this option may also be temporal and removed later. The translate feature is not there yet for me either, so at least this is helpful as a temporal escape hatch.

niklaswolf:
@patak-dev I'm willing to work on the feature to select the languages to not translate. I think it will be useful to have the options to disable translations only for some languages or disable translations as a whole. I'll create a new issue.

PR #1359, opened 2023-01-20
feat(native): Improved layout in elk-native

piotrski:
- Two new variants have been added to UnoCSS: `native:` and `native-mac:`. Styles for these variants will only be generated if the `TAURI_PLATFORM` variable is properly set during the build process.
- The application now respects the user's system preferences for the style and behavior of the style bar on the Mac
- The title bar is always visible in the app, which should fix a problem when the window was not draggable on larger screens. elk-zone/elk-native#43
- This is not the final iteration, in the future, I plan to move the navigation in narrow windows to the left side.

Here are before and after screenshots:
--
650px:
before-650
after-650
1000px:
before-1000
after-1000
1200px:
before-1200
after-1200
1400px:
before-1400
after-1400

closes elk-zone/elk-native#44

### DIFF
--- a/nuxt.config.ts
+++ b/nuxt.config.ts
@@ -47,9 +47,10 @@ export default defineNuxtConfig({
     'floating-vue/dist/style.css',
     '~/styles/default-theme.css',
     '~/styles/vars.css',
+    '~/styles/global.css',
     ...process.env.TAURI_PLATFORM === 'macos'
-      ? ['~/styles/global.css']
-      : ['~/styles/global.css', '~/styles/scrollbars.css'],
+      ? []
+      : ['~/styles/scrollbars.css'],
     '~/styles/tiptap.css',
     '~/styles/dropdown.css',
   ],
--- a/nuxt.config.ts
+++ b/nuxt.config.ts
@@ -47,7 +47,9 @@ export default defineNuxtConfig({
     'floating-vue/dist/style.css',
     '~/styles/default-theme.css',
     '~/styles/vars.css',
-    '~/styles/global.css',
+    ...process.env.TAURI_PLATFORM === 'macos'
+      ? ['~/styles/global.css']
+      : ['~/styles/global.css', '~/styles/scrollbars.css'],
     '~/styles/tiptap.css',
     '~/styles/dropdown.css',
   ],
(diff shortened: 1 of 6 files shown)


nuxt.config.ts
```diff
@@ -47,7 +47,9 @@ export default defineNuxtConfig({
     'floating-vue/dist/style.css',
     '~/styles/default-theme.css',
     '~/styles/vars.css',
-    '~/styles/global.css',
+    ...process.env.TAURI_PLATFORM === 'macos'
+      ? ['~/styles/global.css']
+      : ['~/styles/global.css', '~/styles/scrollbars.css'],
```
patak-dev, line 52: We could keep global out, no?
```suggestion
    '~/styles/global.css',
    ...process.env.TAURI_PLATFORM === 'macos'
      ? []
      : ['~/styles/scrollbars.css'],
```

patak-dev:
This is awesome, thanks @piotrski! @JonasKruckenberg could you review this one too?

nuxt.config.ts
piotrski: Looks neat!

JonasKruckenberg (approved):
I checked on macOS and this looks great 👍🏻 much better than before. I also like that we now have a strategy for adding more platform specific styles later on, like a custom titlebar on Windows and Linux.

JonasKruckenberg:
Should we just do a new `elk-native` release with these fixes or do you want to cut a new `elk` release soon anyway?

patak-dev:
I think we should do a release soon to have the translation preference

JonasKruckenberg:
> I think we should do a release soon to have the translation preference

Okay! Then just ping me once you release and we can coordinate a release like last time 👍🏻

patak-dev:
Im now thinking we should do it Monday, better to avoid changing elk.zone during the weekend now that we have so many users. Feel free to release native before

JonasKruckenberg:
What did you think @piotrski? I guess this change is a rather big one and I would appreciate getting the release out asap, but maybe waiting for the proper release would be better (stability and fixes that might come in before that?)

patak-dev:
@piotrski I think this PR affected the styles on Web, see the size of the right sidebar here. Would you review what happened here?

image

patak-dev:
@piotrski @JonasKruckenberg I pushed https://github.com/elk-zone/elk/commit/2a836eda3d2db3cb37589b7ea97f9c23e6af0c58 that reverts some of the changes to the default layout. I don't know how this affects the native styles, please follow up with a PR that keeps the current style in the web if you would like changes for the native wrapper.

PR #749, opened 2023-01-04
feat: add image preview swipe

wheatjs:
Closes #534 

https://user-images.githubusercontent.com/7064956/210495626-405aec5e-7a90-4190-a901-facf3fb6d26d.mp4

PR #1367, opened 2023-01-21
fix: bring back image pinch zoom

peterbud:
# Description

This PR aims to bring back the pinch zoom functionality for image previews. This functionality seems to got lost in PR #749 

# What is the purpose of this pull request?
-  [X] Bug fix
-  [ ] New Feature
-  [ ] Documentation update
-  [ ] Translations update
-  [ ] Other

### DIFF
--- a/components/modal/ModalMediaPreviewCarousel.vue
+++ b/components/modal/ModalMediaPreviewCarousel.vue
@@ -2,6 +2,7 @@
 import { SwipeDirection } from '@vueuse/core'
 import { useReducedMotion } from '@vueuse/motion'
 import type { mastodon } from 'masto'
+import { useImageGesture } from '~/composables/gestures'
 
 const { media = [], threshold = 20 } = defineProps<{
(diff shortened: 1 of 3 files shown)


wheatjs (commented):
To be honest this implementation doesn't seem to work that well on my phone. When zooming into an image, scroll bars appear and it doesn't focus on where I'm attempting to zoom. I was going to implement this https://github.com/manuelstofer/pinchzoom#readme library as it seems more battle tested and looking at the source code, it is not trivial. Unless we can find a simpler implementation that works just as well, I think we should go with the 3rd party library for this one. Do you want to try implementing it @peterbud? If not I can give it a go this week.

userquin (commented):
we need another lib

peterbud:
hi @wheatjs sure, go ahead with your implementation. I'm not sure which is the "another library" @userquin meant, but I trust your judgment guys. Closing this PR.

PR #1370, opened 2023-01-21
feat(i18n): add country variants support

userquin:
This PR also includes:
- `en` variants
- `ar` variants
- `es` variants

~~Still on draft, I need to update PWA web manifest support.~~


patak-dev:
I don't think we should add every country variant. These are too many options, and they are all the same in most cases. The dropdown goes out of the screen here for example:

image

It would be good to migrate from the dropdown to a flat list on the screen. But for the variants, we should only add one when there is a difference compared to the generic one. Also, the project is too young and the UI is changing fast, having to maintain all these variants at this point isn't the best strategy. I think we should put this PR on hold for the time being.

patak-dev:
For reference, these are slack options
image

A section of Twitter options
image

A section of Discord options (I think our UI should look like this one by the way)
image

I like the idea of supporting more variants, I'm just discussing here when to do all this work. I don't think we need more variants than these three apps at this point of the project. We can merge a better structure that lets us have ES (espana) and ES (lationamerica) though.

userquin:
@patak-dev from my experience with multi language projects, you should consider to include the infraestructure, country variants can be removed (later we can be added on high demand), the cost is the same we've right now (we need to fix nuxt i18n module)

I'll drop all spanish variants except Spain (the current we've right now ), for arabic I'll keep only Egypt and for english I'll keep both (all country varianst will be there, I'll just comment the entries).

patak-dev (approved):

edimitchel (approved):

userquin:
Added `Dockerfile` to `.eslintignore`, when merging main, `lint-stage` failing...

locales/ar-MA.json
```diff
@@ -0,0 +1 @@
+{}
```
antfu, line 1: If they are empty, can we remove them?

userquin: Sure, I'll wait nuxt pr to fix wrong merge... 

I'll remove also unused `es` country variants files and update contributing guide

userquin: done: PR merged on nuxt i18n module, not yet released, we can merge this, we can change `en-US.json`and update contributing guide later

edimitchel (approved):

mrcego:
> @patak-dev from my experience with multi language projects, you should consider to include the infraestructure, country variants can be removed (later we can add them on high demand), the cost is the same we've right now (we need to fix nuxt i18n module)
> 
> I'll drop all spanish variants except Spain (the current we've right now ), for arabic I'll keep only Egypt and for english I'll keep both (all country varianst will be there, I'll just comment the entries).
> 
> EDIT: I'll try to include `Español (Lationoamérica)` like Discord and Twitter, I need to check the codes (~IIRC `es-415`, Google?~ `es-419`, from Google recaptcha: https://developers.google.com/recaptcha/docs/language)

Nice! this feature opens new lands about CV in others translations like Vite (if is needed).

PR #1371, opened 2023-01-21
feat: allow disabling translation for specific languages

### LINKED ISSUE #1247: Configuration of the languages in which the translation button is hidden

Actually, the translation button is shown for every toot in a different language than the browser language. We should be able to configure for which languages this button is shown.

### Suggested solution

- [ ] Hide translation button for toots in the same language as the UI.
- [ ] Allow to configure untranslated languages in the settings.
- [ ] Move translate button to the "more" menu for toots where it has been hidden, to translate anyway if needed.

### Alternative

_No response_

### Additional context

_No response_

niklaswolf:
- little refactoring: moved some logic into `useTranslation` composable
- moved configurations regarding translations to the language section of the settings
- added possibility to select different languages (currently supported for translation), for which the translation will be hidden in statuses
- added actual logic to hide the translation if the status language should not be translated 

closes #1247


composables/masto/translate.ts
```diff
@@ -63,9 +97,14 @@ export function useTranslation(status: mastodon.v1.Status | mastodon.v1.StatusEd
     translations.set(status, reactive({ visible: false, text: '', success: false, error: '' }))
 
   const translation = translations.get(status)!
+  const userSettings = useUserSettings()
+
+  const shouldTranslate = 'language' in status && status.language && status.language !== to
```
niklaswolf, line 102: Maybe we could add a check here, if the `to` language is supported for translation. Otherwise, the translation will not work any way. See line 11.

patak-dev: Yes, let's do this 👍🏼

patak-dev:
Nice!

This is how these settings look on Twitter. I like the positive vibe on these, asking you what you know instead of what translation to hide for the wording.
image

But I think we should merge this PR after you do the change you mentioned and then iterate. It is important we get the functionality to help users, and we can improve the settings UI later

niklaswolf:
@patak-dev I agree. Added the changes :)

patak-dev (approved):
I think we can move forward to have the basic functionality and let others iterate on improving the UI.

danielroe:
Would you be able to resolve conflicts on this? 🙏

niklaswolf:
@danielroe conflicts are resolved now :)

danielroe:
Thank you ❤️

PR #1376, opened 2023-01-21
feat: range of font size

patak-dev:
We got several requests for a bigger XL font.

This PR keeps the current scheme and just uses a wider range (to have `xl` be 19px instead of 17px).

Should we replace the named options with a range instead? I think we could offer anything between 10px and 20px for example. I don't see a good reason for limiting the options here.


ksesong:
One argument for limiting the font-size options is to enable a clear hierarchy, font-size are different enough to each other, so I know what each level means. The more options we have, the harder it will be to distinguish one size from another, one level from another.

Do you have any example where you would feel a more granular range would be beneficial?

patak-dev:
@ksesong I don't, my thinking was that after this change we are eliminating two options (14px and 16px) that were valid before and folks may be using. It seemed quite arbitrary. Would you review how the new range feels? I think we could move with this PR at least to let people have a proper XL-sized font.

ksesong:
Sorry, I misunderstood the use of the font-size range. I thought these values were used to set the typographic scale within one page, and not as the base font-size for that scale. 

In that case, I see no particular objection for a range. At the same time, if we look at a similar feature (zoom in, zoom out in a browser window), the changes are granular, 30%, 50%, 67%, 80%, 90%, 100%, 110%, 120%, 133%, 150%, 170%, 200%, 240%. Maybe it is worth following these values, and maybe the descriptions (as these are easier to estimate than generic "small/large" sizes).

ksesong:
Another alternative would be to keep the same options as before, and add an XXL option, possibly adding more options in the future there is any specific demand.

patak-dev:
I think a slider with 10 to 20 is the best here

antfu:
@wheatjs did that before at https://github.com/elk-zone/elk/pull/980, I think it would be great to have it

patak-dev:
Thanks for the slider @wheatjs! @antfu moved to 10px-20px range, and it only changes on pointerup to avoid the effect that @edimitchel was concerned about.

In a future PR, we should work on having the focus around the whole slider for keyboard usage. Leaving that one for someone else.

edimitchel (commented):
I see few things which I don't like:

The grey border on FF
Capture d’écran 2023-01-22 à 15 29 53

Also, not sure it's related with your change, but the notice about deployment seems to be to small

patak-dev:
The right sidebar being small is an issue related to this PR https://github.com/elk-zone/elk/pull/1359#issuecomment-1399312087, we can revert it before the next release if nobody finds the regression.

patak-dev:
@edimitchel I tried to remove the gray outline on Firefox, but I don't know what's going on there. I think we should still move forward with this PR and let someone with Firefox quirks experience fix the style for it (or change the design later if that isn't possible)

edimitchel:
@patak-dev I could take a look this evening

ksesong:
You should apply the following style to make the border disappear in Firefox. No idea how to do it with `unocss` though.

```
input[type="range"]::-moz-range-thumb {
  border: 0; // or border-color: ...
}
```

patak-dev:
Thanks @ksesong! Looks a lot better now (there is still a very thin line but we can live with it, it kind of looks part of the design to be honest)
image

antfu:
I guess we could have smaller and lighter dots for the slides, in order to better distinguish the selected one. In general it LGTM

patak-dev:
@antfu good idea, done at [elk-zone/elk@`7fa8b07` (#1376)](https://github.com/elk-zone/elk/pull/1376/commits/7fa8b0708e1f9bbe4d9bee59b19e0585ce3f45e5), let's merge it and we can iterate on the style later. I think we'll end up with other sliders and we could review what works for all of them.

Mini-ghost:
@patak-dev Can we make the clickable block bigger? The clickable height(3.75px) seems to be a bit small now.

small

Would it feel better for the user experience to have the input be as tall as the clickable point?

bigger

patak-dev:
@Mini-ghost would you send a PR?

PR #1377, opened 2023-01-21
feat: add docker build & push to ghcr action

emilweth:
This merge request adds a build & push action to ghcr, the github container registry which is free for open source projets.
It will build a `ghcr.io/elk-zone/elk:<tag-name>` container for every newly created tags, and a `ghcr.io/elk-zone/elk:main` will be built every time a commit is added on main.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -1,0 +1,43 @@
+name: Build & Push docker container
+on:
+  push:
+    branches:
+      - 'main'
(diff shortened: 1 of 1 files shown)


emilweth:
I can also write a helm chart to deploy elk on kubernetes, but it's better to host it on its own repository

danielroe (approved):
This looks very nice, thank you!

Issue #1348, opened 2023-01-20
Move image modal close button to the right corner


piotrski:
Can we move the image modal close button to the right corner to prevent overlap with window controls on mac?

![image](https://user-images.githubusercontent.com/244174/213726775-cc80230f-6907-49ea-95a5-91f794563622.png)

patak-dev:
Yes, lets move it 👍🏼

Aakarshan-369:
Hi! I'm a first-time contributor and I worked on this in PR #1381 
Thanks!

PR #1383, opened 2023-01-22
fix: #1348 Fixed the follow button movement UI

rkmdCodes:
hi I am a new contributor and I have fixed the movement in follow button UI 

Here is the Gif attached ...
gif


components/account/AccountHeader.vue
```diff
@@ -103,7 +103,7 @@ const isNotifiedOnPost = $computed(() => !!relationship?.notifying)
             <AccountHandle :account="account" />
           </div>
         </div>
-        <div absolute top-18 inset-ie-0 flex gap-2 items-center>
+        <div absolute top-18 inset-ie-0 flex gap-2 justify-center>
```
patak-dev, line 106: `items-center` here is correct. The issue is that the bell notification button is taller than the follow button. If the bell notification button size is reduced to match the follow button, then there will not be a pixel shift

rkmdCodes: Thanks for the review!
I edited the commit and reverted the previous changes `justify-center` back to `item-center`  
I found out that the issue is resolved by changing the padding  of bell button from 0.5 rem to 0.4 rem `p2` to `p1.6` but I am getting type check error in CI check. I am still working on the issue.

patak-dev (approved):
Thanks!

components/account/AccountHeader.vue
patak-dev: I think you needed to add `class="p1.6"` for this to work. I changed it back to `p2` though and made the icon `text-sm` so we have more coherent padding.

rkmdCodes:
I learned a lot and this was my first Open source project!

patak-dev:
@rkmdCodes it gets better and better! Welcome to Elk and OSS 🙌🏼

PR #1391, opened 2023-01-22
feat: follow tags from trends page

Shinigami92:
This will let users un/follow tags from the tags trends page directly

See: https://deploy-preview-1391--elk-zone.netlify.app/universeodon.com/explore/tags

https://user-images.githubusercontent.com/7195563/213929073-671d68b7-d6d9-4c53-a8f6-a0d83ce69402.mov


components/tag/TagActionButton.vue
```diff
@@ -12,10 +12,20 @@ const emit = defineEmits<{
...
 
 const toggleFollowTag = async () => {
-  if (tag.following)
+  if (tag.following) {
     await client.v1.tags.unfollow(tag.name)
-  else
+      .then(() => {
+        // eslint-disable-next-line vue/no-mutating-props
+        tag.following = false
+      })
```
danielroe, line 20:
```suggestion
      // eslint-disable-next-line vue/no-mutating-props
      tag.following = false
```

components/tag/TagActionButton.vue
```diff
@@ -12,10 +12,20 @@ const emit = defineEmits<{
...
+  }
+  else {
     await client.v1.tags.follow(tag.name)
+      .then(() => {
+        // eslint-disable-next-line vue/no-mutating-props
+        tag.following = true
+      })
```
danielroe, line 27:
```suggestion
      // eslint-disable-next-line vue/no-mutating-props
      tag.following = true
```

components/tag/TagActionButton.vue
```diff
@@ -12,10 +12,20 @@ const emit = defineEmits<{
 const { client } = $(useMasto())
 
 const toggleFollowTag = async () => {
-  if (tag.following)
+  if (tag.following) {
     await client.v1.tags.unfollow(tag.name)
-  else
+      .then(() => {
+        // eslint-disable-next-line vue/no-mutating-props
```
danielroe, line 18: elsewhere we destructure a value from the prop so we don't mutate the value...

components/tag/TagActionButton.vue
```diff
@@ -12,10 +12,16 @@ const emit = defineEmits<{
 const { client } = $(useMasto())
 
 const toggleFollowTag = async () => {
-  if (tag.following)
+  if (tag.following) {
     await client.v1.tags.unfollow(tag.name)
-  else
+      // eslint-disable-next-line vue/no-mutating-props
+      tag.following = false
```
Shinigami92, line 18:
```suggestion
    // eslint-disable-next-line vue/no-mutating-props
    tag.following = false
```

components/tag/TagActionButton.vue
```diff
@@ -12,10 +12,16 @@ const emit = defineEmits<{
...
+  }
+  else {
     await client.v1.tags.follow(tag.name)
+      // eslint-disable-next-line vue/no-mutating-props
+      tag.following = true
```
Shinigami92, line 23:
```suggestion
    // eslint-disable-next-line vue/no-mutating-props
    tag.following = true
```

components/tag/TagActionButton.vue
Shinigami92: For now this is okay, as it would only affect this for a short time until you reload the context or whatever

I plan in a future PR to build up a cache like `cacheStatus` for tags
This way we can also (maybe) save some api-calls when accessing a tag page a second time, but I need to try that

components/tag/TagActionButton.vue
```diff
@@ -12,10 +12,16 @@ const emit = defineEmits<{
 const { client } = $(useMasto())
 
 const toggleFollowTag = async () => {
-  if (tag.following)
+  if (tag.following) {
```
danielroe, line 15:
```suggestion
  tag.following = !tag.following
  if (tag.following)
```

components/tag/TagActionButton.vue
```diff
@@ -12,10 +12,16 @@ const emit = defineEmits<{
...
     await client.v1.tags.unfollow(tag.name)
-  else
+    // eslint-disable-next-line vue/no-mutating-props
+    tag.following = false
+  }
+  else {
```
danielroe, line 20:
```suggestion
  else
```

components/tag/TagActionButton.vue
```diff
@@ -12,10 +12,16 @@ const emit = defineEmits<{
...
     await client.v1.tags.follow(tag.name)
+    // eslint-disable-next-line vue/no-mutating-props
+    tag.following = true
+  }
```
danielroe, line 24:
```suggestion
```

Shinigami92 (commented):
apply suggestion

components/tag/TagActionButton.vue
Shinigami92: I put it below the if-else, as it should be only applied when the await doesn't throw

components/tag/TagCard.vue
```diff
@@ -11,12 +11,34 @@ const to = $computed(() => {
...
+  if (!el && !text)
+    go(evt)
+}
+
+function go(evt: MouseEvent | KeyboardEvent) {
+  if (evt.metaKey || evt.ctrlKey)
+    window.open(to)
+  else
+    router.push(to)
+}
```
danielroe, line 30: I would be more comfortable if we were able to use a normal link. can we prevent the click from bubbling up from the button instead?

Shinigami92: I tried, and nope

This is copied and adjusted from https://github.com/elk-zone/elk/blob/2a836eda3d2db3cb37589b7ea97f9c23e6af0c58/components/status/StatusCard.vue#L46-L64

Later (in a separate PR) we could extract it into a composable

I tried: `@click.stop`, `@click.prevent`, `@click.native` and some other things
Then I looks how the actions in a status are doing it, and found this above...

components/tag/TagCard.vue
```diff
@@ -11,12 +11,34 @@ const to = $computed(() => {
...
 </script>
 
 <template>
-  <NuxtLink :to="to" block p4 hover:bg-active flex justify-between>
+  <div
+    block p4 hover:bg-active flex justify-between
```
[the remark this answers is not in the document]
Shinigami92: For now okay
But maybe we need to adjust later a bit as in a status it also does not apply for the whole status card but only near the text

components/tag/TagActionButton.vue
```diff
@@ -17,6 +17,9 @@ const toggleFollowTag = async () => {
   else
     await client.v1.tags.follow(tag.name)
 
+  // eslint-disable-next-line vue/no-mutating-props
+  tag.following = !tag.following
```
danielroe, line 21: I think this should be before the request - using optimistic ui as we do elsewhere

Shinigami92: You overlooked my explanation!
https://github.com/elk-zone/elk/pull/1391#discussion_r1083544459

danielroe: I did indeed!
I think we should use try/catch then and revert the change if it throws

Shinigami92: okay will do it

danielroe (approved):

PR #1392, opened 2023-01-22
fix: unescape html entities in page title

### LINKED ISSUE #1372: `'` is replaced with `&#39;` in title for post
Example: https://elk.zone/m.webtoo.ls/@patak/109722747892738830

userquin:
closes #1372


userquin:
IIRC, that's not true in the context we're using it, I need to add a check to unescapce these entities: ensure the title is enclosed by double quotes: this one `"<title>"` or this one  `XXXX in Elk: "<title>"` can be safely escaped

composables/setups.ts
```diff
@@ -22,6 +22,11 @@ export function setupPageHeader() {
       if (buildInfo.env !== 'release')
         titleTemplate += ` (${buildInfo.env})`
       return titleTemplate
+        .replaceAll('&#34;', '"').replaceAll('&quot;', '"')
+        .replaceAll('&#38;', '&').replaceAll('&amp;', '&')
```
danielroe, line 26: what about using `tiny-decode` here too?

userquin: done

userquin:
We've a small problem, what if the account name plus server has more than 60 chars? The pattern should be `<account> in <server>: "<message>"` (it depends on i18n, for example in spanish `<account> en <server>: "<message>"`) and so if `<account> in <server>:` has more than 60 chars long we lost the initial `"`...

danielroe (approved):

PR #1395, opened 2023-01-22
fix: css bg color typo in top bar on mobile

horvbalint:
This is a very small fix (1char) for a typo introduced a few days earlier. 

Because of this typo the blurred top bar on mobile looks like this:
Screenshot from 2023-01-22 21-08-32

Instead of what it is supposed to look like :)
Screenshot from 2023-01-22 21-08-47


patak-dev (approved):

patak-dev:
Thanks! I was wondering why I wasn't enjoying the effect anymore in mobile ❤️

PR #1405, opened 2023-01-23
fix: Allow viewport zoom

neilstuartcraig:
Currently, users can't scale (pinch zoom) on Elk mobile devices, this means that it's difficult/impossible to read/see some images posted. This PR allows zooming, which is also an accessibility requirement, according to [Mozilla Developer Network docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Viewport_meta_tag#viewport_basics):

> `maximum-scale`: ...Any value less than 3 fails accessibility...

> `user-scalable`: ...Setting the value to 0, which is the same as no, is against Web Content Accessibility Guidelines (WCAG)...

This PR changes the `viewport` meta element from:
```
<meta name="viewport" content="width=device-width,initial-scale=1,maximum-scale=1,user-scalable=0,viewport-fit=cover">
```

to
```
<meta name="viewport" content="width=device-width,initial-scale=1,user-scalable=1,viewport-fit=cover">
```

(noting that `maximum-scale` has a default and maximum of `10` so this will be the maximum unless it's re-added with a specific value)

### DIFF
--- a/components/modal/ModalContainer.vue
+++ b/components/modal/ModalContainer.vue
@@ -72,7 +72,7 @@ const handleFavouritedBoostedByClose = () => {
     </ModalDialog>
     <ModalDialog
       :model-value="isMediaPreviewOpen"
-      w-full max-w-full h-full max-h-full
+      w-full h-full
       bg-transparent border-0 shadow-none
       @update:model-value="closeMediaPreview"
     >
--- a/components/modal/ModalMediaPreview.vue
+++ b/components/modal/ModalMediaPreview.vue
@@ -37,7 +37,7 @@ onUnmounted(() => locked.value = false)
 </script>
 
 <template>
-  <div relative h-full w-full flex pt-12 w-100vh @click="onClick">
+  <div relative flex pt-12 w-100vh @click="onClick">
     <button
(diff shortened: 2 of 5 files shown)


cooperaj:
Mmm, unless I'm missing something this isn't doing anything in Android 13 Chrome?

neilstuartcraig:
> Mmm, unless I'm missing something this isn't doing anything in Android 13 Chrome?
> 
> Using the deployment preview above

Oh, I didn't see the preview link (that's really cool!). 
I confirm it's not doing anything on Android 13 Firefox either. I'll have a look and get it sorted, thanks for the quick reply.

neilstuartcraig:
This is now allowing pinch-zoom on the whole page but not on images. I'll keep going! I believe it'll be CSS again now that the viewport config is unlocked.

peterbud:
See #668 and  #1367

patak-dev:
Thanks for the PR @neilstuartcraig, but disabling zooming in the app is intentional (as other similar apps do). We need to add pinch to zoom for images (and where it makes sense in the future). There is work ongoing to support this as linked by @peterbud.

PR #1413, opened 2023-01-24
feat: improve set font size user experience

Mini-ghost:
This PR has adjusted two parts:

1. Make set font size clickable block bigger (94b836ecc37a7bb090e0ed7316fea607a9df9b07)
  small
  bigger

2. Add tooltip for every set font size point (839b5b02422c0329c311a9afba78f0dd39d1bdfb)
  tooltip


patak-dev:
Thanks for the PR!
I see two issues compared to the current one at https://main.elk.zone
I'm on Chrome/MacOS M1

- If I hover just above one of the options in the slider, the cursor starts to rapidly change better normal and pointer.
- Grabbing the selecting option and sliding no longer works.

PR #1414, opened 2023-01-24
feat(native): add native settings section

marekvospel:
This pull request adds a settings section that is only shown in elk native app and `minimize_to_tray` configuration option.

## Related
https://github.com/elk-zone/elk-native/pull/58


patak-dev:
I wonder if we should add these options in elk-native itself, we should be able to add some hooks to extend the options 🤔 
cc @danielroe @JonasKruckenberg

danielroe:
Yes. We could potentially move the tauri module that exists in this repo -> elk native repo itself.

JonasKruckenberg:
I don't know how we would integrate this with the current submodule setup, but if you have any smart ideas, moving as much elk-native specific code into that repo sounds like a good solution to me generally

marekvospel:
I don't think it would be possible to move the tauri module into elk-native repository, as that could cause some type issues (just adding tauri-runtime-only types in this repository I had problems with types, see https://github.com/elk-zone/elk/blob/8bd02816203f5e00c7440689a877a5888107613a/modules/tauri/runtime/native-settings-undefined.ts)

At least I can't think of any good and easy solution

marekvospel:
I've resolved the merge conflicts. Should we wait till https://github.com/elk-zone/elk-native/issues/65 is resolved before merging this?

I feel like for now we could just merge it and once it is decided whether we will continue to use git submodules or start using other methods, just migrate some of this code. I'll gladly migrate this into elk-native once and in case it is necessary.

marekvospel:
@JonasKruckenberg This method could be added to `tauri-plugin-store` to make it easier to work with the values reactively. However it is framework specific, so there could be some problems with dependencies and other framework integrations would probably need to be added.
https://github.com/elk-zone/elk/blob/c383d93989224b519fdb826e94d16c19c3d6fa2a/modules/tauri/runtime/native-settings.ts#L8-L35

marekvospel:
@Shinigami92 any chance for a review? This PR has been sitting here without any activity for a while

modules/tauri/runtime/native-settings.ts
```diff
@@ -0,0 +1,47 @@
...
+        await store.set(key, value)
+        await store.save()
+        refValue = value
+        trigger()
+      },
+    }
+  })
+
+  const result = await store.get(key)
+  if (result !== undefined && result !== null)
```
Shinigami92, line 31:
```suggestion
  if (result != null)
```

modules/tauri/runtime/native-settings.ts
```diff
@@ -0,0 +1,47 @@
...
+    refValue = result as T
+
+  return ref
+}
+
+async function nativeSettings() {
```
Shinigami92, line 37:
```suggestion
async function nativeSettings(): Promise<NativeSettings> {
```

modules/tauri/runtime/native-settings.ts
```diff
@@ -0,0 +1,47 @@
...
+  return {
+    minimizeToTray: await tauriStoreRef(store, 'minimize_to_tray', true),
+  }
+}
+
+interface NativeSettings {
```
Shinigami92, line 43:
```suggestion
export interface NativeSettings {
```

modules/tauri/runtime/native-settings.ts
```diff
@@ -0,0 +1,47 @@
...
+  minimizeToTray: TauriStoreRef<boolean>
+}
+export type { NativeSettings }
```
Shinigami92, line 46:
```suggestion
```

modules/tauri/runtime/native-settings.ts
```diff
@@ -0,0 +1,47 @@
...
+
+async function nativeSettings() {
+  return {
+    minimizeToTray: await tauriStoreRef(store, 'minimize_to_tray', true),
+  }
+}
+
+interface NativeSettings {
+  minimizeToTray: TauriStoreRef<boolean>
+}
```
Shinigami92, line 45: IMO you should move this above `nativeSettings()`

modules/tauri/runtime/native-settings.ts
```diff
@@ -0,0 +1,47 @@
...
+export type { NativeSettings }
+export const useNativeSettings = async () => await nativeSettings() as NativeSettings | undefined
```
Shinigami92, line 47:
```suggestion
export { useNativeSettings: nativeSettings };
```

not sure if `:` or `as`

Shinigami92 (commented):
I could just do a code review
I don't use the native tauri app, you should ask someone else that is more familiar with the tauri app

marekvospel:
> I could just do a code review I don't use the native tauri app, you should ask someone else that is more familiar with the tauri app

@JonasKruckenberg could you please take a look at this when you're free?

modules/tauri/index.ts
```diff
@@ -9,8 +9,10 @@ export default defineNuxtModule({
     const nuxt = useNuxt()
     const { resolve } = createResolver(import.meta.url)
 
-    if (!process.env.TAURI_PLATFORM)
+    if (!process.env.TAURI_PLATFORM) {
+      addImports({ name: 'useNativeSettings', from: resolve('./runtime/native-settings-undefined') })
```
Shinigami92, line 13: Could you please explain why it is needed to have this `-undefined` version?
Cant this be e.g. inlined like `addImports({ name: 'useNativeSettings', from: () => undefined })` or similar?

modules/tauri/runtime/native-settings.ts
```diff
@@ -0,0 +1,45 @@
...
+        await store.set(key, value)
+        await store.save()
+        refValue = value
+        trigger()
+      },
+    }
+  })
+
+  const result = await store.get(key)
+  if (result !== null)
```
Shinigami92, line 31: is it correct here that undefined is not covered?

modules/tauri/index.ts
marekvospel: I didn't find any way to add an import inline, the import object requires `from: string` value, and I couldn't find any other auto import releated function, that would do such a thing.

marekvospel:
Closing, as elk-native is no longer actively developed.

PR #1420, opened 2023-01-24
feat: allow select a language by keyboard in the setting by prepending English language names

shuuji3:
## Problem
Currently, the language selector in the setting doesn't allow us to select a specific item by typing the language name if the name has no alphabet characters. So we have to click the item by mouse or use the arrow keys several times.

## Solution
This PR adds the English name for each language item in the selector so that users can select the specific language by just typing a few first characters of its English word. For example, "Japanese" can be selected by typing only one character `j`. Previously, it required 8 times `↓` keystrokes (open menu + 7 times down).

Screenshot 2023-01-24 at 23 45 11
Screenshot 2023-01-24 at 23 42 42

### DIFF
--- a/components/settings/SettingsLanguage.vue
+++ b/components/settings/SettingsLanguage.vue
@@ -1,16 +1,22 @@
 <script lang="ts" setup>
 import type { ComputedRef } from 'vue'
+import ISO6391 from 'iso-639-1'
 import type { LocaleObject } from '#i18n'
 
 const userSettings = useUserSettings()
 
 const { locales } = useI18n() as { locales: ComputedRef<LocaleObject[]> }
+
+// Find English name from language code (e.g. 'ja-JP' -> 'Japanese')
+function getEnglishName(langCode: string): string {
+  return ISO6391.getName(langCode.split('-')[0])
+}
 </script>
 
 <template>
   <select v-model="userSettings.language">
     <option v-for="item in locales" :key="item.code" :value="item.code" :selected="userSettings.language === item.code">
-      {{ item.name }}
+      {{ getEnglishName(item.code) }} - {{ item.name }}
     </option>
   </select>
 </template>


shuuji3:
Here's Language setting page: https://deploy-preview-1420--elk-zone.netlify.app/settings/language

components/settings/SettingsLanguage.vue
```diff
@@ -1,16 +1,22 @@
...
+import ISO6391 from 'iso-639-1'
 import type { LocaleObject } from '#i18n'
 
 const userSettings = useUserSettings()
 
 const { locales } = useI18n() as { locales: ComputedRef<LocaleObject[]> }
+
+// Find English name from language code (e.g. 'ja-JP' -> 'Japanese')
+function getEnglishName(langCode: string): string {
+  return ISO6391.getName(langCode.split('-')[0])
```
shuuji3, line 12: The `ISO6391` library has language code data like `en` instead of `en-US` so I had to use the first part of the given language code here.

userquin:
@shuuji3 IIRC you cannot do that, you're mixing languages in the html, you can use the arrow up/down to move from one to another

For example, I'm from Spain, I cannot find `Español` using `e` letter,  I need to use `s` (Spanish)... 
Since Elk will try to select best match you're breaking Elk for ppl not speaking english...

shuuji3:
@userquin Thank you for your feedback.🙂 Maybe we can add English names only to the language items that have difficulties navigating by keyboard. At the same time, I'm a bit concerned that it might cause an inconsistent presentation of the menu items.

If we add it to only the selected language, I'm not sure how to select which languages should have additional English names since I'm not familiar with users' ordinary input methods except Japanese and English (I guess Arabic, Chinese, Ukraine, and Japanese would be candidates?)

shuuji3:
I couldn't come up with a good solution for this for a while. But maybe we can replace the language selector with the component used in the compose panel, which can accept typing both the original language name and the English name.

Screenshot 2023-07-23 at 16 27 59

PR #1424, opened 2023-01-24
feat: add support for arm64 based docker images

top:
1. use vars as image name instead of fixed one
2. add support for arm64 platform based docker images


tachyons:
There is a significant increase in CI time due to emulation, 3 minutes to 27 minutes

sxzz:
Thanks for your contribution anyway! Would you like to switch to vars and create another PR for it? :)

Issue #952, opened 2023-01-10
Let opt-out usage of inclusive syntax


edimitchel:
Exclusive to french (based on what I know), I know lot of people doesn't like the usage of dot-syntax in french.
For several reasons: not officially supported by the French Academia, heavy to read and make sentence longer.

So, as Twitter doesn't practice dot-syntax, we could in a sense follow to encourage more the migration and not force to use that syntax.
Also, I know Mastodon instances is using intensively that syntax.

I purpose to add a new wellness option, or a separated language (french without dot-syntax), or an option only for language which could have dot-syntax like difference.

edimitchel:
I can understand your position and I agree.
However, I would like to add that the French dot-syntax is not welcomed to everyone, by default we would enable this syntax, and we won't encourage to disable it. For UX purpose, a few people could prefer to have lighter text without that syntax (without taking attention to the masculine form).
That's why I suggest to let this option as an opt-out option because by default we encourage people to use it.
The day the French academia is officially publish the French inclusive notation, yes, we could remove that because it will be globalized to anyone, not only to people praticing it daily.

Ember-ruby:
this just seems like more work just to be less inclusive, which is, er, *not great*

i get your position, but languages aren't governed by an org, do english speaking people only use language that's in (e.g) the oxford dictionary? no. as that's not how language works.

patak-dev:
There were some discussions in Discord and in other issues here that we should take into account. While I agree that using inclusive language is better, we need to see how this affects screen readers and a11y in general. Yesterday I discussed with someone that pointed out that some folks (on the Austistic spectrum for example) had real issues parsing the text when characters like `.` or `@` where used. I think it would be good to get the opinion of some experts on this matter to see what is the best default on the a11y while also promoting a well-intended switch in the way we use our languages.

kerunix:
I completely agree that this can affect screen readers and a11y in general. I found a few ressources (mostly in french since this is a french only issue as far as I know) that give other options to have a better and more inclusive use of the language while maintaining good a11y and screen readers compatibility. Most of those resources recommend that the way we write our text takes into account both genders.

When it comes to nouns, instead of using "développeurs" (masculine for developer), the dot syntax recommends using "développeur·euses" which is more inclusive but less supported by screen readers and can also make the text less readable for folks with dyslexia or on the autistic spectrum. The recommended way then becomes "développeurs et développeuses", which is both more inclusive AND supported by screen readers.

[...]

PR #1434, opened 2023-01-25
fix: French translation revision

fenarinarsa:
Replaced bi-gendered terms with more neutral terms or non-gendered grammar when possible to make the translation lighter and limiting middle dots.
Following the recommendations of the _Office québécois de la langue française_ for non-sexist language.


florian-lefebvre (approved):
Great idea! I think we could solve #952 this way

locales/fr-FR.json
```diff
@@ -321,7 +321,7 @@
       "push_notifications": {
         "alerts": {
           "favourite": "Messages aimés",
-          "follow": "Nouveaux abonné·e·s",
+          "follow": "Nouveaux abonnés",
```
prazdevs, line 324: This one isn't gender neutral

prazdevs (commented):
There has been some talks regarding this and how we could handle inclusivity with such a binary-gendered language French is. 
It is still a big subject as for i18n. 
Some wordings take away the "human" aspect and i'm not really into them personnally.

locales/fr-FR.json
fenarinarsa: I know using "account" instead of "user" isn't a 100% satisfying solution but French being what it is, it's the closest translation we can get without using the middle dot, which is not a 100% satisfying solution either for big words in UIs (in my opinion). As I said I'm using the official Québec recommendation which advise first to use words that have defined genders or not gendered but can describe the whole group of persons, and use the middle dot if it's not possible.
But it's also closer to the reality, since an account does not exactly equals user on social networks. A Fediverse account can have no user behind it (bot) or many users (company). So in the end it's not that annoying.

prazdevs: Sounds like the best compromise anyways for inclusivity 👍

prazdevs (approved):

edimitchel (approved):

patak-dev:
Let's merge this on the basis of better a11y, how to properly move forward on this topic is still open for many languages and it would be good to keep discussing.

PR #1439, opened 2023-01-25
feat: hide reply count

### LINKED ISSUE #1436: Hide/condense reply count
This is a thing Mastodon intentionally does. When a post has more than 1 reply, it displays "1+" instead of the count, with the intention of having people not immediately judge a post on the amount of feedback it received. It would be nice to have a similar option in Elk.
image

boehs:
closes #1436


boehs:
awaiting discussion in #1436

boehs:
@antfu ready for review :)

antfu (approved):

PR #1445, opened 2023-01-25
feat: add background color change on status card hover

### LINKED ISSUE #1241: "cursor: pointer" is set to only text of a status instead of the whole card
### Pre-Checks

### Describe the bug

At the moment, only the content (i.e., the text) of a status has the following CSS:
```css
.status-body.with-action p {
    cursor: pointer;
}
``` 

However, as the UX is currently, it is not just the textual section of a status, but the whole status card that reacts to a click, navigating to the status detail page. Thus, I believe that the whole status card should have the aforementioned CSS, instead of solely the body.

As an extra UI clue that could be beneficial to the UX, a change in background color could be added to the status card on hover, similar to how hover works on notification cards at the moment.

https://user-images.githubusercontent.com/2052855/212788271-163a6446-698f-4f9b-a4eb-693d7c65a867.mov

### Additional Context

_No response_

Mini-ghost:
resolves #1241

When hover on `<StatusCard>`

**Before**
before

**After**

after

### DIFF
--- a/components/status/StatusBody.vue
+++ b/components/status/StatusBody.vue
@@ -48,9 +48,3 @@ const vnode = $computed(() => {
     </template>
   </div>
 </template>
-
-<style>
-.status-body.with-action p {
-  cursor: pointer;
-}
-</style>
--- a/components/status/StatusCard.vue
+++ b/components/status/StatusCard.vue
@@ -84,8 +84,7 @@ const showReplyTo = $computed(() => !replyToMain && !directReply)
     tabindex="0"
     focus:outline-none focus-visible:ring="2 primary"
     cursor-pointer
-    hover="bg-black/3"
-    dark:hover="bg-white/3"
+    hover:bg-active transition-100
     :lang="status.language ?? undefined"
     @click="onclick"
     @keydown.enter="onclick"
--- a/components/status/StatusCard.vue
+++ b/components/status/StatusCard.vue
@@ -83,6 +83,9 @@ const showReplyTo = $computed(() => !replyToMain && !directReply)
     :class="{ 'hover:bg-active': hover }"
     tabindex="0"
     focus:outline-none focus-visible:ring="2 primary"
+    cursor-pointer
+    hover="bg-black/3"
+    dark:hover="bg-white/3"
     :lang="status.language ?? undefined"
     @click="onclick"
     @keydown.enter="onclick"


Mini-ghost:
@patak-dev Could I trouble you to review this PR for me? Your help would be much appreciated.

patak-dev:
Thanks for the PR @Mini-ghost. I think we could add this but the effect should be more subtle. If you see how Twitter works, it is a very small color diff when you hover. I think it is too distracting if not while reading your home feed. The Mastodon app I think has no hover diff at all.

We also have some glitches if we do this

1. The show more here
2. The solid circle around the avatar
image
3. The margins in conversations (I think we could remove them)
image

Maybe an option is to make this PR only about the cursor, and then in another PR in the future the hover color diff could be tried out

Mini-ghost:
@patak-dev Thank you for your review, I put together the following three points to make sure I have grasped your suggestion:

1. Reduce the color difference when hovering (to make it more subtle and less distracting).
2. Address the color issue related to "The show more here" element.
3. Address the color issue related to "The solid circle around the avatar."

I worked on the first two problems, and I found solutions for them. However, when trying to address the third problem, my initial approach using `group` and `elk-group-hover` in the outer layer caused issues with the display of `<StatusActionButton />`. Now, I'm unsure about how to resolve this problem.

patak-dev:
It is better now, but I think the color change should be even more subtle. Check Twitter to see how small the diff when hovering is.

There is one extra point, there is an extra top/botton margin on Conversations when you hover: 
image

About the avatars, if the color diff is very subtle, then probably isn't that noticeable and we could leave it as is for now

Mini-ghost:
The previous corrected version mentions the transparency settings used by Twitter:

- Default: `rgba(0, 0, 0, 0.03)`
- Dim: `rgba(255, 255, 255, 0.03)`

I decreased the opacity to `0.02`. Is this improvement better?

**Twitter**
截圖 2023-07-29 下午8 14 54

**Elk previous version (This PR)**
截圖 2023-07-29 下午8 32 30

**Elk current version (This PR)**
截圖 2023-07-29 下午8 33 39

patak-dev:
The color diff is still higher than in Twitter:

image

The effect may be even bigger because of the lack of borders

image

Maybe even more subtle. Or we can merge the PR without any color diff on hover for the main timelines for now.

Mini-ghost:
Oh, yes! The color difference is more noticeable in elk's dark mode compared to Twitter. In "lights out" mode on Twitter, the background color is `rgb(0, 0, 0)` (black), whereas elk's dark mode uses `rgb(17, 17, 17)` for the background. Even with the same hover background setting, the overall feeling will be different between the two. 

Personally, I find it hard to see any difference when hover on post in Twitter's "lights out" mode 🥲.

PR #1446, opened 2023-01-25
fix: default user language when writing a publication

### LINKED ISSUE #1441: When publicating, language is always English (not considering language set on settings)
Hello!
When creating a publication (new or a reply). The default language is English even if, for example, I have in my settings that my default language is Spanish.
So, I have to change in every publication the language of the publication to spanish.

victordelva:
fixes #1441


userquin:
@victordelva can you solve it for any language, not only for spanish? split the language with `-` then get the first entry, we're using only language part in the picker.

userquin:
@victordelva it is failiing due to an internal error, you don't need to try to fix it, sometimes works...

patak-dev (approved):

patak-dev:
Gracias Victor!

PR #1455, opened 2023-01-25
chore: upgrade to Nuxt 3.1

antfu:


danielroe:
As we use 3rd-party components, we will probably need 3.1.1 - if you can test nuxt edge to see if there is any other bug, that would be good to know.

antfu:
/cc @danielroe Are you aware of this runtime error? 
https://deploy-preview-1455--elk-zone.netlify.app/home

Screenshot 2023-01-26 at 10 31 27

danielroe:
I am. It should be set in the entry but seems maybe a race condition that it hasn't evaluated by the time individual asset chunk has loaded.

PR #1456, opened 2023-01-25
fix(pwa): add maskable icon to web manifest

### LINKED ISSUE #1449: PWA is not fully configured for Android
On Android, if a progressive web app is not fully and properly configured, it can still be installed via a prompt, but it will show as a web shortcut, and will include the icon of the browser that installed it alongside the app icon (rather than just the app icon by itself). It will also not show or be searchable in the list of apps on its own; it will just functionally be a web shortcut in the device's default browser.

I saw a lot of this while installing Wordle-like games last year. Take a look at my home screen; notice how Elk (circled), along with most of the Wordle-like games (slashed through), has the Edge icon along with it.

Screenshot_20230125-081214~2

In some cases that's because the app isn't actually a PWA. For others, it's because it is a PWA but not properly configured. As you can see, Android makes no distinction between the two.

However, the two icons in the lower right (Quina and Hondo, checked) are _also_ both progressive web apps, and neither installs as a shortcut or with the Edge logo, because both of those projects generate the proper files and manifest entries. (They're also both my apps, for full transparency, which is how I know this.)

If I recall correctly, Android expects to find certain manifest entries, and images with a specific name and size, in proper PWA packages. This may help. https://github.com/gokulkrishh/awesome-meta-and-manifest

[Edit: screenshot from a Pixel 4a on Android 13, though I don't thiink device or OS version are relevant in this case.]

userquin:
Latest changes on Chromium based browsers complains about not having a maskable icon (LightHouse): just added the variants adding 10px padding to current `apple-touch-icon.png` icons.

Tested on Android from Chrome:

Screenshot_20230125_181456_One UI Home

I need someone to check it on iOS...

/cc @antfu sorry man, can you try installing preview and check the icon is ok? 

/cc @josh-collinsworth can you test preview PR on Edge Mobile?

closes #1449


josh-collinsworth:
@userquin Know what? I'm sorry, I'm wrong. I think the "bug" I reported is actually a normal effect from using non-Chrome browsers to install PWAs on Android, and not anything to do with the PWA itself.

I must have used Chrome to install the ones that worked properly, and used Edge to install the others, and completely forgotten.

So yeah, I can confirm that it installs properly when using Chrome as the installing user agent, and that it does _not_ work with Edge—but, again, this is expected.

Still, the maskable icon is a win, though I'd recommend it be higher resolution. (Here's what the PWA loading screen looked like before—all black—and what it looks like with this PR—the round, lower-res icon.)

*Before:*
Screenshot_20230125-113813

*After:*
Screenshot_20230125-113425

josh-collinsworth:
Unless there's time and energy to spend on making the maskable icon high-res, I'd suggest maybe this issue and PR should just be closed, since the default seemed to work well for me. (Again, I apologize for sending you on a wild goose chase with this report.)

userquin:
@josh-collinsworth thx for feedback, I'll add a 512x512 maskable icon, I need to create and optimize them...

userquin:
> Unless there's time and energy to spend on making the maskable icon high-res, I'd suggest maybe this issue and PR should just be closed, since the default seemed to work well for me. (Again, I apologize for sending you on a wild goose chase with this report.)

can you test again?

josh-collinsworth:
Looks much better, resolution-wise. It _does_ still have the light-colored circle background. If that's the desired effect, I'd say 👍

userquin:
> Looks much better, resolution-wise. It _does_ still have the light-colored circle background. If that's the desired effect, I'd say 👍



> Looks much better, resolution-wise. It _does_ still have the light-colored circle background. If that's the desired effect, I'd say 👍

can you test again? maskable icons sizes were wrong (180px, updated to 512px)

josh-collinsworth:
Still looks good on my end.

patak-dev (approved):

patak-dev:
Awesome!

dunxen:
Sorry to draw attention to this again, but it seems on the iOS 16.4 Beta (which I'm running now) it seems it prefers the PWA icons (since they're supporting a lot more on the manifest side of things also push which has been working for me with Elk 🎉).

This is how the icon looks for me now:
signal-2023-02-22-133235

I can open another issue if one does not currently exist.

userquin:
@dunxen so, it seems iOS now using 512x512 pwa icon instead maskable one and ignoring `apple-touch-icon`?

dunxen:
> @dunxen so, it seems iOS now using 512x512 pwa icon instead maskable one and ignoring `apple-touch-icon`?

Yes looks like that!

userquin:
@dunxen FYI: https://twitter.com/firt/status/1628378241726545921

dunxen:
> @dunxen FYI: https://twitter.com/firt/status/1628378241726545921

Thanks! Seems to be working on 16.4 Dev Beta 2:

image

PR #1469, opened 2023-01-26
feat: add bold and italic buttons to editor

victordelva:
imagen
imagen


patak-dev:
This looks good on big screens, but I'm afraid we are already out of space for mobile. @userquin also was working on some advanced features that would require space. And we got a request to hide the code block button. I think we should already have these two new ones and the code block behind a new format button that opens a toolbar on hover maybe?

victordelva:
@patak-dev 
I'll try to work on that

victordelva:
I used a dropdown. this is how it looks on smaller screens. 
What do you think @patak-dev ?

imagen

boehs (approved):

patak-dev:
It looks like there is a glitch in mobile, the keyboard collapses when the new button is pressed. Maybe is normal but we coupd check if that could be avoided

boehs:
I think a simple e.preventDefault fixes

boehs:
closes #1469

PR #1474, opened 2023-01-26
feat: initial support for lists

boehs:
#372

Issue #372, opened 2022-12-08
feat: user lists


ayoayco:
Implement creating / showing lists on Elk.

The Mastodon web client has this feature:

![Screenshot 2022-12-08 at 9 56 59 AM](https://user-images.githubusercontent.com/4262489/206403174-58d02e30-e80b-45ee-9ac4-6734205f4c4d.png)

maegul:
In line with @jmarkel 's [comment](https://github.com/elk-zone/elk/issues/372#issuecomment-1368246424), I too am interested!  I like what I've seen of elk a lot so far, but for me the lists is kind of a must have and I'll stay with the default webview for now entirely because of the lack of lists in elk .

Generally though, this feels like a nice opportunity to innovate on the filtering the fire-hose front on the client side, to whatever extent that's feasible so I'll be keeping an eye on elk for sure!

ayoayco:
@maegul these are good ideas! We've had some discussions lately on "pinning" public timelines, I guess hashtags could be pinned too.

Are you able to [create separate issues](https://github.com/elk-zone/elk/issues/new/choose) so we keep the scope of this one to user lists?

kerim:
I would like to replace "home" with a list of my choice, as well as have a couple of other pinned lists that I check regularly. I would also like to be able to add people to lists when I follow them.

kissu:
Even Twitter has that one nowadays, more friendly than ever. 😄 
<img width="610" alt="CleanShot 2023-01-18 at 18 53 06@2x" src="https://user-images.githubusercontent.com/5133074/213257457-39422aa9-7f43-4e88-b517-48a7ee76f472.png">

gerwitz:
+1 for @kerim's scenario. I follow promiscuously but then use a "regulars" list for my daily timeline browse.

LauraLangdon:
It would be super helpful to be able to add a user to a list directly from that user’s profile!

keelancook:
I'll chime in here as well. Love the look of Elk. Can't use it until lists are in place. They are 100% how I organize all of my followers. Mastodon is too noisy otherwise. 

My ideal use case would be an app that allows me to completely edit the timeline options on the side. Allow me to choose whether or not I show "home" or "bookmarks" or "favorites" and give me the option to replace these with my own lists, or followed hashtags. That would speed up my use of Mastodon tenfold. 

[...]

PR #1479, opened 2023-01-26
feat: more list support

boehs:
#372 

- [x] lists (#1474)
- [x] view accounts in list
- [x] add accounts to list
  - [x] from account
  - [ ] ~~from list page~~ (not this pr)
- [x] remove accounts from list
  - [x] from account
  - [x] from list page
- [ ] change list name
- [ ] create list
- [ ] delete list

known bugs:

- [x] listid not reactive
- [x] bad styling for account "add to list". Possibly bad accessibility as well.

image
image
image

leaving current unchecked features for future prs


boehs:
@patak-dev review time ❤️

patak-dev:
Nice! 

I think the tabs are great for now, and also the icon to remove from the list.

Do you need to follow a user to be able to have it in a list? I see the button only appears when you follow someone. I imagined lists were also a way to see posts from certain people without following them 🤔 

About the dropdown when pressing the list button on the profile, lets remove the colors and add a checked icon when the user is in that list. I think the current red/green may confuse users. Later on, maybe we should show all the lists the user is part of first, and then the rest as a separate list below, and animate the transition.

boehs:
> Do you need to follow a user to be able to have it in a list? I see the button only appears when you follow someone.

As much as it pains me, I believe so. Every client I've looked at only allows it when following, and the [API Docs](https://docs.joinmastodon.org/methods/lists/#404-not-found-3) are grim.

> lets remove the colors and add a checked icon when the user is in that list.

Sounds good. Unfortunately, remix icons doesn't have a "playlist remove" icon, only the add icon. Other possibilities:

- Normal +/- icons
- User+/User- icons (like on the list accounts page)

Neither are very good, though I'm much more inclined to the second. Curious to know your thoughts or other ideas.

patak-dev:
I think the second option could work until we find something better

boehs:
@patak-dev fixed :)

boehs:
actually don't merge yet, I found a better icon

boehs:
alright, should be good now

patak-dev:
Awesome! Lets merge it to get the functionality. Maybe we could later have the lists the user is in as primary color. Or maybe only show the lists the user is part of, and then a add to list button.

PR #1481, opened 2023-01-27
fix: prevent click event when publish button is disabled

### LINKED ISSUE #1480: prevent click event when pubish button is disabled
CPT2301271417-639x277
It looks weird now. If publish button is disabled, it should not trigger any event or network request.

sechi747:
closes #1480

### DIFF
--- a/composables/masto/publish.ts
+++ b/composables/masto/publish.ts
@@ -29,6 +29,8 @@ export const usePublish = (options: {
   }, { deep: true })
 
   async function publishDraft() {
+    if (isPublishDisabled)
+      return
     let content = htmlToText(draft.params.status || '')
     if (draft.mentions?.length)
       content = `${draft.mentions.map(i => `@${i}`).join(' ')} ${content}`


ushuz:
Maybe add `disabled="isPublishedDisabled"` to the publish button as well?

https://github.com/elk-zone/elk/blob/1b4202d6a6994a1ab5b7319c46078bb22ebe12a0/components/publish/PublishWidget.vue#L324-L331

sechi747:
> 

I try it, but it'll make tooltip disappear and the cursor will change to `default`

components/publish/PublishWidget.vue
```diff
@@ -97,6 +97,8 @@ async function toggleSensitive() {
 }
 
 async function publish() {
+  if (isPublishDisabled)
```
userquin, line 100: can you move this check to `publishDraft` on `publish.ts` module'? rn only used here...

sechi747: > can you move this check to `publishDraft` on `publish.ts` module? rn only used here...

sure, I have resolved it.

PR #1379, opened 2023-01-21
feat: add finnish locale

danielroe:
This is courtesy of [Martin Vermeer](https://fediscience.org/@martinvermeer). Thank you! ❤️

PR #1487, opened 2023-01-27
Revised Finnish localization

osmaa:
ref #1379 which I found to be of pretty bad quality (sorry to original author)


osmaa:
apologies, I missed perhaps the most visible item, the language name itself
https://github.com/elk-zone/elk/commit/53b3670b2c741275f729a33fcd37b214c079fad2

PR #1492, opened 2023-01-27
fix(native): Fixed link url in native app

piotrski:
This should fix elk-zone/elk-native/issues/37

### DIFF
--- a/components/status/StatusActionsMore.vue
+++ b/components/status/StatusActionsMore.vue
@@ -31,8 +31,9 @@ const { client } = $(useMasto())
 
 const getPermalinkUrl = (status: mastodon.v1.Status) => {
   const url = getStatusPermalinkRoute(status)
+
   if (url)
-    return `${location.origin}/${url}`
+    return !import.meta.env.TAURI_PLATFORM ? `${location.origin}/${url}` : `https://elk.zone/${url}`
   return null
 }
 


components/status/StatusActionsMore.vue
```diff
@@ -1,4 +1,6 @@
 <script setup lang="ts">
+import { env } from 'process'
+import { metaProperty } from '@babel/types'
```
patak-dev, line 3: Are these imports needed?

piotrski: No, it was just ts trying to be too clever. I've already removed it.

danielroe (changes requested):
I can see we might have situations like this throughout the app. Maybe a feature we can toggle on in the tauri module that forces the `elk.zone` hostname so we don't have to reference tauri here?

piotrski:
@danielroe Because it's just a browser behavior, and Tauri has no control over it. I guess the only thing we could do is to monkey patch `location.origin` in the native bundle so that it always returns `elk.zone` but I'm not a fan of that solution.

danielroe:
I understand the need and agree. I'm suggesting that rather than check the environment variable here, we instead have a global setting (maybe in app config) that changes behaviour across _everywhere_ in the app that accesses `location.origin`.

PR #1499, opened 2023-01-28
fix(ui): statusCard style in RTL

zyyv:
Fix the reply timeline right position.

| Before | After |
| -- | -- |
|  image | image |


components/status/StatusCard.vue
```diff
@@ -95,12 +96,13 @@ const showReplyTo = $computed(() => !replyToMain && !directReply)
...
           v-if="showReplyTo"
-          ml-20px pt-1 pl-5
+          m="l-5" rtl-m="l-0 r-5"
+          p="t-1 l-5" rtl-p="l-0 r-5"
           :status="status"
           :is-self-reply="isSelfReply"
           :class="faded ? 'text-secondary-light' : ''"
         />
-        <div flex="~ col gap-1" items-center pos="absolute top-0 left-0" w="77px" z--1>
+        <div flex="~ col gap-1" items-center pos="absolute top-0 left-0" rtl="pos-left-auto right-0" w="77px" z--1>
```
userquin, line 105: From contributing guide:

For absolute positioned elements, don't use `left/right`: for example `left-0`. Use `inset-inline-start/end` instead. `UnoCSS` shortcuts are `inset-is` for `inset-inline-start` and `inset-ie` for `inset-inline-end`. Example: `left-0` should be replaced with `inset-is-0`.

components/status/StatusCard.vue
```diff
@@ -95,12 +96,13 @@ const showReplyTo = $computed(() => !replyToMain && !directReply)
       <template v-if="status.inReplyToAccountId">
         <StatusReplyingTo
           v-if="showReplyTo"
-          ml-20px pt-1 pl-5
+          m="l-5" rtl-m="l-0 r-5"
+          p="t-1 l-5" rtl-p="l-0 r-5"
```
userquin, line 100: From contributing guide:

- Do not use `left/right` padding and margin: for example `pl-1`. Use `padding-inline-start/end` instead. So `pl-1` should be `ps-1`, `pr-1` should be `pe-1`. Same rules applies for margin.
- Do not use `rtl-` classes, such as `rtl-left-0`.

components/status/StatusCard.vue
```diff
@@ -79,7 +79,8 @@ const showReplyTo = $computed(() => !replyToMain && !directReply)
   <div
     :id="`status-${status.id}`"
     ref="el"
-    relative flex="~ col gap1" p="l-3 r-4 b-2"
+    relative flex="~ col gap1"
+    p="l-3 r-4 b-2" rtl-p="l-4 r-3"
```
userquin, line 83: this should also be fixed

zyyv: Thanks so much, please forgive my mistake😅

userquin (approved):
nice

PR #1501, opened 2023-01-28
fix(a11y): do not prevent arbitrary zooming on mobile devices

### LINKED ISSUE #1507: Consider restoring browser zoom
## Summary

Zoom is currently disabled in Elk — i.e., native pinch gestures do not allow a user to zoom in on any part of the Elk UI.

## Additional Details

Adrian Roselli [succinctly captures](https://adrianroselli.com/2015/10/dont-disable-zoom.html) many reasons not to disable zoom, many of which may apply to Elk users.

For my personal browsing needs and habits I find the inability to pinch-zoom images a frequent frustration, which Adrian explains:

> **The user may want to see more detail in an image.**
> Especially with images that are scaled down to fit in the viewport, compounded when they include text.

I hope you'll consider restoring this functionality in a future update.

Thanks!

### Configuration

* **Elk Version** —  v0.6.2
* **Browser** — Safari
* **OS** — iOS 16.2

## Workaround

None — while browser zoom can increase Elk's text size, it has no impact on images.

## Personal Notes

Thank you very much for building such a robust, cozy interface! 💫 It's been a real breath of fresh air, and it quickly became the only app I use to hang out on Mastodon ☺️

ronilaukkarinen:
Noticed cannot zoom on mobile devices at all because it's completely disabled on meta viewport. I don't understand the reasoning behind this. I know there is a font-size scale setting but it's not sufficient for many users with the problems on eye-sight. As Elk works as PWA, zooming should always be enabled.

## Why it Matters

The `maximum-scale` parameter limits the amount the user can zoom. This is problematic for people with low vision who rely on screen magnifiers to properly see the contents of a web page.

Users with partial vision and low vision often choose to enlarge the fonts or other elements on their browser to make text and images on the web easier to read and recognize. The browser's viewport focus is everything visible in the browser window at a given moment. Maximizing the browser to full size on a high-resolution monitor creates a large the viewport focus area and may include the entire web page.

If the browser window is small, the viewport focus area only includes a small part of the web page. The browser's viewport focus does not affect the programmatic focus. Users can scroll up and down the web page, but the programmatic focus does not follow the viewport. The Web Content Accessibility Guidelines calls for developers to design pages so that they support resize up to 200%.

Please always ensure that the `user-scalable="no"` parameter is not present in the `<meta name="viewport">` element and the `maximum-scale` parameter is not less than 2 or not defined at all to allow scaling as user pleases.

The overflow-x: hidden is present so let's at least give user a choice to zoom. Thank you.

closes #1507

### DIFF
--- a/nuxt.config.ts
+++ b/nuxt.config.ts
@@ -130,8 +130,7 @@ export default defineNuxtConfig({
   app: {
     keepalive: true,
     head: {
-      // Prevent arbitrary zooming on mobile devices
-      viewport: 'width=device-width,initial-scale=1,maximum-scale=1,user-scalable=0,viewport-fit=cover',
+      viewport: 'width=device-width,initial-scale=1',
       bodyAttrs: {
         class: 'overflow-x-hidden',
       },


ronilaukkarinen:
On the side note, I don't have a dev build running and Nuxt is currently not a part of any of my workflows, I just noticed this on the production environment. Tried to search but didn't see other references to a viewport, I hope this change is enough. If not or it's plain wrong, let's revise this. I wanted to point out the issue via PR to underline it a bit more.

userquin:
@ronilaukkarinen the change was included here https://github.com/elk-zone/elk/pull/234, I don't know why, but I use a lot pinch to zoom on mobile

ronilaukkarinen:
> @ronilaukkarinen the change was included here #234, I don't know why, but I use a lot pinch to zoom on mobile

Right. Thanks, tried to look for it but didn't find it. No explanation is provided. My wife is deaf and blind (she sees a bit but uses pinch to zoom A LOT) and I base for her way of using the Internet to all my decisions. And we should do decisions to the user experience via the weakest amongst us.

In my experience it is REALLY hard to "accidentally" pinch and zoom. I really think this change should be reverted back.

joelanman:
https://adrianroselli.com/2015/10/dont-disable-zoom.html

Shinigami92 (approved):
I tested it on my iPhone, but only like every 50 try I could pinch zoom
For me this is okay right now

userquin (approved):
I spoke to @Shinigami92, we should fix all accessibility issues and this should be reverted.

PR #1506, opened 2023-01-28
fix: use "⌘" as modifier key name on macOS

shuuji3:
This replaces the wrong "Ctrl" with "Command" as a modifier key name in the command pallet on macOS.

**Before**
Screenshot 2023-01-29 at 2 55 46

**After**
Screenshot 2023-01-29 at 2 55 55

I used `navigator.userAgentData.platform` to detect the platform name. However, this is implemented only in Chromium-based browsers at this moment.

ref. ["userAgentData" | Can I use... Support tables for HTML5, CSS3, etc](https://caniuse.com/?search=userAgentData)

If we want to work this on any browser, we need to use another way until all major browsers implement `userAgentData`. Then, does anyone know the best way to reliably detect the macOS in the Vue ecosystem?

### DIFF
--- a/components/command/CommandPanel.vue
+++ b/components/command/CommandPanel.vue
@@ -64,6 +64,8 @@ const result = $computed<QueryResult>(() => commandMode
   : searchResult,
 )
 
+const modifierKeyName = $computed(() => navigator.userAgentData?.platform === 'macOS' ? 'Command' : 'Ctrl')
+
 let active = $ref(0)
 watch($$(result), (n, o) => {
   if (n.length !== o.length || !n.items.every((i, idx) => i === o.items[idx]))
@@ -233,8 +235,8 @@ const onKeyDown = (e: KeyboardEvent) => {
     <!-- Footer -->
     <div class="flex items-center px-3 py-1 text-xs">
       <div i-ri:lightbulb-flash-line /> Tip: Use
-      <CommandKey name="Ctrl+K" /> to search,
-      <CommandKey name="Ctrl+/" /> to activate command mode.
+      <CommandKey :name="`${modifierKeyName}+K`" /> to search,
+      <CommandKey :name="`${modifierKeyName}+/`" /> to activate command mode.
     </div>
   </div>
 </template>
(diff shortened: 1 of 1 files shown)


shuuji3:
It seems like TypeScript doesn't allow to use newer attribute `userAgentData`.

userquin:
@shuuji3 you can use `useIsMac`, you can find it on `composables/misc.ts` module

shuuji3:
@userquin Thanks, that's exactly what I looked for and somehow missed finding.

components/command/CommandPanel.vue
```diff
@@ -64,6 +65,8 @@ const result = $computed<QueryResult>(() => commandMode
   : searchResult,
 )
 
+const modifierKeyName = $computed(() => useIsMac() ? 'Command' : 'Ctrl')
```
userquin, line 68:
```ts
const isMac = useIsMac()
const modifierKeyName = $computed(() => isMac .value ? 'Command' : 'Ctrl')
```

components/command/CommandPanel.vue
```diff
@@ -1,4 +1,5 @@
 <script setup lang="ts">
+import { useIsMac } from '~/composables/misc'
```
userquin, line 2: you don't need to import the module, we've auto import enabled

shuuji3:
Thanks for reviewing. Applied your suggestion.

components/command/CommandPanel.vue
```diff
@@ -64,6 +64,9 @@ const result = $computed<QueryResult>(() => commandMode
   : searchResult,
 )
 
+const isMac = useIsMac()
+const modifierKeyName = $computed(() => isMac.value ? 'Command' : 'Ctrl')
```
danielroe, line 68:
```suggestion
const modifierKeyName = $computed(() => isMac.value ? '⌘' : 'Ctrl')
```

shuuji3: That's a good idea.🙂  Here's the new screenshot:

Screenshot 2023-01-29 at 15 50 01

shuuji3:
Thanks for merging!

PR #1508, opened 2023-01-28
refactor: move pwa/build-info -> appConfig

danielroe:
This is a slight performance optimisation. We likely don't want to turn on/off pwa being enabled (or update build info) at runtime. So by moving this into app config (which is built into code) we both save the payload and enable the possibility of tree shaking.

This is a precursor to tree-shaking out different unstorage drivers (but this would be a breaking change so will defer to a different PR).


modules/pwa/index.ts
```diff
@@ -23,6 +23,9 @@ export default defineNuxtModule<VitePWANuxtOptions>({
     }
     let webmanifests: LocalizedWebManifest | undefined
 
+    nuxt.options.appConfig = nuxt.options.appConfig || {}
+    nuxt.options.appConfig.pwaEnabled = !nuxt.options.dev || process.env.VITE_DEV_PWA === 'true'
```
userquin, line 27: how about tauri, rn it is disabled

danielroe: This PR doesn't change the current behaviour, just moves the setting from `nuxt.config` -> `pwa/index`. Do you think we need to change the current behaviour - or am I missing something?

userquin: `nuxt.options.appConfig.pwaEnabled = !nuxt.options.dev || process.env.VITE_DEV_PWA === 'true'` that's true also on Tauri

userquin: I mean, `!nuxt.options.dev`

danielroe: Current login is the same: https://github.com/elk-zone/elk/blob/main/nuxt.config.ts#L101

antfu (approved):

danielroe:
One further change coming (discussing with @userquin).

userquin (approved):

PR #1510, opened 2023-01-29
fix: handle account switching when accounts have the same id

angristan:
In the following example, I have 3 accounts:

| Handle                  | Local ID |
| ----------------------- | -------- |
| `angristan@mstdn.io`    | 1        |
| `angristan@kpop.social` | 1        |
| `admin@mstdn.io`        | 5363     |

-> The ID can't be used to identify an account by itself, we need the instance as well.

This PR changes these checks from id to full handle ([the `acct` field](https://github.com/neet/masto.js/blob/392cee4b0f9311da2f04ef45858b2b7864b6828b/src/mastodon/v1/entities/account.ts#L47)).

## Before 

It's not very clear in the video, but basically when I try to switch between accounts with the same ID, this functions just open the profile of the first stored user with that ID:

```ts
const clickUser = (user: UserLogin) => {
  if (user.account.id === currentUser.value?.account.id)
    router.push(getAccountRoute(user.account))
  else
    switchUser(user)
}
```

As a result, I can never switch to my second user with an `id` of `1`.

https://user-images.githubusercontent.com/11699655/215296460-9cfa5262-8d72-438b-b62e-271c1f13385b.mp4

## After

By using `handle@domain`, we can differentiate the accounts correctly:

https://user-images.githubusercontent.com/11699655/215296453-318a39aa-7c84-45ad-8b34-ac49e6729975.mp4


userquin:
@patak-dev user ids are not unique between servers: https://github.com/elk-zone/elk/pull/384#discussion_r1045231186

composables/users.ts
```diff
@@ -101,19 +100,13 @@ if (process.client) {
...
       if (oldUser)
         window.addEventListener('visibilitychange', windowReload, { capture: true })
     }
   }, { immediate: true, flush: 'post' })
-
-  // for injected script to read
-  const currentUserHandle = computed(() => currentUser.value?.account.acct || '')
-  watchEffect(() => {
-    localStorage.setItem(STORAGE_KEY_CURRENT_USER_HANDLE, currentUserHandle.value)
-  })
```
danielroe, line 116: I believe this is still needed for https://github.com/elk-zone/elk/blob/1a577343daabfcb5dd1c45fd1ee630d464a00bb3/plugins/setup-head-script.server.ts#L11-L16 to pick up...

danielroe (changes requested):
This looks great - only that we still need to store the value of the current user handle in localStorage for user settings restoration on reload.

composables/users.ts
angristan: Isn't this (updated) line sufficient?
```ts
const currentUserHandle = useLocalStorage<string>(STORAGE_KEY_CURRENT_USER_HANDLE, mock ? mock.user.account.id : '')
```

When I switch between accounts, it updates the `elk-current-user-handle` local storage value.

danielroe: Oh, you're quite right!

danielroe (approved):

PR #1517, opened 2023-01-29
feat(settings): add pinch to zoom setting

userquin:
This PR also includes translations for spanish and deutch (@Shinigami92).


userquin:
@Shinigami92 should we add also headers for max zoom removed on previous PR?

Shinigami92 (approved):

Shinigami92:
> @Shinigami92 should we add also headers for max zoom removed on previous PR?
> 
> We'll get weird behavior on Safari (it is about your comments in `ui discord channel`)

I cant follow, sorry
What do you mean? Could you provide screenshots or describe more what you mean?

userquin:
It is about weird scroll behaviour you reported this morning in ui channel (scroll left and right)

Shinigami92:
Ah I understand, it's not that important so much right now

danielroe:
Could we disable pinch to zoom UI by default, perhaps?

userquin:
> Could we disable pinch to zoom UI by default, perhaps?

should be accesible by default ;)

aeharding:
> > Could we disable pinch to zoom UI by default, perhaps?
> 
> should be accesible by default ;)

imo should be disabled by default when added to Home Screen, to feel like a native app.

patak-dev:
For reference, the Twitter web app on mobile has pinch to zoom disabled, and the GitHub app has it enabled (although GitHub isn't trying that hard yet on mobile web). I also thought that was normal here to have this disabled (I closed a PR before the one that got merged). I see the point though about this being enabled by default (although if we follow that through, we should probably have a bigger font size by default, use the high-contrast theme by default, and so on). Are there links or other app examples we could use to justify the default to use? (not sure what is the best to be honest)

userquin:
I'll send a new PR with opt-in, you can then choose one of them

Issue #1512, opened 2023-01-29
Character count should include content warning text 


maxcalvin:
The publish/reply/compose area character count doesn't include the content warning characters. So when publishing at max content length with a warning, it is blocked/rejected by the server.

PR #1535, opened 2023-01-31
fix: character count should includes spoiler text

Mini-ghost:
resolves: #1512

In Mastodon, content warning (spoiler text) will be included in the character count

mastodon

But not currently
main

After this PR, when the content warning (spoiler text) is activated, it will be included in the word count, and it will be ignored when it is closed.

https://user-images.githubusercontent.com/39984251/215755701-69f024c9-969f-403f-8fc6-90f535b40079.mp4

### DIFF
--- a/composables/masto/publish.ts
+++ b/composables/masto/publish.ts
@@ -21,6 +21,17 @@ export function usePublish(options: {
   const isExpanded = $ref(false)
   const failedMessages = $ref<string[]>([])
 
+  const publishSpoilerText = $computed({
+    get() {
+      return draft.params.sensitive ? draft.params.spoilerText : ''
+    },
+    set(val) {
+      if (!draft.params.sensitive)
+        return
+      draft.params.spoilerText = val
+    },
+  })
+
   const shouldExpanded = $computed(() => expanded || isExpanded || !isEmpty)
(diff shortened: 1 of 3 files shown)


composables/masto/publish.ts
```diff
@@ -21,6 +21,17 @@ export function usePublish(options: {
   const isExpanded = $ref(false)
   const failedMessages = $ref<string[]>([])
 
+  const publishSpoilerText = $computed({
+    get() {
+      return draft.params.sensitive ? draft.params.spoilerText : ''
+    },
+    set(val) {
+      if (!draft.params.sensitive)
+        return
```
danielroe, line 30: Do we need this?

```suggestion
```

Mini-ghost: No, just to make the logic more rigorous, is it recommended to delete it?

danielroe: I can't really think when this would be called, so I think it probably doesn't matter either way.

danielroe (approved):

composables/masto/publish.ts
Mini-ghost: @danielroe  Got it, Normal operation here would never call. I wish I could remove this if you prefer cleaner code.

PR #1553, opened 2023-02-01
docs: add new elk instances

WhiskeyOmega:
added freelancers.online and cupoftea.social to the ecosystem


WhiskeyOmega (commented):
Added my servers freelancers.online and cupoftea.social to the ecosystem so updated readme

PR #1557, opened 2023-02-01
feat: add opt-in pinch to zoom setting

userquin:
Same as #1517 but disabled by default

### DIFF
--- a/composables/settings/definition.ts
+++ b/composables/settings/definition.ts
@@ -16,6 +16,7 @@ export interface PreferencesSettings {
   hideAccountHoverCard: boolean
   grayscaleMode: boolean
   enableAutoplay: boolean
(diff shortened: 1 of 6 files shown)


danielroe (approved):

patak-dev (approved):
As I said in the other PR, I'm still unsure what is the best option:  working closer as a native app and working to make it a11y from this POV, or enabling this by default and accepting that it works as a web app in mobile.

While we discuss about what is best long-term for Elk, I think it is less disruptive for current users to add this option disabled by default, that is the current behavior. So we avoid changing this twice on them. Maybe a new issue should be created that has more context and a good rationale explaining why we should switch this new default.

cc @ronilaukkarinen, reference PR:
- https://github.com/elk-zone/elk/pull/1501

PR #1563, opened 2023-02-01
fix: update GitHub card paths lookup

skryukov:
This PR fixes some problems with GitHub preview card:
- it fixes a problem with parsing of repos with `.` in name
- it also adds a fallback for (hopefully) all reserved paths from the `github-reserved-names` package. It also adds a whitelist for all supported reserved paths (only `/sponsors/{username}` at the moment)

Thread with some broken links:
https://elk.zone/universeodon.com/@elkdev/109790632883374579


patak-dev (approved):
Awesome! I wonder if at some point we could avoid the new package by using some pattern in the description to know if it is a repository, issue, or PR. But given that we load this lazily, I think this is the safest way right now. Thanks for taking care of this 🙏🏼

Issue #1565, opened 2023-02-02
Web share target misconfiguration


TheWebTech:
I've noticed that for some reason the URL isn't received from the web share API when you select Elk as a share target. Or at least it doesn't fill the URL into the text box.

It does appear to get the text.

To reproduce:
* install elk as a pwa
* go here https://web-share.glitch.me/ click share and choose elk. URL will be left off.

Not entirely sure where the web manifest file is in this project but that could be the culprit.
https://developer.chrome.com/articles/web-share-target/

Otherwise my second theory is that we're just not checking for the URL param and passing it to the text field.

userquin:
@TheWebTech yes, the url is not handled, the manifest can be found here: https://github.com/elk-zone/elk/blob/main/modules/pwa/i18n.ts#L106 and share target here https://github.com/elk-zone/elk/blob/main/service-worker/share-target.ts#L50

The client can be found here https://github.com/elk-zone/elk/blob/main/components/publish/PublishWidget.vue#L108 and composable here https://github.com/elk-zone/elk/blob/main/composables/web-share-target.ts ()

userquin:
@TheWebTech it seems Android not supporting url, check this entry https://developer.chrome.com/articles/web-share-target/#verifying-shared-content

horvbalint:
I might not understand the problem correctly, but if I share an URL with Elk, it opens and the link is placed inside the text area as I would expect.

horvbalint:
@TheWebTech @userquin
I see the problem, when I created the manifest I merged the 'text' and 'url' options into one, because we handle them the same way. But this means that if both are shared in the same time, then the 'url' is overwritten by the 'text'.
For me it worked because I shared from the url-bar and that way only the url was shared with Elk.

I will fix it after work :)

userquin:
Fixed via #1572

PR #1572, opened 2023-02-02
fix: Web Share Target text field merge

horvbalint:
Attempts to fix the issue: #1565 

As a solution we no longer merge every text type data shared with us into a single field inside the fetch request. Because this (sometimes) caused data loss. Now the three data types (title, text, url) are collected separately and then concatenated together in new lines before inserting them inside the text-area.


horvbalint:
@TheWebTech can you please try this out and see If it solves the problem on your phone?

userquin:
LGTM

Screenshot_20230202_190803_Chrome

userquin:
@horvbalint do we need to sanitize html? Maybe my last commit should be reverted, it may be wrong if we don't check if there is html

TheWebTech:
huh... odd. On desktop used the deploy preview installed it, then tried to share to it. The URL did not come through.

Used the mobile preview, and it worked just fine when installed by clicking the install prompt that the app itself pops up. It does not work if the user installs via the browser's UI.

I believe the issue with the method of install applies to both desktop and mobile. It's odd. Anyways that's separate from this issue.

Seems like you've got it working.

horvbalint:
@userquin @TheWebTech I think with this last commit this is ready to be reviewed, so I will remove the draft flag. 
Texts that are shared are now treated as text when added to the editor and not as html, so no tag will be rendered inside the editor. I also add the text to the editor line-by-line so the bug of the lines collapsing on edit should be fixed as well.

TheWebTech (approved):
LGTM woo!

userquin (approved):

PR #1576, opened 2023-02-02
feat(i18n): first draft Catalan translations 

rbnval:


userquin:
@rbnval where is the translation file? You must add the country `ca-ES`

PR #1579, opened 2023-02-02
chore: add .nvmrc to ensure node v18

milos018:
Adding a .nvmrc for easier Node version change for local development. Run "nvm i" and switched to version specified in .nvmrc, currently Node v16 + added into Contribution docs.


patak-dev (approved):

.nvmrc
```diff
@@ -0,0 +1 @@
+v16
```
sxzz, line 1: I think we can switch it to the latest LTS, Node.js 18

sxzz (changes requested):

.nvmrc
milos018: Sure, but are you referring to the .nvmrc file or the Contribution guide?
Usually, you should state the min Node version, but then .nvmrc should match that version. What do you think?

sxzz: I think we can notice that the minimum version is v16, but we recommend that v18 is better for the developers.

And set `.nvmrc` to v18.

CONTRIBUTING.md
```diff
@@ -16,7 +16,9 @@ To develop and test the Elk package:
 
 1. Fork the Elk repository to your own GitHub account and then clone it to your local device.
 
-2. Ensure using the latest Node.js (16.x)
+2. Ensure using the latest Node.js (16.x).
+If you have [nvm](https://github.com/nvm-sh/nvm), you can run ```nvm i``` to install the required version.
```
sxzz, line 20:
```suggestion
If you have [nvm](https://github.com/nvm-sh/nvm), you can run `nvm i` to install the required version.
```

danielroe (commented):
I worry about adding these extra config files.

There are also others. `asdf` uses `.tool-versions`. `volta` directly adds an entry in `package.json`. And if we're not explicitly using `nvm` in the project, or recommending its use, should we be including configuration specific to it in the repo?

milos018:
The last two projects I was involved in both used .nvmrc, which was my idea, but I am not familiar with other tools. If we do it for all of them, I fully agree it is too many config files. Also, there is a file-folding option in VS Code and other editors AFAIK.

sxzz (approved):

PR #1581, opened 2023-02-02
fix: list should be user-only

userquin:
This PR includes 
- `user-only` on navigation 
- `auth` middleware in lists and list pages
- spanish translation for new entries

### DIFF
--- a/locales/en.json
+++ b/locales/en.json
@@ -119,7 +119,7 @@
     "delete_list": {
       "cancel": "Cancel",
       "confirm": "Delete",
-      "title": "Are you sure you want to delete this list?"
+      "title": "Are you sure you want to delete \"{0}\" list?"
     },
     "delete_posts": {
       "cancel": "Cancel",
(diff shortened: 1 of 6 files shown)


patak-dev (approved):

locales/en.json
```diff
@@ -119,7 +119,7 @@
     "delete_list": {
       "cancel": "Cancel",
       "confirm": "Delete",
-      "title": "Are you sure you want to delete this list?"
+      "title": "Are you sure you want to delete \"{0}\" list?"
```
boehs, line 122:
```suggestion
      "title": "Are you sure you want to delete the \"{0}\" list?"
```

boehs (commented):
too late

locales/en.json
patak-dev: Ups, sorry, please send a new PR 🙏🏼

userquin: @boehs send a PR to fix it please :pray: , can you include that change in your current PR?

PR #1590, opened 2023-02-03
fix(ui): add viewport header to enable/disable pinch to zoom in Safari

### LINKED ISSUE #1587: Pinch to zoom still enabled on PWA iOS even when setting disabled
Tested #1557 on iOS on main.elk.zone and it does not have any effect
I still can zoom in

/ping @userquin @patak-dev

userquin:
closes #1587


Shinigami92 (approved):
Works instantly, even in the settings page were you can toggle the flag on and off 👌

patak-dev:
Awesome!

PR #1597, opened 2023-02-03
chore(i18n): `en` as source language

edimitchel:
As we have now variants enable, en-US is no more the source of language as it's a variant itself.

So, I suggest to switch too EN ~~and force the [BCP47](https://en.wikipedia.org/wiki/IETF_language_tag) language system~~

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -22,7 +22,8 @@
   ],
   "i18n-ally.preferredDelimiter": "_",
   "i18n-ally.sortKeys": true,
-  "i18n-ally.sourceLanguage": "en-US",
+  "i18n-ally.sourceLanguage": "en",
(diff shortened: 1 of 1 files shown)


userquin:
I don't use VSCode so I cannot check it: maybe Anthony can help on this.

Shinigami92:
According to this line

https://github.com/lokalise/i18n-ally/blob/435df209ec85bd43181a24635f268dce9d591c90/src/core/Config.ts#L79

it is already the default 🤔 
is it?

edimitchel:
I though the same so actually, we could remove all the config as `en` seems to be the default language source?

Glaydus (approved):
This is how it should be.
I have it set since changing en_US to en.

PR #1599, opened 2023-02-03
fix: undo changes on push notifications settings stops working

userquin:
Vueuse history stops working.

### DIFF
--- a/composables/push-notifications/usePushManager.ts
+++ b/composables/push-notifications/usePushManager.ts
@@ -1,4 +1,5 @@
 import type { mastodon } from 'masto'
+
 import type {
   CreatePushNotification,
   PushNotificationPolicy,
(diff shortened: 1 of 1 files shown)


patak-dev (approved):
I think we can merge this now. But is there a reason for these settings to work in a different way from all the others? Why are we allowing the user to cancel here? In all other settings screens, we apply the modification and move on. Should we also do the same here?

userquin:
> I think we can merge this now. But is there a reason for these settings to work in a different way from all the others? Why are we allowing the user to cancel here? In all other settings screens, we apply the modification and move on. Should we also do the same here?

All these setting are saved on backend, mastodon rn just send a request per check changed, we allow send all at once

patak-dev:
Ah, I see we also do the same for the Profile. Is it costly to do multiple requests? I think it may be better to just do the changes with out the save. But ya, ignore my prev comment because this is now the pattern we use for other things too.

PR #1602, opened 2023-02-03
fix: make nav title sticky + add more grabbale area for native

edimitchel:
- Revamp the padding around logo
- make the logo sticky in case of small height screen
- Add more global native grab attribute to let user to grap from more empty space

native-drag


boehs:
I'm not quite sure what that gif is supposed to generate 😁

JonasKruckenberg:
https://user-images.githubusercontent.com/25329995/216710929-73e9f653-622a-4554-93ae-938b9434f139.mov

So your change basically makes it so that only the immediate area around the elk logo is draggable, the div that had the attribute before that was already the correct place I think (bc you would expect the top bar to be draggable) but maybe it being a link now interfered with that?

edimitchel:
> I'm not quite sure what that gif is supposed to show 😁

Now should be better!

JonasKruckenberg (approved):
Can confirm this works great 👍🏻

patak-dev (approved):

PR #1603, opened 2023-02-03
feat: quickfixes

### LINKED ISSUE #155: Verified website, github, twitter in header
@antfu thinking about how to implement this with the current UI. Maybe a primary color check icon next to the link? Or just use a primary color for the link when it is verified? If we don't want to use green, I think we may need the tick

https://mas.to/@addyosmani@indieweb.social

https://elk.zone/@addyosmani@indieweb.social

### LINKED ISSUE #1410: Placement of ALT button ruins images.
Thank you for all your work on this beautiful client.

The single thing that prevents me from using it is the obtrusive ALT button on the lower-left of each image. The value placed on alt text for images is one of the great things about Mastodon. But the placement of a button here destroys images for those who regard them seriously, such as photographers, painters, graphic artists, et al. Imagine going to a museum and seeing a big ALT sticker in the corner of every painting.

Perhaps there's an alt location for the ALT button?

Thanks for your consideration,
Jed

### LINKED ISSUE #1528: Emoji unsaturated in profile metadata
The emojis in profile metadata, the ones with automatic icons added, are unsaturated.
Problem is that, color can be a important part of an emoji's information. An obvious case is flags, often used in the "Location" or "Language" field.

Screenshot as example :
image

Is there a real value to this filter? It looks a bit cleaner but the tradeoff seems too big in my opinion.

boehs:
consider merging without squashing, each fix is contained in it's own commit

- feat: add outline to verified links closes #155
- fix: remove saturation for links fixes #1528
- feat: option to hide alt indicator closes #1410
- add wellness section to preferences (no issue?)

### DIFF
--- a/locales/en.json
+++ b/locales/en.json
@@ -400,6 +400,7 @@
       "github_cards": "GitHub Cards",
       "grayscale_mode": "Grayscale mode",
       "hide_account_hover_card": "Hide account hover card",
+      "hide_alt_indi_on_posts": "Hide alt indicator on posts",
       "hide_boost_count": "Hide boost count",
       "hide_favorite_count": "Hide favorite count",
       "hide_follower_count": "Hide follower count",
--- a/locales/en.json
+++ b/locales/en.json
@@ -409,7 +409,8 @@
       "label": "Preferences",
       "title": "Experimental Features",
       "user_picker": "User Picker",
-      "virtual_scroll": "Virtual Scrolling"
+      "virtual_scroll": "Virtual Scrolling",
+      "wellbeing": "Wellbeing"
     },
     "profile": {
       "appearance": {
--- a/components/account/AccountHeader.vue
+++ b/components/account/AccountHeader.vue
@@ -156,8 +156,8 @@ const isNotifiedOnPost = $computed(() => !!relationship?.notifying)
           <ContentRich :content="field.value" :emojis="account.emojis" />
         </div>
       </div>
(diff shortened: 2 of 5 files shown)


patak-dev:
@boehs this looks good to me, are you planing to add more commits on top of it? Even if it is a bit of extra work I would prefer we keep doing one PR per feature/fix as it is a lot easier then merging and discuss each individually 🙏🏼 
We can still move on with this one though.

boehs:
> @boehs this looks good to me, are you planing to add more commits on top of it? Even if it is a bit of extra work I would prefer we keep doing one PR per feature/fix as it is a lot easier then merging and discuss each individually 🙏🏼 We can still move on with this one though.

sounds good. It kinda occurred to me as I was 3rd commit in. Sure, 3 is better than none!

locales/en.json
```diff
@@ -400,6 +400,7 @@
       "github_cards": "GitHub Cards",
       "grayscale_mode": "Grayscale mode",
       "hide_account_hover_card": "Hide account hover card",
+      "hide_alt_indi_on_posts": "Hide alt indicator on posts",
```
edimitchel, line 403:
```suggestion
      "hide_alt_indicator_on_posts": "Hide alt indicator on posts",
```
?

boehs: Fair enough. I'll package this with a future pull request if that's ok

PR #1606, opened 2023-02-03
feat: allow running elk with a single server

### LINKED ISSUE #953: Feature request: Specify instance name for local instance installs and have instance lifecycle events supported
Ask:

Specify instance name for local instance installs and have web experience feel native to that instance.

For example, I have an instance running at https://elk.universeodon.com - If you try and login, it asks for an instance name - if i specify one, it should bypass and just ask to login or authorize the app to Universeodon.com or register on universeodon.com.

Additionally, if this local instance name is provided, there should be links to login/register to the default instance.  

When you set it to a default instance, it would be nice if the default routes matched that of mastodon - without the /servername.com/ - this will help people not confuse googlebot if they impelement this as a front end.

It would also be nice if there was a place for instance notifications to be supported so if I create a notification, there is a flashing bell or something for local instance notifications.  It's built into the standard UI now, but not very visibly ;)

image

Eventually, it would be nice if i could just set my theme and have it load elk, but that can be saved for later ;)

userquin:
This PR will use `process.env.SINGLE_INSTANCE_SERVER` to allow only sign-in users to only public server.

To test this PR you should clone the branch and start dev server with `SINGLE_INSTANCE_SERVER=true`.

~~I need to add some animation to the sign-in button.~~

Since we can sign-in from multiple places, there is no way to include the error in the ui, and so, I've included a `ModalError`  dialog:

imagen

This PR includes new entries (provided for `en` and `es` locales):
- `action.sign_in_to`

#953


composables/useSingleSignIn.ts
```diff
@@ -0,0 +1,41 @@
...
+
+    busy = true
+
+    await nextTick()
+
+    try {
+      location.href = await (globalThis.$fetch as any)(`/api/${publicServer.value}/login`, {
+        method: 'POST',
+        body: {
+          force_login: true,
```
userquin, line 20: this should be `users.value.length > 0`

composables/useSingleSignIn.ts
```diff
@@ -0,0 +1,41 @@
...
+    }
+    catch (err) {
+      console.error(err)
+      busy.value = false
+      await openErrorDialog({
+        title: t('common.error'),
+        messages: [t('error.sign_in_error')],
+        close: t('action.close'),
+      })
+    }
```
danielroe, line 37: can we somehow share this code with https://github.com/elk-zone/elk/blob/f3642ed9d54443a9dd73930255d37b938b4e45b8/components/user/UserSignIn.vue#L16-L53?

composables/useSingleSignIn.ts
```diff
@@ -0,0 +1,41 @@
+export const useSingleSignIn = () => {
```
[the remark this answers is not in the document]
danielroe: Actually, I misread this code - I see now that we don't really want to tree-shake it out - I think we just need to dedupe it with the code in the `oauth()` function.

composables/useSingleSignIn.ts
userquin: sure: I will check also `entry-*.js` asset size ~500KB (current is about 128KB), maybe we can use dynamic import on nav (maybe due to duplicate code)

userquin: imagen

components/nav/NavUser.vue
```diff
@@ -15,7 +19,21 @@
...
+      v-if="singleInstanceServer"
+      flex="~ row" gap-x-1 items-center justify-center btn-solid text-center text-sm px-2 py-1 xl:hidden
+      :disabled="busy"
+      @click="oauth()"
+    >
+      <span v-if="busy" aria-hidden="true" block animate animate-spin preserve-3d class="rtl-flip">
+        <span block i-ri:loader-2-fill aria-hidden="true" />
+      </span>
+      <span v-else aria-hidden="true" block i-ri:login-circle-line class="rtl-flip" />
+      {{ $t('action.sign_in') }}
```
danielroe, line 33: I think we should make this clearer for people who have used Elk before but might not realise:

In English, something like: 'Sign in to <strong>m.webtoo.ls</strong>'

danielroe (approved):
This is 👌 Happy to merge and iterate, but worth noting a nice-to-have from https://github.com/elk-zone/elk/issues/953 (and I think the final thing that would close the issue) would be omitting the `/servername.com/` in the routing.

userquin:
> This is 👌 Happy to merge and iterate, but worth noting a nice-to-have from #953 (and I think the final thing that would close the issue) would be omitting the `/servername.com/` in the routing.

ok, this can be included later if requested

userquin:
@danielroe a few questions:
- should we also update docs? If so, where should be included? 
- Should `SINGLE_INSTANCE_SERVER` be prefixed with `NUXT_`?

userquin:
@danielroe sign-in working, permalink middleware change solves it, thx man, I guess this PR is ready to be merged 🎉

The sign-out still there, we can fix it later, not related to this PR.

danielroe:
Great. Merging to iterate and test on canary.

PR #1608, opened 2023-02-03
fix: hide side nav in zen mode

### LINKED ISSUE #1605: regression -- huge sidebar in zen mode
Is there any way to bring back the behavior where both sidebars hide on the desktop while in zen mode?

I just upgraded from an older build where both sides hide, then reappear on hover, which was utterly delightful.  It spoiled me, because I could fullscreen my window and just bliss out on pure content. 

image

I understand that you need a rediscovery affordance for non-hoverable devices, but this regression is really disappointing, and totally undermines the zen-ness of the experience.   (Who cares that little UX affordances disappear when the biggest baddest one stays put?)

image

image

edimitchel:
Attribute `zen-hide` is not working (cc @antfu any idea?)

Resolves #1605


patak-dev (approved):
Weird... let's merge this as a temporal fix

edimitchel:
My bad, this is caused because the `zen-mode` is a classic class where it's not working with unocss

PR #1611, opened 2023-02-03
fix: reply language defaults to status language

### LINKED ISSUE #1609: Reply drafts language should default to original status language
I would love Elk to behave like the mastodon default Web UI. When replying to an english toot, the language is automatically set to english. When replying to a german toot, it's set to german... in addition to a language preference for new toots of course.

_Originally posted by @sariash in https://github.com/elk-zone/elk/issues/1196#issuecomment-1383571520_
            
I think this is a separate issue I can work on.
Basically the language selection of reply drafts should default to the language of the original status.

niklaswolf:
closes #1609 

This PR sets the default language on replys to the language of the original status

### DIFF
--- a/composables/masto/statusDrafts.ts
+++ b/composables/masto/statusDrafts.ts
@@ -74,6 +74,7 @@ export function getReplyDraft(status: mastodon.v1.Status) {
         inReplyToId: status!.id,
         visibility: status.visibility,
         mentions: accountsToMention,
+        language: status.language,
       })
     },
   }


edimitchel (approved):

boehs (commented):
Thanks for these two PRs!

boehs (approved):

sariash:
Thanks SO much for this. Drove me crazy. I was always forgetting to change the language by hand, and had to edit my post afterwards :)

PR #1612, opened 2023-02-03
feat: add preference to hide emojis in usernames

JohnyTheCarrot:


JohnyTheCarrot:
Forgot to handle emoji-only names, will do tomorrow.

antfu:
This is something I personally would love to have. Not sure how others think about it tho.

JohnyTheCarrot:
> This is something I personally would love to have. Not sure how others think about it tho.

It was originally proposed by @patak-dev in the Discord ==> https://discord.com/channels/1044887051155292200/1044887051155292203/1071163298420097185

patak-dev (approved):
This makes a huge difference. Given that it isn't by default, I think it is ok to include it in the same way that we have grayscale or other kinds of filtering.

PR #1613, opened 2023-02-03
feat: continuous threads

### LINKED ISSUE #1355: Status design and flow
I have been curiously watching Elk's changes to the status page, from the reordering of replies into threads, the repositioning of mentions to a separate line, the hiding of the mention line when the information is already suggested by surrounding context, to Zen Mode.

If we start removing all the boilerplate as this had been the case with these changes, the page actually starts to look like a chat conversation, a list of replies in roughly ordered chronologically, from top to bottom, mostly from old to new, mixing direct replies and sub-conservations, in a single flow.

Elk is not a chat application, but it increasingly looks like one. I feel this shows a willingness to focus on the reading, not only the reading of one status, but of the flow of statuses.

It then feels natural to continue to look at chat applications, which, by their nature, enable a more direct form of communication. With this in mind, I've been to trying to see how we could incorporate some simple design patterns from messaging apps into Elk.

- The action toolbar currently takes the entire width of a post, and acts like an implicit border, a separator between the different posts. I suggest to shorten the action bar, breaking this impression of separation, similar to Slack, Discord, or WhatsApp, where it is often shown aside of the messages in a compact shape.
- The name and handle line is currently shown for all posts, even for statuses that are written by the same individual as part of a thread. I suggest to remove the username line, for the consecutive replies that have the username as the first post, similar to WhatsApp, Discord...
- Replies that start a new thread are only identified by the line "Replying to", followed by the name of parent poster. I suggest to add further context to this line, such as the name of the mentions that are part of that initial thread, and an excerpt of the parent message, similar to WhatsApp, Discord...

I made a quick mock-up to illustrate these changes, the visual aspect will obviously require further work, especially the message preview, but this should sufficient for demonstration purposes.

mockup

With just these few minor changes, replies appear as if they belong to the same document, less segmented, a thread from one individual appears like a longer text, with fewer obstruction, without losing functionality. Messages are seen as part of a context, less likely to be singled out, and picked apart.

I am curious to hear your opinions.

boehs:
closes #1355


boehs:
Looks like this introduces a regression on mobile

boehs:
Ok, so this is turning into a small CSS refactor. I will elaborate on the exact style changes when it is ready, but

1. Fixes two visual bugs (a white bar over profile pictures on scrolling and a border too large above profile pictures)
2. Changes padding around posts to be more consistent (discussion needed)
3. Makes the line have 3px space between both the beginning and end of the PFP

boehs:
currently known regessions:

image
image

and

image
image

current issues: last post in cont mode doesn't have line

locales/en.json
```diff
@@ -401,12 +401,13 @@
...
     "notifications_settings": "Notifications",
     "preferences": {
+      "continuous_threads": "Continuous Threads",
       "enable_autoplay": "Enable Autoplay",
       "enable_pinch_to_zoom": "Enable pinch to zoom",
       "github_cards": "GitHub Cards",
       "grayscale_mode": "Grayscale mode",
       "hide_account_hover_card": "Hide account hover card",
-      "hide_alt_indi_on_posts": "Hide alt indicator on posts",
+      "hide_alt_indicator_on_posts": "Hide alt indicator on posts",
```
userquin, line 410: can you also update rest of `json` files?

boehs: Sure, in a couple hours. Any other feedback?

userquin: it was already there

boehs: This is just a font issue. I'd like to avoid *more* magic numbers

userquin: I mean, with the avatar in the publish widget :sweat_smile:

boehs: Oh yeah, well that's not my fault ;)

components/status/StatusCard.vue
```diff
@@ -132,32 +120,37 @@ const showReplyTo = $computed(() => !replyToMain && !directReply)
       </div>
     </slot>
 
-    <div flex gap-3 :class="{ 'text-secondary': faded }">
+    <div flex mt2 gap-3 :class="{ 'text-secondary': faded }">
```
boehs, line 123: this is a problem if there is a reblog or replying to

userquin:
can you resolve conflict?

PR #1397, opened 2023-01-22
fix: prevent videos from clipping when in fullscreen mode

aeharding:
Hi!

This prevents videos from clipping (display-mode: cover) when in fullscreen.

~~I'm looking for better unocss ways to accomplish this, but I havenoideawhatI'mdoing.jpg so any advice would be appreciated. :) Thank you!~~ Now resolved, thanks @kreha1!

Edit: Worth noting this doesn't affect iOS - it's only a problem on desktop (tested on macOS, should also affect Windows) with chrome/ff/safari.

This is mainly a problem when you try to fullscreen a portrait video on a landscape (aka, regular laptop) screen.

PR #1616, opened 2023-02-04
fix: fullscreen video scaling issues

kreha1:
While using my ultrawide monitor, I encountered a scaling issue with full-screen videos  (firefox, linux). They were stretched and displayed incorrect proportions. I quickly opened dev tools and poked around, and discovered that the "object-fit" property was set to "cover," causing the issue.

After checking it on my 13-inch laptop (firefox, macos) and mobile (firefox and chromium-based, android) I haven't found any negative consequences.

The surprising bit is that scaling issues existed and were degrading video quality on regular displays, even though it went unnoticed until now.

Before:
21:9
image

16:10
image

After:
21:9
image

16:10
image

mobile horizontal
image

mobile vertical
image

### DIFF
--- a/components/status/StatusAttachment.vue
+++ b/components/status/StatusAttachment.vue
@@ -103,7 +103,7 @@ const userSettings = useUserSettings()
         playsinline
         controls
         rounded-lg
-        object-cover
+        object-contain
         :width="attachment.meta?.original?.width"
         :height="attachment.meta?.original?.height"
         :style="{


aeharding:
I believe videos should still be set to "cover" for when *not* full screen though, right?

That's what I was attempting to do in the following PR:

https://github.com/elk-zone/elk/pull/1397

kreha1:
> I believe videos should still be set to "cover" for when _not_ full screen though, right?
> 
> That's what I was attempting to do in the following PR:
> 
> #1397

Ok, I will close this PR then and contribute to yours :).

PR #1617, opened 2023-02-04
fix(i18n): Update zh-TW translations

xatier:
* prefer 您 over 你

In [zh-TW translation](https://crowdin.com/project/mastodon/zh-TW) of Mastodon Web UI, we prefer to use 您 to be more polite. This PR is to make Elk to be consistent with Mastodon Web UI.

### DIFF
--- a/locales/zh-TW.json
+++ b/locales/zh-TW.json
@@ -21,7 +21,7 @@
     "followers_count": " {0} 位粉絲",
     "following": "正在追蹤",
     "following_count": "正在追蹤 {0} 人",
-    "follows_you": "已追蹤你",
+    "follows_you": "已追蹤您",
     "go_to_profile": "轉到個人資料",
     "joined": "已加入",
     "moved_title": "的新帳號是：",
@@ -103,12 +103,12 @@
     "block_account": {
       "cancel": "取消",
       "confirm": "封鎖",
-      "title": "你確定要封鎖 {0} 嗎？"
+      "title": "您確定要封鎖 {0} 嗎？"
     },
     "block_domain": {
       "cancel": "取消",
       "confirm": "封鎖",
-      "title": "你確定要封鎖 {0} 域名嗎？"
+      "title": "您確定要封鎖 {0} 域名嗎？"
     },
     "common": {
       "cancel": "否",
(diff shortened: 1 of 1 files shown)


xatier:
@Mini-ghost would you please take a look?

Also, I maintain a [list of terms](https://github.com/xatier/mastodon-translation-zhTW) we use in the other Mastodon platforms, you can update zh-TW translations in elk if you wish.

Mini-ghost:
@xatier [list of terms](https://github.com/xatier/mastodon-translation-zhTW) is awesome, you can also open a PR for this if you wish too.

For this PR, I don't think honorifics are needed here, like Facebook and Twitter, they use 你 instead of 您.

xatier:
@Mini-ghost using 您 in zh-TW i18n is pretty common in many computer systems, such as Microsoft [1] and Apple [2] products. I believe it would make the users more comfortable when the software sounds more polite. Early users should feel it enjoyable with this PR as we are all used to this term since quite a while ago [3].


If you disagree with this, feel free to dismiss this PR.

[1] https://www.microsoft.com/zh-tw/language/Search?&searchTerm=you&langID=129&Source=true&productid=0
[2] https://applelocalization.com/?q=%E6%82%A8&l=Traditional+Chinese
[3] https://www.youtube.com/watch?v=CLkTR4qkpsc

Mini-ghost:
@xatier I think it's all good, but the choice of wording may have something to do with the image the Elk core team wants to present.

In my opinion:

- Using 您 makes people feel more professional and formal.
- Using 你 gives people the feeling that there is relaxed and lively.

PTAL @antfu

antfu:
As long as it's consistent across the app and both Simplified and Traditional Chinese, I think both are fine.

I personally think 你 is good enough.

/cc @elk-zone/translators-zh-cn would be nice to have your input from the Simplified Chinese perspective.

david50407:
As a zh-tw user, I think 您 is more comfortable and frequently used.

There is not only two Chinese variants: Simplified(zh-hans) and Traditional(zh-hant). IMO, when we use zh-tw, that means dialect in Taiwan, we may consider the culture difference not only the consistent words to other variants.

didavid61202:
I prefer using '你', '您' feels like too formal for the vibe and context of mastodon and elk.

'您' is only better for usages in welcoming text or informing letters,
While using '你' in frequently appearing context will feel more friendly like a close friend intersecting with you.

antfu:
> we may consider the culture difference not only the consistent words to other variants.

I have been living in both cultures for years, and from what I perceive, there is not much cultural difference between the usage of 你 and 您, as both two words exist in both cultures with similar usage.

I think the discussion here is how Elk should position the feeling to the user (more formal and polite or more friendly), which I believe should be consistent.

xatier:
@antfu thanks for sharing your thoughts.

Just to clarify, I'm not familiar with zh-CN culture hence I didn't touch other zh variants. I would leave the decision to the core ELK team. If we'd like to make it consistent with using 您 for all zh translations, happy to update my PR.

Otherwise, we can just drop this.

IMHO, I'd like to have ELK showing the same experience with the Original Mastodon UI, i.e., 您 in zh-TW / 你 in others.

antfu:
After some thought, I think if we have to make the decision and move forward, I will pick "你" for Elk to become more friendly. So let's close this PR. Thanks for bringing this up.

PR #1631, opened 2023-02-05
feat: adjust hide username emojis pref to only hide in timeline

JohnyTheCarrot:
This PR makes it so descriptions are now properly supported in preferences, as well as applying the suggestion of making the hide username emojis pref only apply to the timeline.


components/account/AccountDisplayName.vue
```diff
@@ -3,6 +3,7 @@ import type { mastodon } from 'masto'
 
 defineProps<{
   account: mastodon.v1.Account
+  useShowEmojisPreference?: boolean
```
patak-dev, line 6: Even if it will involve getting the preference in more places, I think it is better to change this prop to `hideEmojis` defaulting to false and opt-in where we want to remove them to be more explicit.

JohnyTheCarrot:
hold on, stuff is broken it seems

patak-dev:
Is this showing emojis for you when looking at the profile with the option enabled?
https://deploy-preview-1631--elk-zone.netlify.app/universeodon.com/@Shini92@mas.to

How it should look https://main.elk.zone/mas.to/@Shini92

patak-dev (approved):
Ok, the issue was that the SW was not updated yet. @userquin I thought we disabled it in previews, or when we check a PR we should still wait for the prompt after new commits are added?

userquin:
Yes, the PWA will be there. You should always open PR previews with private browsing, the sw will be removed once you close the window.

PR #1637, opened 2023-02-05
fix: change-case error 500 on canary

userquin:


userquin:
closed via https://github.com/elk-zone/elk/commit/e31fc4087c0aa66917106a4ab00238de980e7d1c

PR #1640, opened 2023-02-05
feat(i18n): Add ko-KR (Korean) translation

dahlia:


antfu:
Sorry for the long delay. It seems we already have a Korean translation from another PR. Can you review it and submit another PR if there is any improvements from your translations? Thanks. Closing due to conflicts.

PR #1647, opened 2023-02-05
perf: tree-shake dependencies from server

danielroe:
This takes the node server build output (including dependencies) from 23M -> 9.8M.

It also improves compatibility with cloudflare target, though I'm not sure we're fully there.


patak-dev (approved):
This is awesome! Let's merge it and let folks test on main.elk.zone during this week

PR #1654, opened 2023-02-06
fix: sign-out and empty content

### LINKED ISSUE #1621: when sign-out with no more users main content is empty
It seems a regression, there is an error on console, last screenshot on `elk.zone` (`signOut` in `users.ts` module, there is no user and so `masto.ts` module fails checking `user.server` in L30).

I have removed the error using this code but main content is empty (included similar code in this PR https://github.com/elk-zone/elk/pull/1606, check last commit):

el-signout-error

image

userquin:
closes #1621

### DIFF
--- a/middleware/auth.ts
+++ b/middleware/auth.ts
@@ -8,12 +8,12 @@ export default defineNuxtRouteMiddleware((to) => {
     return
 
   if (isHydrated.value)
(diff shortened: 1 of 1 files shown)


danielroe (approved):
👌

boehs:
No more racing 😾

PR #1658, opened 2023-02-06
fix: handle frozen page lifecycle state

### LINKED ISSUE #750: PWA and iOS - it gets suspended and doesn't reload new statuses after relaunch
iOS suspends PWAs after 10 minutes of idle (minimised) time. At relaunch, Elk doesn't seem to understand it's been "resumed" and the timeline is stuck in an old state. The only way to refresh is to kill the app and relaunch, but it's not that comfortable.
Thank you!

### LINKED ISSUE #1584: Elk PWA on main.elk.zone does not refresh on Firefox on Android
I have installed main.elk.zone as a PWA on Firefox on my Android device running Android 13.

Within the last few days, it has almost entirely stopped refreshing. Tapping the Elk icon (only visible in landscape mode) does not refresh, and it only refreshes a few times per day. I don't know exactly what triggers it to be able to update. The only thing I can do to refresh it is to stop the entire Firefox app and then start it again. Otherwise, it has picked up newer content only a few times; I'm seeing only hours-old content even though the web view and Tusky both see current content.

userquin:
@danielroe just added the plugin, we'll need to close all `database` and `wss` connections : for reference check **Frozen** state here https://developer.chrome.com/blog/page-lifecycle-api/#developer-recommendations-for-each-state

closes #750 
closes #1584

### DIFF
--- a/composables/pwa.ts
+++ b/composables/pwa.ts
@@ -1,0 +1,6 @@
+// https://thomashunter.name/posts/2021-12-11-detecting-if-pwa-twa-is-installed
+
+const ua = navigator.userAgent
+const ios = ua.match(/iPhone|iPad|iPod/)
+const standalone = window.matchMedia('(display-mode: standalone)').matches
+export const isPWAInstalled = !!(standalone || (ios && !ua.match(/Safari/)))
--- a/composables/pwa.ts
+++ b/composables/pwa.ts
@@ -1,6 +1,0 @@
-// https://thomashunter.name/posts/2021-12-11-detecting-if-pwa-twa-is-installed
-
-const ua = navigator.userAgent
-const ios = ua.match(/iPhone|iPad|iPod/)
-const standalone = window.matchMedia('(display-mode: standalone)').matches
-export const isPWAInstalled = !!(standalone || (ios && !ua.match(/Safari/)))
--- a/shims.d.ts
+++ b/shims.d.ts
@@ -2,7 +2,6 @@
 /// <reference types="vite-plugin-pwa/info" />
 /// <reference types="vite-plugin-pwa/client" />
 
-
 declare global {
   namespace NodeJS {
     interface Process {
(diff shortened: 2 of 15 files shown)


userquin:
We need to resolve the `TODO` in `page-lifecycle.client.ts` plugin before merging this PR

edimitchel:
Works fine on Android!

shims.d.ts
```diff
@@ -2,6 +2,7 @@
 /// <reference types="vite-plugin-pwa/info" />
 /// <reference types="vite-plugin-pwa/client" />
 
+
```
danielroe, line 5:
```suggestion
```

composables/pwa.ts
```diff
@@ -0,0 +1,6 @@
+// https://thomashunter.name/posts/2021-12-11-detecting-if-pwa-twa-is-installed
+
+const ua = navigator.userAgent
+const ios = ua.match(/iPhone|iPad|iPod/)
+const standalone = window.matchMedia('(display-mode: standalone)').matches
+export const isPWAInstalled = !!(standalone || (ios && !ua.match(/Safari/)))
```
danielroe, line 6: can we inline this in plugins/pwa.client.ts instead to make it clearer?

userquin: this is part of the TODO, maybe we can revert that change

danielroe (approved):

composables/pwa.ts
danielroe: Looks good to me to merge whenever you're happy. I would do as you've suggested:

> maybe we can keep current code (just remove the comment and undo moving isPWAInstalled to its own module)

userquin:
@danielroe this is ready, on iOS emulator, the prompt for update seems to be not working, but it works for new toots once the installed PWA minimized for 10 minutes.

cyberalien:
Tested in iOS 16.2, app working normally, except for update as @userquin mentioned. Not seeing update notification anywhere.

cyberalien:
Tested it again on localhost with https, everything works now. Even got PWA update notification.

Screenshot 2023-02-06 at 23 55 28


Though there was a weird error with update notification, but after clicking button to update app (had to click it twice - first click didn't do anything), got correct home page.

PR #1664, opened 2023-02-07
refactor: move single instance config to runtime

danielroe:
this was a suggestion from @userquin and on reflection I think it would be useful for single instace to be configurable at runtime, so docker image can enable it at runtime rather than rebuilding, for example


nuxt.config.ts
```diff
@@ -115,6 +114,7 @@ export default defineNuxtConfig({
       translateApi: '',
       // Use the instance where Elk has its Mastodon account as the default
       defaultServer: 'm.webtoo.ls',
+      singleInstance: false,
```
userquin, line 117: shouldn't use `process.env.NUXT_PUBLIC_SINGLE_INSTANCE` here?

danielroe: Yes, Nuxt will automatically override it.

userquin (approved):

PR #1669, opened 2023-02-08
feat(i18n): add missing Polish (and English) translation entries

Glaydus:


Glaydus:
Adding missing translations (including English).
Minor tweaks to **_SettingsProfileMetadata.vue_** were required.

components/settings/SettingsProfileMetadata.vue
```diff
@@ -66,12 +66,14 @@ const chooseIcon = (i: number, text: string) => {
...
-          type="text" placeholder="Label"
+          type="text" placeholder-text-secondary
+          :placeholder="$t('settings.profile.appearance.profile_metadata_label')"
           input-base
         >
         <input
           v-model="form.fieldsAttributes[i - 1].value"
-          type="text" placeholder="Content"
+          type="text" placeholder-text-secondary
+          :placeholder="$t('settings.profile.appearance.profile_metadata_value')"
```
patak-dev, line 76: @userquin would you add a comment to this PR showing how you would like we announce these two new key?

userquin: The message should be this:

New entries to translate (only included in `en-` and `pl-PL`):
- `settings.profile.appearance.profile_metadata_label`
- `settings.profile.appearance.profile_metadata_value`

userquin: We only need to copy the source message and paste it on i18n channel.

PR #1673, opened 2023-02-08
feat(ui): Adding Mute/Block user/domain to the post dropdown menu as well

### LINKED ISSUE #1518: Easier way to mute
On Mastodon, mute action is directly accessible. On Elk, we need to click on the profile, loading it (sometime it is a bit slow, depending on the instance I guess?) then we can mute.

Maybe adding the mute action to the three dots menu? Or in the tooltip?

peterbud:
This PR adds parity functions, enabling Mute/ Block user/Block domain directly from the post dropdown menu, without the need to open the user's profile.

Fixes #1518

### DIFF
--- a/components/status/StatusActionsMore.vue
+++ b/components/status/StatusActionsMore.vue
@@ -1,5 +1,6 @@
 <script setup lang="ts">
 import type { mastodon } from 'masto'
+import { toggleBlockAccount, toggleBlockDomain, toggleMuteAccount, useRelationship } from '~~/composables/masto/relationship'
 
 const props = defineProps<{
   status: mastodon.v1.Status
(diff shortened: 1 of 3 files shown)


Shinigami92:
@peterbud could you please rebase and fix the conflict?

components/status/StatusActionsMore.vue
```diff
@@ -246,6 +247,51 @@ const showFavoritedAndBoostedBy = () => {
...
+            />
+            <CommonDropdownItem
+              v-else
+              :text="$t('menu.unblock_account', [`@${status.account.acct}`])"
+              icon="i-ri:checkbox-circle-line"
+              :command="command"
+              @click="toggleBlockAccount(useRelationship(status.account).value!, status.account)"
+            />
+
+            <CommonDropdownItem
```
aaga, line 281: The "Block/Unblock Domain" items should only be shown when the account is on a different domain to the current user. See the equivalent `v-if` in `AccountMoreButton.vue`

patak-dev:
Thanks for the work here @peterbud! We merged it through https://github.com/elk-zone/elk/pull/2184 (sorry I missclicked and ended up removing the attribution 🙏🏼)

PR #1684, opened 2023-02-09
chore: upgrade to nuxt v3.2.0

danielroe:


danielroe:
@sxzz Do you know the cause of the type issue?

sxzz:
@danielroe I have no idea with this one. It could be a breaking change with Nuxt 3.2.

/cc @antfu

danielroe:
I can confirm the issue is caused by the upgrade from Vue 3.2.45 to 3.2.47. When reverted there are no type errors.

PR #1691, opened 2023-02-09
fix(ui): make nav links unselectable on mobile

### LINKED ISSUE #1661: Main nav icon buttons are too easy to select as text on mobile
I use Elk on iOS and I find I’m often highlighting the icon buttons in the nav bar when I go to press them, and then they become unusable while highlighted. I’ve hesitated to make an issue, because I know that setting `user-select: none` isn’t always the best for accessibility, but this happens to me so frequently, that I thought I’d raise the issue. Perhaps using `@media (pointer: coarse)` and targeting for mobile would be the best coarse of action, if `user-select: none` was applied?

knowler:
Resolves #1661

Summary of the bug (#1661): on mobile (at least iOS) it’s fairly easy to accidentally select all of the links in the bottom navigation bar which is frustrating as a user since none of the links are usable while they are selected and unselecting them can be a challenge. Further, there is no use-case for being able to select these; copying the selection will just copy a bunch of new lines or spaces.

This applies `user-select: none` to each of the nav links. To be safe, I’m only applying the rule when the `pointer: coarse` media query matches.

This was my first time using Nuxt and Uno, so hopefully I’m doing things correctly — let me know otherwise.


boehs:
Don't cellphones match `pointer: none` instead of coarse?

knowler:
> Don't cellphones match `pointer: none` instead of coarse?

No, a cellphone or touch device would match `coarse` because a finger (or stylus) would be a pointer device with lower accuracy.

Issue #1683, opened 2023-02-09
Add Matrix and email profile metadata icons


salarua:

ZMYaro:
Elk uses the [Remix icon pack](https://remixicon.com).  I am not sure which icon from that, if any, would make sense for Matrix, but email should definitely be doable!

salarua:
there doesn't seem to be a Matrix icon at all. it wouldn't be difficult to make though since it's literally just a lowercase m in square brackets: [m]

ZMYaro:
@salarua I just checked, and it appears someone already submitted a request for one! https://github.com/Remix-Design/RemixIcon/issues/209

salarua:
I noticed today that Matrix now has an icon. Thank you!

PR #1693, opened 2023-02-10
Match account metadata icon names with trailing colons and add/update more icons

ZMYaro:
This also addresses the email half of #1683.


ZMYaro:
Not sure why the diff is showing every icon line changed.  I added/altered `Birth`, `Birthday`, `BuyMeACoffee`, `'Buy Me A Coffee'`, `Email`, `E-mail`, `Facebook`, `Home`, `'Home Page'`, `Homepage`, `Joined`, `KoFi`, `Ko-fi`, `Meatspace`, `Skype`, `Snapchat`, `Substack`, and `Tumblr`.

patak-dev:
We need to implement the matching separated from the list (maybe you could add an `alias` field or a new alias map?). The issue with the PR right now is that when rendering the full list to pick from, you get several duplicates:
image

ZMYaro:
@patak-dev Looking into this now, even after constructing an alias map, I am not sure how you want to decide which text label is the default for each icon.  For instance, I am not sure how to decide whether the key icon should insert the “GPG”, “Keyoxide”, “PGP”, or “OpenPGP” label by default.

patak-dev:
@ZMYaro ah, I didn't realize that PGP and Language were both already duplicated. I think the issue with this PR is that it is a lot more evident after it.

About what Label to use, we could select the most common one as the default (or any of them if all are similar in use). This is only to avoid duplicates in the picker, so I think it is ok to pick one of them. If the user writes the label by hand, the icon would still be properly assigned.

ZMYaro:
@patak-dev I am wondering then if having the alias list isn't the best solution, and it would be better to have a list of featured labels specifically for that component?  Because we presumably don't need both “Home Page” and “Homepage”, but _do_ want “GPG” and “Keyoxide”.

patak-dev:
In the case of “GPG” and “Keyoxide”, they better have different icons if possible, no? Not an issue if some of them end up duplicated though for now. Just that we can't have `Ko-Fi` and `KoFi`

ZMYaro:
> In the case of “GPG” and “Keyoxide”, they better have different icons if possible, no?

@patak-dev Right now, they have the same icon (you can see right now in Elk stable).  If we picked one label for each icon, we would have to pick one and rule out all the others.  That was why I suggested it might work better to have a list of labels to suggest in the `SettingsProfileMetadata` menu and keep the existing map of labels to icons.

PR #1694, opened 2023-02-10
Account metadata chip redesign proposal

ZMYaro:
This modifies the chip layout to include each field's name, makes every link field a chip instead of just links with icon names,makes the entire chip a clickable link instead of just the URL, and changes the icon to a checkmark for verified fields.  Proposed solution to #155.


edimitchel:
That's so nice!

patak-dev:
Thanks for the PR! We definitely need to improve here. But in my opinion, we should keep the icons only design because most of the time it is redundant (as in: GITHUB/github.com) or can be inferred from the context.
I also think that replacing the icon with a check mark isn't a good move because it feels like penalizing the verified links. The icons look nicer to me.

I don't have a better proposal right now but I would prefer we explore something simpler as in:
- keep the current design
- replace verified circling, with making the links in primary color instead of secondary (and have in a tooltip read "Verified" somewhere)

ZMYaro:
> we should keep the icons only design because most of the time it is redundant (as in: GITHUB/github.com) or can be inferred from the context.

I think it is worth noting not everyone knows every brand logo, and tooltips are not an option for mobile users.  Additionally, I would like to, in the future, be able to also apply icons to known domains regardless of text label (e.g., someone who links to personal and professional Twitter accounts could get the Twitter icon on each, but have the text labels “Personal” and “Professional”).

> I also think that replacing the icon with a check mark isn't a good move because it feels like penalizing the verified links. The icons look nicer to me.

I also wish there were a good way to have both the other icons and the verified checkmarks, but every layout I tried with both felt too busy.  I am still very open to suggestions, but this is the best I have found so far.

> have in a tooltip read "Verified" somewhere

I agree; I will add that in a moment (and ARIA label as well).


> replace verified circling, with making the links in primary color instead of secondary

Here is what that looks like:
image
image

ZMYaro:
And then this is the same implementation, but with the circled checkmark icon:
image
image

antfu:
Maybe we could:

- Remove the outline
- Use fixed green for the check instead of the primary color
- Move the "check" icon to the end of the URL (or the title)

ZMYaro:
@antfu Here are versions with a checkmark beside the label:
text_check
text_check_no_outline

antfu:
Then maybe put the check after the link might make it clear?

ZMYaro:
> Then maybe put the check after the link might make it clear?
Putting it next to the link could make it easier for someone to trick people with something like
image

Though one option could be to design a custom, more distinct checkmark badge (like Twitter has) rather than use one from Remix Icons.

ZMYaro:
BTW, since some folks on Masto asked, here is how it looks in green, and with the checkmark at the bottom-right of the icon:
image
image

antfu:
I like the right-bottom check design! 👍 Maybe we could remove the outline now, and process with this change?

---

I honestly don't think "tricking" is an issue we should concern, as they can already provide arbitrary links. If they have access to the website they can always get verified. Or you can have `Twitter: https://github.com/antfu`, showing Twitter's icon, while still getting verified.

ZMYaro:
> Maybe we could remove the outline now, and process with this change?
I like it, but I think a lot of folks are not a fan of the overlapping checkmark.

> If they have access to the website they can always get verified.
But with the example above, that makes it appear they own a site they do not—in that case, falsely appearing to be an official Google account.  Obviously you and I may pick up on the differences between the Remix circled checkmark icon and the attempted Unicode copy, but a lot of people will not.  Better, IMO, to either create a more distinct icon that cannot be copied, or else put the checkmark somewhere else.

antfu:
> I like it, but I think a lot of folks are not a fan of the overlapping checkmark.

I guess we can't find a design that works for everyone. I think we could move forward and have it, and see how ppl react to that, like how to do for any other things. /cc wdyt @patak-dev?

patak-dev:
I like the idea of removing the outline, and I'm ok with the checkmark overlapping the icon on the bottom-left corner. The mock I saw was with the checkmark on the top-right and it felt a lot more intrusive than this version.
I think we should move forward too 👍🏼

ZMYaro:
> I guess we can't find a design that works for everyone.

@antfu Less of an issue that not 100% of people are happy, and more that the majority of people I asked (admittedly, small sample size) preferred not overlapping.  Though maybe it would be beneficial to do another poll with the bottom-right overlap instead since, as @patak-dev said, that feels less intrusive.

> I think we should move forward too
@patak-dev I can try to clean this up further sometime this week.

I still need to figure out an elegant way to make the link green if we are going to do that.  Because the link is generated by the Mastodon API, we can't just add the `text-green` UnoCSS attribute to it.  Unfortunately, I am not versed in Uno, so I am not sure what to try besides just using regular CSS.

ZMYaro:
> I like the idea of removing the outline, and I'm ok with the checkmark overlapping the icon on the bottom-left corner.

@patak-dev I just submitted changes with that, as well as localized mouseover text (taken from the default Mastodon client).

I did try writing a function to inject the `text-green` attribute into the link HTML returned by the Mastodon API, but neglected to consider `ContentRich` would just strip it out 😅

edimitchel:
Small comment about the chip space before icon which could be reduced or the space between icon and text made coherent with the left padding:
image

Also, one opinion: on hovering a chip containing a link, hovering it should also add the underline on the link as the chip is fully clickable.

ZMYaro:
> Small comment about the chip space before icon which could be reduced or the space between icon and text made coherent with the left padding:

I might mess with the padding a little bit more, but I kind of don't want to make the chips even larger than they already are.

> on hovering a chip containing a link, hovering it should also add the underline on the link as the chip is fully clickable.

Funnily, this is also something I tried implementing by injecting the style into the link HTML returned by the Mastodon API, but neglected to consider `ContentRich` would just strip it out 😅

GalaxySnail:
What's the current status of this PR?

PR #1699, opened 2023-02-10
feat(editor): Respect default privacy setting of the account

### LINKED ISSUE #1040: Default post privacy setting

People uses "unlisted/followers" as default posting privacy. But Elk always set posting privacy setting as "public" by default and there is no setting or fetching web settings from server

### Suggested solution

Best solution: Fetch the web setting from home server. and use it as default posting privacy

### Alternative

Second solution: Add a setting to set a default posting privacy for Elk

### Additional context

_No response_

peterbud:
This PR changes default privacy setting of a post by considering the account's preferenced privacy setting when starting composing a post.

Also fixed an existing issue that empty drafts were not always cleared.

Fixes #1040 #1199

### DIFF
--- a/components/publish/PublishWidget.vue
+++ b/components/publish/PublishWidget.vue
@@ -27,7 +27,7 @@ const emit = defineEmits<{
 const { t } = useI18n()
 
 const draftState = useDraft(draftKey, initial)
-const { draft } = $(draftState)
+const { draft, isEmpty } = $(draftState)
 
 const {
   isExceedingAttachmentLimit, isUploading, failedAttachments, isOverDropZone,
@@ -48,6 +48,8 @@ const { editor } = useTiptap({
     set: (newVal) => {
       draft.params.status = newVal
       draft.lastUpdated = Date.now()
+      if (isEmpty)
+        clearEmptyDrafts()
     },
   }),
   placeholder: computed(() => placeholder ?? draft.params.inReplyToId ? t('placeholder.replying') : t('placeholder.default_1')),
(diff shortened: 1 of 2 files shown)


LittleSound (approved):

userquin:
@peterbud can you send 2 separated PR, this will be reverted, see #1730

peterbud:
@userquin sorry for the inconvenience.

However I'd need some help. The problem is that with the currant main branch the 'home' draft never gets cleared. Ever. See also #1199 . That means you will always have an 'elk-drafts' forever with the last privacy settings, and therefore the result of `getDefaultDraft` is never applied to a new post/toot.

The `clearEmptyDrafts` is called only at the unMounted hook of the `PublishFullWidget` but even that won't be called, as that compnent is never unmounted as I see, my guees is becaues `keepAlive:true`

I mistakenly tought the it's good to call `clearEmptyDrafts` once the post is empty, however it seems that a single `@` is also consiedered as empty, so that is not working as we see.

So where I'd need some advice: what is the best place / lifecycle hook to call `clearEmptyDrafts` to get rid of the empty drafts therefore preventing them danging there forever?

userquin:
> The `clearEmptyDrafts` is called only at the unMounted hook of the `PublishFullWidget` but even that won't be called, as that compnent is never unmounted as I see, my guees is becaues `keepAlive:true`

You should also add that logic with `onDeactivate`.

> So where I'd need some advice: what is the best place / lifecycle hook to call clearEmptyDrafts to get rid of the empty drafts therefore preventing them danging there forever?

Use the same logic in this PR, just check if it has only 1 character that is `@` and you're done: change `isEmptyLogic`

peterbud:
OK, I'll send a new PR. Many thanks!

Issue #1696, opened 2023-02-10
Docker image doesn't allow signup to Fediverse instance


maxheadroom:
Looks like the last change on the Dockerfile (Commit 331f517d758b477075a42b9b30660cb6670111fe) introduced running the service as non-root user. The result seems to be, that one can't sign in to any server anymore. When trying to sign in one only get's an error message that the service can't connect to the server.

With the default settings, there is no error message in the output of nodejs in the container. Even with ```DEBUG="*"``` set there is no error message or other information in the output of nodejs why the service can't connect to the fediverse instance.

If one removes the ```USER elk``` from the Dockerfile and re-creates the Docker image, it works just fine.

kkoyung:
I had this problem before. I guess it is due to the permission of persistent storage.

In the `Dockerfile`, it declares a persistent storage `/elk/storage`. When you create a contain without mounting volume for that directory, docker will automatically create an unnamed volume for it. By default, the unnamed volume created by docker is owned by root. However, the `Dockerfile` also set the default user to be UID 1000 and GID 1000. Therefore, that user cannot write file to /elk/storage.

To make it work, you have to mount a volume to /elk/storage which allow user with UID 1000 to write.

maxheadroom:
I'll give that a try and and update the docker-compose.yaml accordingly if it works.

b2cc:
@maxheadroom : just checked with latest `main` on Openshift, and the login error still persists for me on Openshift. Does it work for you on native docker now?

maxheadroom:
Yes, it works on native Docker for me. The PR that was merged here are basically the changes I had to make to the docker-compose.yml file to fix the issue for me.

b2cc:
Ok, just solved this  - was a weird routing issue on my end. For anyone stumbling across this, maybe this will help them, so I'm putting this out here.

My setup looks pretty much as in this comment: https://github.com/cert-manager/cert-manager/issues/2517#issuecomment-889961035 , and I ran into the same issue described there. Basically I have a router/gateway with a 1:1 NAT in front of the Mastodon and Elk apps. The Elk app runs in a private IP namespace (think 10.1.x.y.), next to Mastodon, both behind the Router.

What happened for me was that the Elk app was hitting the "internal" interface of the firewall when it tried to connect to Mastodon during the "Login" process, and the request was not routed correctly back to Mastodon, triggering this weird "Cannot connect to the server" error message, which was technically correct, but didn't tell me anything at that point.

[...]

PR #1701, opened 2023-02-10
fix: extended docker-compose.yml to addres Bug #1696

maxheadroom:
and provided some basic documentation for Docker self-hosting


patak-dev:
@b2cc @kkoyung would you help us to review this PR? Thanks!

kkoyung (approved):
The note in docker-compose.yaml file can help people to properly setup their own self-hosting Elk using docker. I saw people ask as in #1696 a few times. This PR should help.

patak-dev:
Thanks @kkoyung!

kkoyung:
@b2cc This PR gives extra information for people to properly setup their self-hosting Elk directly on Docker Engine. I tested on my server.

I am not familiar with Openshift. I suspect it may work differently from the original Docker Engine.

Is any folder named `v3` (or something similar) generated at `/elk/data` after trying to login? If not, the `node` process inside your container may be still not granted to write file to `/elk/data`. It yes, it could be caused by another problem.

b2cc:
@kkoyung : just checked and the `/elk/data` folder is still empty, although I'm positive that the process can write into it. I'll follow up in the other ticket, maybe it is a different issue.

PR #1705, opened 2023-02-10
feat: track scroll position

### LINKED ISSUE #294: Timeline position after clicking "Show x new items"
**Expected behaviour:** clicking to reveal new messages that have accumulated would keep my timeline position and let me scroll up to see new messages.

**Actual behaviour:** clicking on "Show x new items" scrolls you up to the present moment of a timeline. You would have to scroll back down to see these messages.

(Chose the Federated timeline so that new posts would appear more quickly.)

[screen-capture.webm](https://user-images.githubusercontent.com/5098874/205295772-39bbbb96-1c54-472c-8d5d-69abb82a6d55.webm)

### LINKED ISSUE #653: Remember user's last viewed location in feed
This is a side effect of some other navigation/UX issues I ran into, but when getting routed to the wrong feed (or the app minimizing itself randomly) when I reopen the app it refreshes my feed, and I lose the item I was reading.

Is it possible to temporarily store a session variable that holds the id of the post that is centered on the screen? 

I'm trying to think of the best way to track this, but I know it's something that Fedilab recently added to their app. As you scroll through your feed, it tracks your current position fairly well (not sure where the trigger point is), and if you go into a thread or suspend the app, it resumes where you were instead of refreshing the feed you're looking at. The refresh behavior is frustrating if you're reading something that is really interesting and for one reason or another get bumped to the wrong place and then lose your place in your feed.

### LINKED ISSUE #1538: Preserving the user's scrolling position when loading further posts
It would be nice if Elk could keep my scroll position the same when I click _Show x new items_.
At the moment I am taken to the top of the list and have to scroll down to find where I left off.

userquin:
This PR just tracks the scroll position on current page via client plugin, then, the plugin just tries to restore it with a timeout callback.

We can also exclude pages from scroll tracking using `noScrollTrack: true` in page meta, right now working in all timelines with the exception of `notifications`, seems the transformation applied took more than 600ms, I'll try to fix it using some callback or hook.

I also need to fix force scroll to top, right now all code in the plugin commented to test scroll tracking, I'll move the logic on new plugin to the existing one.

All settings page with no tracking.

supersedes #851 

closes #653
closes #294
closes #1538


userquin:
With latest changes we can remove `scroll-to-top` plugin and any call to `scrollToTop`

DataDrivenMD:
hey there - is this branch still active?

PR #1717, opened 2023-02-11
feat: add translation status to language settings

userquin:
This PR also includes missing translations for Spanish.


patak-dev (approved):

edimitchel:
Very interesting but I would suggest to make the translation status more discreet

userquin:
We can reduce the size, any suggestion @edimitchel ?

PR #1719, opened 2023-02-11
fix(a11y): increase status action size on small and medium screens

### LINKED ISSUE #1709: Tap targets for reply, boost, favorite, bookmmark just a little too small.
Elk on tablet and touch screen laptop screens I find the tap targets are too small, frequently resulting in misclicks. Google's recommended minimum is 48 x 48 px. 

The width and height of each of these buttons is currently under 38 for all 4 buttons.
screenshot showing the 4 buttons that appear below a post within Elk.

If visuals are an issue, it is possible to make the tap targets be that larger 48 px size without affecting the visual design. It just may require positioning the button elements absolutely.

userquin:
Maybe we can change it to `md:p-2`: changed, see title

closes #1709

### DIFF
--- a/components/status/StatusActionButton.vue
+++ b/components/status/StatusActionButton.vue
@@ -60,7 +60,7 @@ useCommand({
   >
     <CommonTooltip placement="bottom" :content="content">
       <div
-        rounded-full p4 sm:p-2
+        rounded-full p4 md:p-2
         v-bind="disabled ? {} : {
           'group-hover': groupHover,
           'group-focus-visible': groupHover,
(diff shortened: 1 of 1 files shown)


patak-dev:
The buttons don't look balanced for sm/md after this PR. Maybe we could increase their size in sm only, it is tricky because this is a more general issue across the app.

patak-dev:
I wonder if we should detect the pointer type, because people can use md in desktop.

userquin:
@patak-dev feel free to close this, we need to review a11y, selecting max font size on settings, button size area < 48px (~41px)

PR #1720, opened 2023-02-11
feat(i18n): update gl-ES

xmgz:
minor changes

I have not found "List" and "Hide reply count" in the .json. IDK what are "feature flags"


userquin:
can you check gl-ES here https://elk-docs.netlify.app/docs/guide/contributing#translation-status?

It is `list.*` not `List`

xmgz:
my error was to asume that language files are automatically populated with new strings from source, but I have to manually import them from en.json every time. Thanks. I'll now PR full translated file.

userquin (approved):

Issue #1374, opened 2023-01-21
Showing a status details with show network is very slow


edimitchel:
On mobile, with a slow network, when I want to show a status by clicking on it, I have to wait a while before getting the post and actually also the comments.

I would like to quickly get started details without comments at first, then getting all information lazily.

https://user-images.githubusercontent.com/2922851/213884814-eb80a0dd-a0d3-4bab-a916-e9700549bfc4.mp4

PR #1729, opened 2023-02-12
fix: better loading handling on status page

horvbalint:
This PR tries to fix the issue #1374 

It also fixes some other small bugs, like not visible time-line loading skeletons and I suspect that there was also a bug, where the page would not scroll to the focused status, when the loading finished.


edimitchel:
That's a lot better than what I experienced before!

Thank you!

pages/[[server]]/@[account]/[status].vue
```diff
@@ -28,28 +26,24 @@ const { data: context, pending: pendingContext, refresh: refreshContext } = useA
...
-  watchOnce(pendingContext, async () => {
-    await nextTick()
-    scrollTo()
-  })
-}
-
-const focusEditor = () => {
-  publishWidget.value?.focusEditor?.()
-}
+const publishWidget = ref()
```
Comment (edimitchel) [inline pages/[[server]]/@[account]/[status].vue:45]: Convention suggest to use ref(null)

```suggestion
const publishWidget = ref(null)
```

horvbalint:
@edimitchel I accepted your suggestion, but now typescipt gives an error. I am not sure why, this is my first time using the Composition API and from the docs it seems like this should just work.
Can you please suggest a solution?

edimitchel:
> @edimitchel I accepted your suggestion, but now typescipt gives an error. I am not sure why, this is my first time using the Composition API and from the docs it seems like this should just work. Can you please suggest a solution?

I could say that by having null, it will suggest the type as null, you should add a generic type to ref
`ref<HTMLElement | null>` but I think this won't be helpful.
You can revert your change (and ignore my previous comment)

edimitchel (approved):

patak-dev:
Thanks for the PR @horvbalint! I think it is important to help folks with bad connections. But after this PR we have a layout shift if you click on a reply. It first appears and then jumps. I think we should fix this before merging. I don't know how to fix this, if I understand correctly we are using the context to decide what is the best scroll position. Maybe we could wait for the previous messages in the thread to arrive before changing the scroll position.

horvbalint:
Thanks @patak-dev, I am afraid that there will be no good solution without layout shift (which I agree is also not a good solution). We can keep the focused status always on the top, but that will look silly, when it has only ancestors and no descendans.
Sadly we can't wait for the ancestors only, because they arrive together with the descendans. If we knew the number of ancestors, we could figure something reasonable out, but I did not found a method for that in masto.js. 
Maybe we could pass the ancestor count to the status page when coming from another timeline? So it could work OK when coming from a thread, and we would fall back to the old 'everything appears at once' solution, when opening a status link directly (so there is no previous page that could pass the ancestor count to the status page)
I will experiment with it and try to figure something out on the weekend, if nothing works the timeline sekelton is already making a difference imo, so we are still a bit better then before 😄

patak-dev:
It does feel better, indeed. Ok. Let's merge it so we get some testing in main.elk.zone. 

I think something that could be interesting for replies is to leave them in their place in the timeline where the user clicked, like if all other posts would disappear and then slide quickly to position once ancestors are available (trying to minimize how much it moves, even choosing not to move if there are enough ancestors for example)

PR #1731, opened 2023-02-12
feat(i18n): Emojimart localization and `fetching` entry at `en` …

mrcego:
…and `es` bases, and `es-419` locale

I'm lack of knowledge about advanced ways to import dynamically modules/files from third party dependency, so I'm summoning @userquin for help (lol).

More info: https://github.com/missive/emoji-mart#-internationalization

Updated: added `tooltip.pick_an_icon` entry too.

### DIFF
--- a/components/publish/PublishEmojiPicker.client.vue
+++ b/components/publish/PublishEmojiPicker.client.vue
@@ -23,10 +23,14 @@ async function openEmojiPicker() {
     })
   }
   else {
-    const { Picker } = await import('emoji-mart')
+    const [Picker, dataPromise, i18n] = await Promise.all([
+      import('emoji-mart').then(({ Picker }) => Picker),
+      import('@emoji-mart/data/sets/14/twitter.json').then((r: any) => r.default).catch(() => {}),
+      importEmojiLang(locale.value.split('-')[0]),
+    ])
 
     picker = new Picker({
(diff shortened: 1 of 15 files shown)


userquin:
maybe we change to this or split 2 last imports in separated entries:

imagen

userquin:
@patak-dev maybe we can revert my last commit and resolve 3 dynamic imports in parallel: check screenshot on previous comment

patak-dev (approved):
I think we can move forward with this, thanks for working on it!
Later on, it would be cool if we could use placeholders to avoid delaying showing the picker, it took a bit of time the first time I press the emoji button.

mrcego:
> I think we can move forward with this, thanks for working on it!
> Later on, it would be cool if we could use placeholders to avoid delaying showing the picker, it took a bit of time the first time I press the emoji button.

It should be showing a _fetching_ text instead. Maybe my mistake with translation?

patak-dev:
Oh, maybe my bad, and it was already taking time to load at first because it is loaded lazily the first time, no?

mrcego:
> Oh, maybe my bad, and it was already taking time to load at first because it is loaded lazily the first time, no?

I was able to reproduce it but setting 4G slow network. Interesting. Placeholder is shown but dissapears seconds later with emojis loading one by one.

userquin:
> Oh, maybe my bad, and it was already taking time to load at first because it is loaded lazily the first time, no?

yes, we can keep json files with the promise, but the picker will also await to be resolved (twitter.json)

PR #1733, opened 2023-02-12
feat(editor): Respect default privacy setting of the account (2nd)

peterbud:
This is a second try for #1699, based on the discussion with @userquin


composables/masto/statusDrafts.ts
```diff
@@ -30,7 +30,7 @@ export function getDefaultDraft(options: Partial<Mutable<mastodon.v1.CreateStatu
     params: {
       status: status || '',
       inReplyToId,
-      visibility: visibility || 'public',
+      visibility: currentUser.value?.account.source.privacy || visibility || 'public',
```
patak-dev, line 33: There was another regression here, when you answer to a private post, the reply should be private by default.

The default privacy only should be taken into account if it makes the post more private than the replying to post

peterbud: Thanks. Can you check the updated logic?

patak-dev (approved):

PR #1745, opened 2023-02-13
fix: Re-enable user card hover over mentions in posts

### LINKED ISSUE #1718: Show user card when hovering over tagged (profiles) links.
User cards are shown when the cursor hovers over a profile photo

image

But not when it hovers over a tagged link:

image

My proposal is to add this last behavior, just like Twitter web does:

image

peterbud:
This PR re-enables to display user cards when the user hover overs a mention tag.

This functionality was there for a while but then disappeared as mention links became relative and they did not start with http anymore. (Prerequisite for showing the user card is that the link matches `UserLinkRE` regexp)

Fixes #1718

### DIFF
--- a/composables/content-parse.ts
+++ b/composables/content-parse.ts
@@ -618,7 +618,7 @@ function createTransformNamedMentions(mentions: mastodon.v1.StatusMention[]) {
       const href = node.attributes.href
       const mention = href && mentions.find(m => m.url === href)
       if (mention) {
-        node.attributes.href = `https:\/\/${currentServer.value}/@${mention.acct}`
+        node.attributes.href = `/${currentServer.value}/@${mention.acct}`
         node.children = [h('span', { 'data-type': 'mention', 'data-id': mention.acct }, `@${mention.username}`)]
         return node
       }
--- a/composables/content-parse.ts
+++ b/composables/content-parse.ts
@@ -618,7 +618,7 @@ function createTransformNamedMentions(mentions: mastodon.v1.StatusMention[]) {
       const href = node.attributes.href
       const mention = href && mentions.find(m => m.url === href)
       if (mention) {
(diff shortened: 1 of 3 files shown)


composables/content-parse.ts
```diff
@@ -618,7 +618,7 @@ function createTransformNamedMentions(mentions: mastodon.v1.StatusMention[]) {
       const href = node.attributes.href
       const mention = href && mentions.find(m => m.url === href)
       if (mention) {
-        node.attributes.href = `/${currentServer.value}/@${mention.acct}`
+        node.attributes.href = `https:\/\/${currentServer.value}/@${mention.acct}`
```
patak-dev, line 621: The `href` being relative to the deployment is intentional, as we want the user to stay in Elk instead of going to the original server. We need to find a way to fix this without modifying this link.

patak-dev (changes requested):

composables/content-parse.ts
peterbud: The `AccountHoverWrapper` is inserted if the href matches the regexp `UserLinkRE` here:
[UserLinkRE](https://github.com/elk-zone/elk/blob/main/composables/content-render.ts#L88-L94)

Unfortunately `UserLinkRE` starts with http/s, so the relative links are failing, and therefore no user card on hover.

What would be in your view the best solution:
* Add a new regexp without http/s at the start and test against it and append the AccountHoverWrapper
* Change the `UserLinkRE`  regexp (but that is used at another place as well, so I'm not preferring it)
* something else?

peterbud: So I have reverted back, and just changed the regex to work with or without http at the beginning. Looks like working.

patak-dev:
Sorry it took me so long to get back to this one @peterbud! We've been a bit busy with Vite 4.3 😄 
Let's merge it, this is a big one.

patak-dev (approved):

PR #1748, opened 2023-02-14
chore: use pnpm run instead of nr in scripts

patak-dev:
Using `nr` in the post-install script is breaking Codeflow.


patak-dev:
Oh, you are right. The issue seems to be elsewhere, I'll check.

PR #1751, opened 2023-02-14
fix(i18n): use American spelling

valtlai:
For consistency with the other occurrences of _favo(u)rite_

Menu item “Show who favourited and boosted” in the Elk user interface

### DIFF
--- a/locales/en-GB.json
+++ b/locales/en-GB.json
@@ -6,6 +6,9 @@
     "favourite": "Favourite",
     "favourited": "Favourited"
   },
+  "menu": {
+    "show_favourited_and_boosted_by": "Show who favourited and boosted"
+  },
   "nav": {
     "favourites": "Favourites"
   },
--- a/locales/en.json
+++ b/locales/en.json
@@ -214,7 +214,7 @@
     "open_in_original_site": "Open in original site",
     "pin_on_profile": "Pin on profile",
     "share_post": "Share this post",
-    "show_favourited_and_boosted_by": "Show who favourited and boosted",
+    "show_favourited_and_boosted_by": "Show who favorited and boosted",
     "show_reblogs": "Show boosts from {0}",
     "show_untranslated": "Show untranslated",
     "toggle_theme": {


boehs:
You've identified a larger consistency problem than this single change, in that file, I see *many* mismatches. In any case, I believe on a project level, we do not use American spelling. Being said, we have a locale file specifically dedicated to `en-US`, here: https://github.com/elk-zone/elk/blob/main/locales/en-US.json

So to be clear, unless @patak-dev has some other opinion, I'd like to see the following:

1. Change all instances of `favorite` to `favourite` in `en.json`
2. Populate `en-US` with freedom spelling

boehs:
> This change seems to be ok, the default locale is `en-US` and it is configured using base `en.json` file (that's why `en-US` file is empty).
> 
> @valtlai can you add deleted/updated entry on the corresponding entry in `en-GB.json` file?.

Why are the locale keys for the `en` base in GB while most of their values are in en-US

valtlai:
> @valtlai can you add deleted/updated entry on the corresponding entry in `en-GB.json` file?.

Sure, done.

userquin:
@boehs where do you see that? All keys are in `en.json` file, `en-US.json` file is empty and `es-GB.json` file only contains variants from `en-US`.

If the default language was `en-GB`, then the base would still be `en.json` but with the English variants, and the `en-US.json` file would not be empty, it would just have the same messages as it does now `en-GB.json` but for the American variant.

`i18n` module was changed to support this behavior, if files entry found on locale configuration, all messages will be merged using the order specified in files prop, and so, `en-US` can be empty (`files: ['en.json', 'en-US.json' ]), and  `en-GB` (`files: ['en.json', 'en-GB.json' ]) will be merged properly, check `config/i18n.ts` module (check countryLocaleVariants on L11). 

This change was introduced to reduce the process of maintaining N files where 99% of the messages are the same. We only need to change some messages when they vary from the base used.
Another example is `es-ES`, which is even worse, (we have 19 variants, but reduced to the generic Latin American `es-419`), (Spanish is the base `es`) and `es-419` (Spanish for Latin America).

boehs:
> @boehs where do you see that? All keys are in `en.json` file, `en-US.json` file is empty and `es-GB.json` file only contains variants from `en-US`.

Right. Look at this in the `en.json` file
image

boehs:
> Why are the locale keys for the `en` base in GB while most of their values are in en-US?

Hence why I said

> Why are the locale keys for the en base in GB while most of their values are in en-US?

This should probably be fixed

userquin:
@boehs I don't think so, but if you want to change it, PR welcome man, no problem.

I think anyone can understand any key that contains favorite or favourite, it would be problematic if we used keys in other languages (I'm Spanish, and I ALWAYS use keys in English and in all my source code, be it JS, TS, Java, C,... or any other programming language: I've seen people using ñ in variable names 😅)

userquin (approved):

userquin (approved):

PR #1757, opened 2023-02-15
feat: cache avatars in workbox

### LINKED ISSUE #1713: Cache Avatar Images in Client (workbox)
When scrolling through the timeline it is common to hit avatars for the same user multiple times, and usually you follow your community that does not change that much

I would like to propose to cache the avatars on first load so scrolling trough the timeline does not fetch the avatars via internet everytime

At #VueAmsterdam2023 the avatars did load slowly and so I thought this can improved

Shinigami92:
closes #1713

### DIFF
--- a/service-worker/sw.ts
+++ b/service-worker/sw.ts
@@ -3,7 +3,7 @@
 import { cleanupOutdatedCaches, createHandlerBoundToURL, precacheAndRoute } from 'workbox-precaching'
 import { NavigationRoute, registerRoute } from 'workbox-routing'
 import { CacheableResponsePlugin } from 'workbox-cacheable-response'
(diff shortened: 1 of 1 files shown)


Shinigami92:
I'm not sure if this is correct...
Somehow on my desktop Chrome I get "disk cache" hits and when I debug with throttling and configure it to EDGE it seems to work fine without this fix

However on iOS iPhone you can see in the video in the referenced issue that they are not loaded from cache and every avatar is fetched again and again :thinking:

Shinigami92:
Okay YES, it is really confusing myself but I can confirm this definitely works now on my iPhone :astonished: 
Every second occurring image is not loaded multiple times but instantly appear, even after a page reload they are then instantly loaded

Shinigami92:
did not work as expected

PR #1758, opened 2023-02-15
refactor: inject pwa client plugin from module

danielroe:


userquin:
We should protect Pwa** sfc, we're using them directly and so it will fail on dev, check `PwaPrompt` usage on default layout

danielroe:
I think they're all protected with `v-if`...

userquin:
you're right, `$pwa?.`

pages/share-target.vue
```diff
@@ -8,7 +8,7 @@ definePageMeta({
 
 useWebShareTarget()
 
-const pwaIsInstalled = process.server ? false : useNuxtApp().$pwa.isInstalled
+const pwaIsInstalled = process.client && !!useNuxtApp().$pwa?.isInstalled
```
userquin, line 11: the middleware runs before? if so, you can replace it with `useNuxtApp().$pwa.isInstalled`

userquin (approved):

PR #319, opened 2022-12-03
feat: basic keyboard shortcuts

hartmut-co-uk:
implements https://github.com/elk-zone/elk/issues/39

### TODOs
- [x] MVP
  - [x] add 'Keyboard Shortcuts' modal with instructions
      - [x] open via `?` -> `shift + ?`
      - [x] close modal via `shift + ?` / `esc`
  - [x] add first basic shortcut: `c` for 'Compose'
  - [x] add first two _sequence_ navigation shortcuts:
    - [x] `g..h` Open Home Timeline
    - [x] `g..n` Open Notifications
  - [x] technical details
    - [x] enable/disable based on _isAuthenticated_ status
    - [x] disable when focused on INPUT/TEXTBOX/contenteditable
  - [x] Toot _active_/_open_(?/_hover_?) specific actions
    - [x] Favourite `f`
    - [x] Boost `b`
  - [ ] ~navigate timeline prev/next status~ (moved to a future iteration due to complexity)
    - [ ] ~Next status `j`~
    - [ ] ~Prev status `k`~
  - [x] styling & design
    - [x] basic design 'Keyboard Shortcuts' modal
    - [x] responsive
- [x] define list of shortcuts to provide with this PR (limited number for a first iteration?)
- [x] use I18n (started off with _en-US_ for prototyping)
- [x] define next steps / create follow-up issue(s)   
        -> https://github.com/elk-zone/elk/issues/1855
        -> https://github.com/elk-zone/elk/issues/1856
        -> https://github.com/elk-zone/elk/issues/1857
        -> https://github.com/elk-zone/elk/issues/1858

### Options for improvement
- [ ] performance / ressource utilization
  - [ ] disable event listeners for mobile devices (without hardware keyboard)
- [ ] UI/UX
  - [ ] ??? for page navigation (e.g. `g .. h` / `g .. n`) show a notification message ~"Going to home" by @edimitchel 
- [ ] code / architecture
  - [ ] simplify / find better solution for `j`/`k` navigation.
         Current solution uses DOM & querySelectors, hard to maintain / comes with high risk of regressions while working on other features..
  - [ ] export composables for each keybinding, components self-register triggers, centralised tracking of listeners ([ref discord](https://discord.com/channels/1044887051155292200/1047780695105343530/1049603332404154388)) by @QiroNT

PR #1763, opened 2023-02-15
feat: allow for navigating the timeline with j and k

JohnyTheCarrot:


patak-dev:
Did you check the discussion in this PR https://github.com/elk-zone/elk/pull/319 ? Would you comment here on this approach so we leave a trace of the reasoning for it here?

JohnyTheCarrot:
Gonna close this for now, will see if I revisit it at a later date.

PR #1766, opened 2023-02-16
fix: Spanish translation for request to follow is incorrect

santiagozky:
The current translation implies a request to follow you has been made (instead of requesting to follow someone). The phrase was also not stated as a verb, as the other actions.


patak-dev (approved):
Gracias!

PR #1782, opened 2023-02-17
chore: remove `useHeadFixed` workaround

### LINKED ISSUE #1776: Elk.zone Home tab shows incorrect name
This is my first issue here, bear with me
this is semi-consistent behavior (thus why i am making this issue) and I'm unsure if this has been addressed or not nor do I have a concrete fix
Currently on Firefox with Windows 10

How to replicate (very hit or miss):
Go to https://elk.zone/home
Click someone's profile, I specifically used someone from my own instance.
Title changes to "Posts (Username here)"
Go back to home
Title of the tab will get "stuck" on a name, and wont change back.
image
(The solid black tab is the currently selected tab, and the "Home | Elk" tab is a demonstration on what this tab's title should look like.)

If this is already been posted before, I'm incredibly sorry, you may feel free to close this issue if needed.

userquin:
This PR updates Nuxt to 3.2.2 and `@unhead/ssr` to `1.0.22` removing the workaround.

closes #1776

### DIFF
--- a/components/nav/NavTitle.vue
+++ b/components/nav/NavTitle.vue
@@ -29,7 +29,7 @@ router.afterEach(() => {
       @click.prevent="onClickLogo"
     >
       <NavLogo shrink-0 aspect="1/1" sm:h-8 xl:h-10 class="rtl-flip" />
-      <div hidden xl:block text-secondary>
+      <div v-show="isHydrated" hidden xl:block text-secondary>
         {{ $t('app_name') }} <sup text-sm italic mt-1>{{ env === 'release' ? 'alpha' : env }}</sup>
       </div>
     </NuxtLink>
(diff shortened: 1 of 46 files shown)


userquin:
I need to fix tabs, not working properly...

LittleSound (commented):
Maybe we can return the content with correct language when rendering on the server side, therefore we will not have to add a lot of `isHydrated` to avoid title and UI text flickering.

However, doing in such way would require the server know what language the user is currently choosing. There are solutions for that. For example, we can get language by fetching it from the user's Mastodon server, afterwards, we can cache and include the user's chosen language in Elk's cookie and sending it to the server.

I understand that the README promises not to use cookies, so I came up another idea. We can display an opening page when the site opens, and then display the final screen after getting enough data to display the screen correctly (similar to what many mobile apps do).

I know this topic may be a bit different from the problem this PR description is trying to solve, but I've seen attempts in the code to solve the problem I'm discussing, so allow me to give my opinion here.

patak-dev:
Closing as #1939 has been merged

PR #1787, opened 2023-02-18
chore: upstream fixed a issue in `useHead`

LittleSound:
`unhead` fixes an issue in `useHead` (PR: https://github.com/unjs/unhead/pull/96). I wrote a function for Elk as a temporary solution (PR: #199) before fixing this problem upstream. Now that it has been fixed, I will update the dependency to the repaired version in this PR and remove and modify the code from the previous temporary fix.

### DIFF
--- a/components/status/StatusDetails.vue
+++ b/components/status/StatusDetails.vue
@@ -22,7 +22,7 @@ const createdAt = useFormattedDateTime(status.createdAt)
 
 const { t } = useI18n()
 
-useHeadFixed({
+useHead({
   title: () => `${getDisplayName(status.account)} ${t('common.in')} ${t('app_name')}: "${removeHTMLTags(status.content) || ''}"`,
 })
 
--- a/composables/setups.ts
+++ b/composables/setups.ts
@@ -12,7 +12,7 @@ export function setupPageHeader() {
     return acc
   }, {} as Record<string, Directions>)
 
-  useHeadFixed({
+  useHead({
     htmlAttrs: {
       lang: () => locale.value,
       dir: () => localeMap[locale.value] ?? 'ltr',
--- a/composables/vue.ts
+++ b/composables/vue.ts
@@ -1,8 +1,5 @@
 import type { ComponentInternalInstance } from 'vue'
 import { onActivated, onDeactivated, ref } from 'vue'
-import type { ActiveHeadEntry, HeadEntryOptions, UseHeadInput } from '@vueuse/head'
-import type { HeadAugmentations } from '@nuxt/schema'
-import { useHead } from '#head'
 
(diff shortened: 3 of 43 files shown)


userquin:
@LittleSound there is PR in progress #1782

antfu (approved):

userquin:
@antfu changing only the workaround will not work, check my PR, we also need to fix the problem having `keepAlive`, since there is a lot of title changes when you traverse the app: for example, navigate ntofications tabs, go to your account and navigate followers, following, back to posts and navigate tabs... the title will have a race condition

LittleSound:
Thanks for trying to fix this, I'll take some time to look at #1776 and #1782 you said to see what the problems are so far.

userquin:
https://streamable.com/e1jmkr

LittleSound:
@userquin 

> changing only the workaround will not work, check my PR, we also need to fix the problem having keepAlive, since there is a lot of title changes when you traverse the app: for example, navigate ntofications tabs, go to your account and navigate followers, following, back to posts and navigate tabs... the title will have a race condition.

I just updated nuxt to 3.2.2 and used the useHead function to fix the issue raised in 1776, if there are other issues with the useHead function, can you provide a specific way to reproduce them?

> EDIT: `i18n::setLocale` is async, we need to await on the client plugin to avoid flickering (when user `lang !== 'en-US'`) on F5

I think i18n should have the same behavior on server and client, and it should not be asynchronous and causing flickering. Server should be able to read the user's current language in the cookie and return the correct content, at least it works this way in the project I'm working on.

userquin:
@LittleSound

Just switch to another locale (`!== 'en-US'`) and refresh page, you will see the title and nav bar flickering (en-US then user lang).

About i18n on server, we can enable cookie, but the problem is the hydration, if i18n restores properly user locale I guess we don't need the `useHydratedHead` hack.

LittleSound:
@userquin 
Maybe the flickering problem caused by i18n should be a different issue from issue #1776

> About i18n on server, we can enable cookie, but the problem is the hydration, if i18n restores properly user locale I guess we don't need the useHydratedHead hack.

Maybe we can try to solve the problem in this direction, it looks simpler instead of adding `isHydrated` in many places, adding these logic branches obviously increases the developer's mental burden and code complexity.

userquin:
> @userquin Maybe the flickering problem caused by i18n should be a different issue from issue #1776
> 

:thumbsup: 

> Maybe we can try to solve the problem in this direction, it looks simpler instead of adding `isHydrated` in many places, adding these logic branches obviously increases the developer's mental burden and code complexity.

The same complexity we have rn: replace `useHeadFixed`  with `useHydratedHead`... ;)

PR #1792, opened 2023-02-18
fix: handle frozen state

### LINKED ISSUE #1791: Elk does not refresh on Firefox on Android across network changes
Although #1584 was closed and I do not have permissions to re-open it, the problem persists.

When I move between wifi and data, Elk (whether in a browser tab or as a PWA) stops refreshing, and rotating the phone to landscape and clicking the Elk logo also doesn't refresh. I have to use the browser refresh menu option. The browser refresh menu option is unavailable in a PWA, so Elk is still unusable as a PWA. I've had to revert to using it only in a browser tab where I have access to the browser refresh feature.

This has been consistent since https://github.com/elk-zone/elk/pull/1658 was resolved, so that was not the only problem here.

I still have to run Elk in a browser tab, and when I change from WiFi to cell networks, or vice-versa, I have to refresh the tab from the browser menu in order to start seeing new content again.

All of this experience is on main.elk.zone, so this is not from me being still on an old version.

userquin:
This shoudn't fix #1791

### DIFF
--- a/page-lifecycle.d.ts
+++ b/page-lifecycle.d.ts
@@ -1,5 +1,5 @@
 declare module 'page-lifecycle/dist/lifecycle.mjs' {
-    type PageLifecycleState = 'pageshow' | 'resume' | 'focus' | 'blur' | 'pagehide' | 'unload' | 'visibilitychange' | 'freeze'
+    type PageLifecycleState = 'active' | 'passive' | 'hidden' | 'frozen' | 'terminated'
 
     interface PageLifecycleEvent extends Event {
         newState: PageLifecycleState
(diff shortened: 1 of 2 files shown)


patak-dev:
@userquin this should fix the issue you linked? Or do you mean we should merge this one, but keep the referenced issue open?

userquin:
@patak-dev yeah, we shoul dkeep open the issue, sorry, it is another problem with network switching

Before merging this I'll send a DM to @edimitchel to check the preview on FF.

userquin:
@patak-dev I guess this can be merged

johnsonm:
I see that #1791 was closed anyway. Was that automation because the bot saw "fix" next to "#1791"? If it shouldn't have been closed, can someone with project permissions re-open it? ☺

userquin:
> I see that #1791 was closed anyway. Was that automation because the bot saw "fix" next to "#1791"? If it shouldn't have been closed, can someone with project permissions re-open it? ☺

Reopened

Issue #1797, opened 2023-02-19
POST requests to /api/* endpoints fail on Cloudflare Pages


DataDrivenMD:
All non-`GET` requests fail when the static app is deployed to Cloudflare Pages. This is true regardless of the `NITRO_PRESET` value that is passed during the build. This is happening because `/functions` output generated by Nitro returns HTTP 405 (Method not allowed) for any POST requests.

Affects: Elk <=0.7.4

DataDrivenMD:
Closing this issue as the fix to Nitro has been merged and will be released soon. For context, see https://github.com/unjs/nitro/pull/965

PR #1798, opened 2023-02-19
Fix: POST requests to /api/* endpoints fail on Cloudflare Pages

DataDrivenMD:
Fixes: https://github.com/elk-zone/elk/issues/1797
[X] Adds Dev dependency `@cloudflare/workers-types`
[X] Adds support for `cloudflare-kv-binding` storage driver
[X] Implement router for `/api/*` endpoints by adding Cloudflare Pages Functions
[X] Fix status codes in `_redirects` file to make it compatible with CF Pages
[X] Add `_routes.json` to ensure that Functions are only triggered by specified routes so as to optimize cost savings
[X] Add example `wrangler.toml` that can be used for local development
[X] Update `.gitignore` to exclude deployment files that may contain secrets/sensitive information


danielroe:
This is impressive! But it shouldn't be necessary. It's a bug in https://github.com/unjs/nitro that we need to resolve there rather than patching it specifically in Elk.

DataDrivenMD:
> This is impressive! But it shouldn't be necessary. It's a bug in https://github.com/unjs/nitro that we need to resolve there rather than patching it specifically in Elk.
> 
> Let's focus on fixing the issue in Nitro first and then no platform-specific code will be needed in the Elk codebase.

OK by me if you want to close the PR without merging. The reason I went ahead and wrote the patch is two-fold:
1. I didn't know how long it would be until the issue in Nitro might be addressed
2. The code generated by Nitro is not cost-efficient. The functions I wrote for my own deployment (and contributed via this PR) minimize costs by avoiding unnecessary function invocations for static assets. The code, here, will significantly increase the runway for deployments under the _Free Tier_ and reduce billable invocations on paid tiers by >95%.

danielroe:
Totally understand. But hopefully https://github.com/unjs/nitro/pull/965 should resolve the issue with assets hitting the endpoint, and my suggested changes to https://github.com/unjs/nitro/pull/796 should resolve the error. 🤞

DataDrivenMD:
Hey @danielroe - just saw those changes to the Nitro repo- thank you for excluding static assets from the function call. Closing this PR since it looks like we'll be able to make Elk work without the custom patch. Thanks for expediting the fix! 

Aside: I didn't realize you were a contributor to Nitro as well (you're prolific!)

PR #1802, opened 2023-02-19
feat: content filter settings

JohnyTheCarrot:


JohnyTheCarrot:
Ready for a sort of check-in review. I.e, do you like where I'm going with this? Are there any code patterns you see that you'd like to see addressed? While I'm not done yet it'd be good to check in, so please **do not merge yet**.

Known issues:
* No form validation/error handling
* The filters list doesn't update after going to edit a specific filter and going back, no idea how to fix this so suggestions are welcome

patak-dev:
@ayoayco I think you are one of the best ones to check this one out, at least from a UI level.

One thing I would change is to move the Filters out of Preferences and directly into their own top-level section. I think it is a very important feature that should be more discoverable to others. Other than that, I think we should move forward and keep iterating.

JohnyTheCarrot:
Also, the Mastodon API doesn't seem to tell the client what timezone expires_at is using.. Kinda makes it hard to use in any way..

ayoayco:
Wow great work so far!

Played around with the UI for a bit. I like that it is under wellness setting, I think that makes sense. I was able to create, update and delete, so I think functionally we have what is needed.

One thing I'm thinking of to improve is how to make it easy even for people not yet familiar with the feature of Mastodon. The icons may not be immediately clear for people who are new. Specifically the "should match whole word" toggle; I think could be a checkbox with words 🤔

I have no design specification, but my general thought is, could we make this UI such that it also educates people of the capabilities of mastodon filters?

JohnyTheCarrot:
> Wow great work so far!
> 
> Played around with the UI for a bit. I like that it is under wellness setting, I think that makes sense. I was able to create, update and delete, so I think functionally we have what is needed.
> 
> One thing I'm thinking of to improve is how to make it easy even for people not yet familiar with the feature of Mastodon. The icons may not be immediately clear for people who are new. Specifically the "should match whole word" toggle; I think could be a checkbox with words thinking also maybe we can have a default text when nothing is selected for the expiry. Is the default "Never" I think?
> 
> I have no design specification, but my general thought is, could we make this UI such that it also educates people of the capabilities of mastodon filters?

> I think could be a checkbox with words :thinking: 

What exactly would you suggest?

JohnyTheCarrot:
Also, do you prefer the change to the expires at field? Is this what you meant?

ayoayco:
> Also, do you prefer the change to the expires at field? Is this what you meant?

Yes, I think seeing "Never" is better than no default text.

Some suggestions after I tried it again:
1. Earlier, I was not able to see the tooltip "Match whole word" and the other icons beside the keywords because I was on mobile. I think generally tooltips are not good for accessibility because they don't show up on mobile screens. Can we explore showing the text here not in a tooltip?
1. Same as above, in the filter lists menu, I saw we have a tooltip that shows "Filters in home, piublic, and etc...". I think this is important and should be a secondary text under the filter title. I think we can explore having 3 lines there: Title, Keywords, and this list of timelines where the filter is applied--make sure lines are truncated though :)
1. Would also love to see the number of keywords. So maybe in the filter list you can do "30 keywords: twitter, elon, musk..."
1. Also check why sometimes they show "No keywords filtered at this time." even after user created keywords
1. I noticed in other preferences, options that are not checked are using secondary color. It would be good to implement it here too to have a uniform experience.

JohnyTheCarrot:
@ayoayco How about this?

Thanks to @Dylan-DPC for the suggestion to use tags.
image
image

ayoayco:
Love this so far!

JohnyTheCarrot:
Still working on this, but my former coworker has decided to help out with some designs, so I'm waiting on those. :)

JohnyTheCarrot:
Aaaaa, sorry. College suddenly happened. Will try and get to this ASAP

PR #1806, opened 2023-02-21
Implement Cloudflare KV storage binding

DataDrivenMD:
This PR:
- [X] Implements `cloudflare-kv-binding` driver from `unstorage` library
- [X] Activates the driver when `NUXT_STORAGE_DRIVER=kv-binding`

When active, the driver will use whichever namespace is bound to the `STORAGE` environment variable, which is injected by Cloudflare at runtime. Deployments that use this driver will ignore the following environment variables: `NUXT_CLOUDFLARE_ACCOUNT_ID`
`NUXT_CLOUDFLARE_NAMESPACE_ID`
`NUXT_CLOUDFLARE_API_TOKEN`
Therefore,
- [X] `.env.example` has been updated to clarify which parameters need to be passed according to the deployment environment/platform

Mocking isn't possible for this driver because the KV namespace is bound at runtime and injected into the runtime environment by Cloudflare. So,
- [X] Added `wrangler.example.toml` to enable local development and (manual) testing

As `wrangler.toml` and Cloudflare `wrangler` artifacts may include secrets,
- [X] updated `.gitignore` to safeguard against inadvertent disclosure of secrets


Kitanit:
add support for [Vercel KV](https://vercel.com/docs/storage/vercel-kv)

PR #1808, opened 2023-02-22
fix: properly parse bolded italics

### LINKED ISSUE #1804: Bold and Italic cannot be displayed
Before posting:
image

After posting:
image

for example: https://elk.zone/universeodon.com/@yenche123/109895545528086301

zaidhaan:
It would appear that the current markdown replacer function for bolded italics (line 448) that converts a match into an ultrahtml `h()` object doesn't expect an array in `c`'s place.
https://github.com/elk-zone/elk/blob/29b37e67c7675d65a554aac05f742922771484bf/composables/content-parse.ts#L447-L455

When supplied with c = `['foo']`, it simply puts `['foo']` in the elements children, which isn't properly parsed by Vue. If instead, it is in the form `'foo'`, it's parsed properly.

I'm not too familiar with ultrahtml but this seems to work for me and seems to be the root cause as far as I can tell.

Before: (Screenshot taken from mocked dev, but I'd guess viewing [this](https://universeodon.com/@yenche123/109895545528086301) would yield similar results)
image

After:
image

Fixes #1804.

### DIFF
--- a/composables/content-parse.ts
+++ b/composables/content-parse.ts
@@ -445,7 +445,7 @@ function replaceCustomEmoji(customEmojis: Record<string, mastodon.v1.CustomEmoji
 }
 
 const _markdownReplacements: [RegExp, (c: (string | Node)[]) => Node][] = [
-  [/\*\*\*(.*?)\*\*\*/g, c => h('b', null, [h('em', null, c)])],
+  [/\*\*\*(.*?)\*\*\*/g, ([c]) => h('b', null, [h('em', null, c)])],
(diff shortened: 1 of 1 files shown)


zaidhaan:
Ah, so you guys have preview links, nice. [this link](https://deploy-preview-1808--elk-zone.netlify.app/@yenche123/109895545528086301) demonstrates this fix ([original](https://universeodon.com/@yenche123/109895545528086301), and [elk](https://elk.zone/universeodon.com/@yenche123/109895545528086301)).

Issue #1279, opened 2023-01-18
Tablet (landscape) doesn't display the full UI


markpinero:
### Pre-Checks

- [X] Check that there isn't [already an issue](https://github.com/elk-zone/elk/issues) that reports the same bug to avoid creating a duplicate.
- [X] Check that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/elk-zone/elk/discussions) or join our [Discord Chat Server](https://chat.elk.zone).
- [X] Providing a screenshot or video to reproduce the issue or show visually what was meant.
- [X] I am willing to provide a PR.

### Describe the bug

When I am using a tablet in landscape, full UI isn’t displayed because 100vh doesn’t account for dynamic viewport heights [(reference)](https://web.dev/viewport-units/).

Solutions:
- For modern browsers, use dynamic viewport units (referenced in link above).
- If not supported, add event listener to track visible viewport.

### Reproduction video or screenshot

https://user-images.githubusercontent.com/1432193/213083092-3689afee-9d2e-467c-b70b-5d8ac99d7cca.MP4

markpinero:
merged!

PR #1817, opened 2023-02-24
fix(default): update layout to use dynamic vh

markpinero:
This PR fixes issue #1279 which overflows the user avatar section and footer. The issue is documented [here](https://web.dev/viewport-units/), where 100vh will take 100% of viewport height regardless of UI chrome. This is solved using the `dvh` unit which changes based on the UI.

This issue only affects tablets at >1280px width (eg: 11/13in iPad Pro).

I'm not very knowledgable in unocss syntax, but copied the `scrollbar-hide` implementation since it applies a non-standard CSS rule-set. Reason being is if they are applied directly to tag (eg: `<div h-screen h-100dvh>`, it can not apply correctly if `h-screen` is after `h-100dvh` in the stylesheet.

The CSS rule I created merges the two to read like so:

```
.h-100dvh {
  height: 100vh;
  height: 100dvh;
}
```

Where it will apply 100dvh where it is [supported](https://caniuse.com/mdn-css_types_length_viewport_percentage_units_dynamic), and 100vh elsewhere.


patak-dev (approved):
Thanks! Sorry it took us a while to merge it

PR #1824, opened 2023-02-25
fix: resize large images attachments on upload

### LINKED ISSUE #1702: Cant upload larger images supported by other clients
Uploading this image and others at the same resolution works on the default clients but fails consistently with elk, I cant see the error string anywhere in the elk code, it seems very strange for a client to enforce a size / resolution limit and not just leave that to the server so not 100% sure whats happening here. 

Screenshot 2023-02-10 at 12 23 25

zaidhaan:
Mastodon's API has a [4096x4096px file size](https://github.com/mastodon/mastodon/blob/730bb3e211a84a2f30e3e2bbeae3f77149824a68/app/models/concerns/attachmentable.rb#L8) limit on image attachments, its web client however simply resizes larger images on the client-side to an acceptable size for the API request.

This PR brings this functionality to elk. The approach is [based on theirs](https://github.com/mastodon/mastodon/blob/730bb3e211a84a2f30e3e2bbeae3f77149824a68/app/javascript/mastodon/utils/resize_image.js), though re-written by myself in a much more stripped down way.

To test, try uploading a large attachment [like this one](https://unsplash.com/photos/7f4WaNrR4uo/download?force=true).

Fixes #1702

### DIFF
--- a/composables/masto/publish.ts
+++ b/composables/masto/publish.ts
@@ -122,7 +122,7 @@ export function useUploadMediaAttachment(draftRef: Ref<Draft>) {
 
   const maxPixels
     = currentInstance.value!.configuration?.mediaAttachments?.imageMatrixLimit
-      ?? 4096**2
+      ?? 4096 ** 2
 
   const loadImage = (inputFile: Blob) => new Promise<HTMLImageElement>((resolve, reject) => {
     const url = URL.createObjectURL(inputFile)
--- a/composables/masto/publish.ts
+++ b/composables/masto/publish.ts
@@ -122,7 +122,7 @@ export function useUploadMediaAttachment(draftRef: Ref<Draft>) {
 
   const maxPixels
     = currentInstance.value!.configuration?.mediaAttachments?.imageMatrixLimit
-      ?? 16777216 // 4096x4096
+      ?? 4096**2
 
   const loadImage = (inputFile: Blob) => new Promise<HTMLImageElement>((resolve, reject) => {
     const url = URL.createObjectURL(inputFile)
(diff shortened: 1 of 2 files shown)


zaidhaan:
Hmm, works fine in dev locally but it looks like the Content Service Policy is blocking image blobs in the preview.
image

Also the error message doesn't look too great, yikes. Error handling could be improved
image

Shouldn't take too long to fix so I won't turn this into a draft just yet.

zaidhaan:
Okay, all done.

CSP blob issue fixed. Resizing works.
image

As for the error message, per my last commit I just made it such that if resizing fails for whatever reason, then just attempt to upload the original file anyway. The user need not know about the details of a resizing failure, but it will still be logged in the console.

composables/masto/publish.ts
```diff
@@ -120,6 +120,62 @@ export function useUploadMediaAttachment(draftRef: Ref<Draft>) {
   let failedAttachments = $ref<MediaAttachmentUploadError[]>([])
   const dropZoneRef = ref<HTMLDivElement>()
 
+  const maxPixels
+    = currentInstance.value!.configuration?.mediaAttachments?.imageMatrixLimit
+      ?? 16777216 // 4096x4096
```
patak-dev, line 125: I think we can do the math, no need to precompute it.
```suggestion
      ?? 4096**2
```

composables/masto/publish.ts
```diff
@@ -120,6 +120,62 @@ export function useUploadMediaAttachment(draftRef: Ref<Draft>) {
   let failedAttachments = $ref<MediaAttachmentUploadError[]>([])
   const dropZoneRef = ref<HTMLDivElement>()
 
+  const maxPixels
+    = currentInstance.value!.configuration?.mediaAttachments?.imageMatrixLimit
+      ?? 4096**2
```
patak-dev, line 125: Ah, linting 🤦🏼‍♂️ 
```suggestion
      ?? 4096 ** 2
```

patak-dev (approved):
Awesome work @zaidhaan, let's merge it so we get some more testing on https://main.elk.zone

PR #1838, opened 2023-03-01
Underline footer links and move sponsor link down with them

ZMYaro:
[As discussed in the \#ui channel on the Elk Discord](https://discord.com/channels/1044887051155292200/1045711701359743098/1080230488880467998).


ZMYaro:
image

patak-dev:
I think it is too much in a single line, and we lose the Privacy 
image

I prefer we keep the sponsor heart button as it was before for now, or we could try two lines 🤔 
```
Mastodon  Discord  Github
About  Privacy  ❤️ Sponsor
```

PR #1842, opened 2023-03-02
fix: allow parsing codeblocks with multiple text nodes in children

### LINKED ISSUE #1840: Multiline code-block only rendering first line
A toot containing a multiline code-block is not rendered correctly.

What it looks like on the original mastodon instance:
Screenshot 2023-03-02 at 09 15 04

What it looks like in elk:
Screenshot 2023-03-02 at 09 15 15

For reference: [the toot in question](https://social.treehouse.systems/@alyssa/109933427007257424).

zaidhaan:
Currently, if we have a `<pre>` with a nested `<code>`, we assume that there's only one child within that code block. But in some cases, we might find some `<br>`s in there causing there to be more than just one child. So we don't want to only consume the first child and throw the rest away, rather we want to properly consume each child.

This could be done iteratively or recursively. I chose to go with the recursive solution since since we might possibly run into a few odd situations in which code blocks (for whatever reason) might have children that then have other nested children (not something to normally expect, but then again, who expects `<br>`s in `<code>`s like what happened just now?).

Before screenshot ([canary link](https://main.elk.zone/social.treehouse.systems/@alyssa/109933427007257424)):
image

Patched screenshot:
(Don't have a deploy link for this one cause it would redirect to a different URL for some reason, but you could pull this and test it out locally using [this link](http://localhost:5314/social.treehouse.systems/@alyssa/109933427007257424))
(Regarding the redirect; this is important, if on `.../social.treehouse.systems/@alyssa/...` the content parsed is the buggy content which this pr fixes, if it instead is the `.../mastodon.social/@alyssa@treehouse.systems/...`, then it's fed different content that's parsed without the pre tag which this pr does not affect since the bug does not take action there)
image

Fixes #1840.

### DIFF
--- a/composables/content-parse.ts
+++ b/composables/content-parse.ts
@@ -162,6 +162,13 @@ export function htmlToText(html: string) {
   }
 }
 
+export function recursiveTreeToText(input: Node): string {
+  if (input && input.children && input.children.length > 0)
+    return input.children.map((n: Node) => recursiveTreeToText(n)).join('')
+  else
+    return treeToText(input)
+}
+
 export function treeToText(input: Node): string {
   let pre = ''
   let body = ''
--- a/composables/content-render.ts
+++ b/composables/content-render.ts
@@ -107,7 +107,7 @@ function handleCodeBlock(el: Node) {
     const codeEl = el.children[0] as Node
     const classes = codeEl.attributes.class as string
     const lang = classes?.split(/\s/g).find(i => i.startsWith('language-'))?.replace('language-', '')
-    const code = recursiveTreeToText(codeEl)
+    const code = codeEl.children && codeEl.children.length > 0 ? recursiveTreeToText(codeEl) : ''
     return h(ContentCode, { lang, code: encodeURIComponent(code) })
   }
 }
--- a/composables/content-render.ts
+++ b/composables/content-render.ts
@@ -107,7 +107,7 @@ function handleCodeBlock(el: Node) {
     const codeEl = el.children[0] as Node
     const classes = codeEl.attributes.class as string
     const lang = classes?.split(/\s/g).find(i => i.startsWith('language-'))?.replace('language-', '')
-    const code = codeEl.children[0] ? treeToText(codeEl.children[0]) : ''
+    const code = recursiveTreeToText(codeEl)
     return h(ContentCode, { lang, code: encodeURIComponent(code) })
   }
 }


zaidhaan:
Whoops, tests failed. Should've done that beforehand. Let me quickly look into that

patak-dev:
Great find! These kinds of bugs are really bad for the experience, as users start to mistrust Elk. Do you have any insights into why things work depending on looking at the direct link or the link through the user server. This is really strange https://github.com/elk-zone/elk/issues/1840#issuecomment-1451569547

patak-dev (approved):

zaidhaan:
> Great find! These kinds of bugs are really bad for the experience, as users start to mistrust Elk. Do you have any insights into why things work depending on looking at the direct link or the link through the user server? This is really strange [#1840 (comment)](https://github.com/elk-zone/elk/issues/1840#issuecomment-1451569547)

Sure thing. That has to do with the `status.content` retrieved for a toot. I'm not too familiar with the internals of mastodon, but anyway, if you look at the content retrieved from the **original instance (formatting included)** (`.../social.treehouse.systems/@alyssa/... `), you get the following:
```
<p>i just typed</p><pre><code>$ gup<br>$ up<br>$ frb<br>$ pub<br>$ lab 21426 m<br></code></pre><p>now thinking i might have an absurd number of shell aliases</p>
```
Notice the existence of `<pre>` and `<code>` above. The parsing for this was what this PR addressed.

If instead, you looked at the page through `.../mastodon.social/@alyssa@treehouse.systems/...`, you would get the following:
```
<p>i just typed</p><p>$ gup<br>$ up<br>$ frb<br>$ pub<br>$ lab 21426 m<br></p><p>now thinking i might have an absurd number of shell aliases</p>
```
Notice the lack of `<code>` and `<pre>`. This is why it renders fine, albeit with a lack of formatting.

Again, not too familiar with mastodon but I think there's no markdown support for mastodon so if you retrieve the content of a toot through mastodon, perhaps it strips away any unsupported formatting?

I'd be curious to know why the [elk.zone link](https://elk.zone/social.treehouse.systems/@alyssa/109933427007257424) to that particular toot redirects to the `.../mastodon.social/...` path, because this does strip away the formatting. I don't think that's something to be desired.

patak-dev:
> Again, not too familiar with mastodon but I think there's no markdown support for mastodon so if you retrieve the content of a toot through mastodon, perhaps it strips away any unsupported formatting?

Yes, I was expecting the HTML to be the same in both though. But I think the servers are stripping some tags depending if you are in Mastodon or not (a fork that supports Markdown generated HTML).
 
> I'd be curious to know why the [elk.zone link](https://elk.zone/social.treehouse.systems/@alyssa/109933427007257424) to that particular toot redirects to the `.../mastodon.social/...` path, because this does strip away the formatting. I don't think that's something to be desired.

If you are checking this logged into a mastodon.social account, then the redirects happen because when you are logged in to your server, you always access the network through your own server and not doing requests to the servers of other folks (good for privacy for example)

PR #1843, opened 2023-03-02
fix: only use username part of mention

### LINKED ISSUE #1803: post length incorrectly includes domain-part of mentions
Mastodon docs (emphasis mine): https://docs.joinmastodon.org/user/posting/#mentions
> You can mention users by typing out their full address, e.g. @alice@example.com. Note that any usage of@word will be interpreted as mentioning the local user with the username word, if that user exists. *Only the username part will count against your character limit – the domain is not counted.*

Elk counts the domain-part of a mention against the total post length, resulting in it showing over 500 characters when, in fact, the server will accept the post.

**Note**: it looks like the calculation is done correctly while drafting a reply, for the users being replied to, but if you then go to edit the reply, it counts the domain-part again so the post appears longer than it was when originally written.

In the native interface, the length counter does not increase while you are typing the domain-part of a mention:
Screen Shot 2023-02-20 at 12 25 36 AM

In Elk, it does:
Screen Shot 2023-02-20 at 12 27 30 AM

zaidhaan:
As mentioned from the [docs link](https://docs.joinmastodon.org/user/posting/#mentions) in the [referenced issue](https://github.com/elk-zone/elk/issues/1803)
> You can mention users by typing out their full address, e.g. @[alice@example.com](mailto:alice@example.com). Note that any usage of@word will be interpreted as mentioning the local user with the username word, if that user exists. Only the username part will count against your character limit – the domain is not counted.

I've taken a look at how mastodon themselves deal with this and used their regex (with attribution in a comment) since I assume theirs would handle any weird edge cases.

Mastodon:
image

Elk canary (incorrect, 6 characters expected):
image

This patch:
image

This should be fine. I tested it with a bunch of random text and it matches up with lengths reported by mastodon.

Fixes #1803.

### DIFF
--- a/components/publish/PublishWidget.vue
+++ b/components/publish/PublishWidget.vue
@@ -71,12 +71,18 @@ const characterCount = $computed(() => {
   // taken from https://github.com/mastodon/mastodon/blob/07f8b4d1b19f734d04e69daeb4c3421ef9767aac/app/lib/text_formatter.rb
   const linkRegex = /(https?:\/\/(www\.)?|xmpp:)\S+/g
 
+  // taken from https://github.com/mastodon/mastodon/blob/af578e/app/javascript/mastodon/features/compose/util/counter.js
+  const countableMentionRegex = /(^|[^/\w])@(([a-z0-9_]+)@[a-z0-9.-]+[a-z0-9]+)/ig
+
   // maximum of 23 chars per link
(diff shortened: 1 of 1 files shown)


patak-dev (approved):
Brilliant, thanks @zaidhaan!

PR #1854, opened 2023-03-05
fix: disallow media description exceeding limits

### LINKED ISSUE #1018: alt text is silently dropped if it overruns the server limit
### Pre-Checks

### Describe the bug

Using Elk, if alt text in a toot exceeds your instance's limit, the alt text is silently deleted. The toot is published, just with the alt text missing.

My server's alt text limit is 1500 chars. I pasted in 4200 characters of alt text: 
Screenshot-2023-01-12,13-11-37

After hitting apply, you can see that the alt text badge shows up, and the text is there:
Screenshot-2023-01-12,13-12-07

But once I hit the publish button, the final toot has no alt text:
Screenshot-2023-01-12,13-12-39

The alt text is just silently dropped when publishing.

### Additional Context

Mastodon has no API (afaik) for determining the maximum amount of characters for alt text. My instance (social.ridetrans.it) has a limit of 1500 chars. Pinafore (an alternative frontend) arbitrarily enforces a 420 character limit, which is too small in my opinion. Elk enforces no limit at all.

I don't know if the solution here is an arbitrary limit, or to detect an error when publishing and returning the user to the public screen (alt text intact) so they can fix the issue before attempting to once again hit that publish button.

zaidhaan:
Mastodon enforces a 1500 character limit on media attachment descriptions. This is visible in their interface too:
image

Elk on the other hand has no such indicator:
image

As such, it enforces no limits on the attachment length, and so if you post something exceeding 1500 characters, you get this in your console:
image

This error is not visible to the user (due to it not being captured in `failedMessages` since the error occurs in `setDescription` (`publish.ts`) which has no access to that array)

In any case, the potential solutions to this issue would be as follows:
- Let the user enter a large description, but display the error once they apply the description
- Indicate to the user how many characters have been entered and how many are left, and prevent setting descriptions that exceed that value

I chose to go with the latter solution, since the first one was a bit difficult to implement given the current architecture of the code.

So with this PR, one would see this after entering too many characters:
image

**Additionally**, since this PR prevents setting long descriptions (i.e. disables the button), I figured I'd also include that inside the status submission ("Publish") button, since that effectively prevents the user from calling an unnecessary API request that will always fail, and will always throw an error message to the user. So long statuses won't be publishable:
image

(Perhaps not the most elegant solution -- disabled button does not look the greatest, and I had to make a simple character component just to avoid code duplication -- but it works)

Fixes #1018.

### DIFF
--- a/components/publish/PublishAttachment.vue
+++ b/components/publish/PublishAttachment.vue
@@ -15,6 +15,9 @@ const emit = defineEmits<{
   (evt: 'setDescription', description: string): void
 }>()
 
+// from https://github.com/mastodon/mastodon/blob/dfa984/app/models/media_attachment.rb#L40
+const maxDescriptionLength = 1500
+
 const isEditDialogOpen = ref(false)
 const description = ref(props.attachment.description ?? '')
 const toggleApply = () => {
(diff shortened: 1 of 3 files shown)


patak-dev:
Awesome! We can later check the styling of the button, we are due a general review of all buttons in the app.
Let's merge it, this works great 👍🏼

patak-dev (approved):

PR #1859, opened 2023-03-05
fix: render replies to fill the screen height

### LINKED ISSUE #1849: Usability issue when displaying replies to posts
Just realised that my problem with the web client (overall it's great - such elegant design) is the lack of affordance when replies to posts are displayed. 

I _thought_ that only 3 replies were ever available because there were no visible hints or clues to tell me that other replies
elk
 would be displayed if I scrolled down. All I saw was white space.

I only discovered this functionality by accident. A classic usability issue. Such important functionality needs to be explicit, not hidden.

Easy to fix by providing some affordance - show your users what they can do.

zaidhaan:
When viewing a post, by default not all replies are rendered ([Mark Ruffalo post](https://main.elk.zone/mastodon.social/@MarkRuffalo/109638532333776039)):
image

This PR fixes this and makes it such that replies would always fill up the document's `clientHeight`
image

For more information regarding the details of this, take a look at the linked issue below.

FYI, Since the original PR that introduced the `<DynamicScroller>` was for performance reasons, I should note that this PR does not affect performance as the Mark Ruffallo post rendered perfectly fine with this fix.

Fixes #1849.

### DIFF
--- a/pages/[[server]]/@[account]/[status].vue
+++ b/pages/[[server]]/@[account]/[status].vue
@@ -42,6 +42,12 @@ async function scrollTo() {
   statusElement.scrollIntoView(true)
 }
 
+const clientHeight = ref(0)
+
+onMounted(() => {
+  clientHeight.value = document.documentElement.clientHeight
+})
+
 const publishWidget = ref()
 const focusEditor = () => publishWidget.value?.focusEditor?.()
 
@@ -96,6 +102,7 @@ onReactivated(() => {
               v-slot="{ item, index, active }"
               :items="context?.descendants || []"
               :min-item-size="200"
+              :buffer="clientHeight"
               key-field="id"
               page-mode
             >


pages/[[server]]/@[account]/[status].vue
```diff
@@ -96,6 +102,7 @@ onReactivated(() => {
               v-slot="{ item, index, active }"
               :items="context?.descendants || []"
               :min-item-size="200"
+              :buffer="clientHeight"
```
Comment (patak-dev) [inline pages/[[server]]/@[account]/[status].vue:105]: Great stuff, this was affecting the UX a lot.

About using `clientHeight`, from [the docs](https://github.com/Akryum/vue-virtual-scroller/blob/master/packages/vue-virtual-scroller/README.md#props):
- buffer (default: 200): amount of pixel to add to edges of the scrolling visible area to start rendering items further away.

I think we should instead use the maximum post height. At this point, this is hard to know because we haven't yet implemented truncation of long posts but once that is in, we should be able to have a reasonable estimate. The issue I see with using the `clientHeight` is that it may be too big in some cases, or too small in others (or change after the component was setup and we aren't changing it). 

So maybe using a fixed value like 800 is better here?

patak-dev (approved):
Let's go with this PR for the moment, as this fixes an important UX issue with Elk. @Akryum if you have time at some point, maybe you could check our usage of DynamicScroller. I agree with @zaidhaan here: https://github.com/elk-zone/elk/pull/1859#discussion_r1126748081, we may be misunderstanding the API because it should already be working without having to use 800px as a guard.

Issue #1422, opened 2023-01-24
Accessibility: Unlabeled links


shortwavesurfer2009:
When using a screen reader the statusbar links just read as "link", "link", "link", "link" and the last link will not focus at all.
![Screenshot_20230124-102351_Fennec](https://user-images.githubusercontent.com/116814522/214335974-aa76d53c-58bb-4d56-bc77-55adcb1a68e3.png)
![Screenshot_20230124-102359_Fennec](https://user-images.githubusercontent.com/116814522/214335980-8d79828d-cf6e-43a2-a886-991220f876eb.png)
![Screenshot_20230124-102403_Fennec](https://user-images.githubusercontent.com/116814522/214335986-660683e7-badb-4bc9-8351-18e89deb16c5.png)
![Screenshot_20230124-102411_Fennec](https://user-images.githubusercontent.com/116814522/214335987-0e41b8d7-2700-4da9-b2c2-2394e88479e1.png)

PR #1867, opened 2023-03-08
fix(a11y): add aria labels to bottom navigation links

zaidhaan:
_Partially_ addresses #1422.

Only thing left would be making the last item (more menu) focusable (and presumably toggle-able using Space/Enter?). I'm not too experienced with a11y but I've observed the following when making changes to the `<label>` in the more menu:
* adding `role="button"` allows using Space/Enter to toggle on Firefox 110.0. But does not make it focusable
* adding `tabindex="0"` makes it focusable / tab accessible but hitting Space/Enter no longer triggers a toggle

Maybe someone could suggest/co-author the proper fix for the more menu so that the linked issue could be closed, or we could just merge this now and address that later.

### DIFF
--- a/components/nav/NavBottom.vue
+++ b/components/nav/NavBottom.vue
@@ -13,13 +13,13 @@ const { notifications } = useNotifications()
   >
     <!-- These weird styles above are used for scroll locking, don't change it unless you know exactly what you're doing. -->
     <template v-if="currentUser">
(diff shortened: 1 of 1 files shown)


patak-dev (approved):

patak-dev:
@zaidhaan would you send this question also to the #a11y channel in https://chat.elk.zone?

Issue #1869, opened 2023-03-08
Local custom emoji become :undefined: on posting


kevinmarks:
Making a separate issue for this from #1551 where it is [mentioned in passing](https://github.com/elk-zone/elk/issues/1551#issuecomment-1420018963) 
If I make a post with a local custom emoji, eg :ohnobubble: it appears inline, then turns into :undefined: on [posting](https://xoxo.zone/@KevinMarks/109988582055689738).
<img width="616" alt="Screenshot 2023-03-08 at 16 22 35+0000" src="https://user-images.githubusercontent.com/99008/223770400-b7d04e55-0691-4c06-9f54-dca774552091.png">
<img width="613" alt="Screenshot 2023-03-08 at 16 22 51+0000" src="https://user-images.githubusercontent.com/99008/223770392-5e25a4b9-5b1d-416f-ac1b-159990927264.png">
<img width="630" alt="Screenshot 2023-03-08 at 16 23 01+0000" src="https://user-images.githubusercontent.com/99008/223770382-8538853b-3f8f-41f7-8f66-fb9ee23a51da.png">

kevinmarks:
Not sure - I have seen this behaviour previously, but not sure which build it appeared in.

cyberalien:
> Closed #1551, thanks for creating this issue @kevinmarks. @cyberalien @userquin do you know if this ever worked after we started using the new machinery in iconify? I think this may not be a regression but a bug introduced back then.

Iconify stuff doesn't affect this. This is not for unique emojis, this is for custom "emojis" set in context.

cyberalien:
Got to reproduce this issue. Definitely an Elk bug, text starts with `:undefined:` when converted from editor to status.

Did some digging, issue happens because custom emojis are missing `data-emoji-id` attribute, so `htmlToText` cannot  convert HTML from editor to text correctly. Not sure what caused it.

PR #1871, opened 2023-03-08
fix: get custom emoji id from title attribute

cyberalien:
Fixes https://github.com/elk-zone/elk/issues/1869


patak-dev:
Thanks for the quick fix @cyberalien! It now works well when typing the custom emoji and publishing. It still doesn't work when you select and copy the custom emoji, then paste and publish (this works for regular emoji). It may be a related issue. Let's merge this one to fix the reported problem first.

patak-dev (approved):

PR #1873, opened 2023-03-08
fix: copy and paste custom-emoji

patak-dev:
Reference:
- #1871

@cyberalien it looks like when we copy and paste, there is no `title` in the attributes and we are instead adding a `alt` that has the full `:name:`

This PR fixes copy and paste, but I don't know if this is correct long term. We also are creating [here](https://github.com/elk-zone/elk/blob/fa3cfd6059f2bdd42bc5d349ac225c29c6f666de/composables/content-parse.ts#L429) a custom emoji without a title but this one has a `emoji-data-id`

Looks like the `emoji-data-id` is lost on copy and paste.
We could still merge this one as a patch.

### DIFF
--- a/composables/content-parse.ts
+++ b/composables/content-parse.ts
@@ -218,8 +218,10 @@ export function treeToText(input: Node): string {
     body = (input.children as Node[]).map(n => treeToText(n)).join('')
 
   if (input.name === 'img' || input.name === 'picture') {
-    if (input.attributes.class?.includes('custom-emoji'))
-      return `:${input.attributes['data-emoji-id'] ?? input.attributes.title ?? input.attributes.alt?.slice(1, -1)}:`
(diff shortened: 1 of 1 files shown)


cyberalien:
With all this weird behaviour going on, maybe it would be better not to assume that value starts/ends with `:`?

Custom emojis can't have double `::`, so it would be safe to just wrap whole chain of attributes, get string, remove first/last entry with regex, which would work the same regardless of which attribute it came from.

Something like this:
```ts
return `:${(input.attributes['data-emoji-id'] ?? input.attributes.title ?? input.attributes.alt ?? '').replace(/^:/, '').replace(/:$/,'')}:`
```

patak-dev:
Sounds good! I'll change it and merge tomorrow when I'm back

PR #1874, opened 2023-03-08
Update README.md

MIfoodie:
You can now download the client! (https://imgur.com/a/cNxpMuU)


patak-dev:
Thanks for the PR @MIfoodie! I think we should keep the tagline as is though. It is still a Web client even if you install it as a PWA.

PR #1876, opened 2023-03-09
feat: markdown editor, close #1845

### LINKED ISSUE #1845: Add non-WYSIWYG editor
The current WYSIWYG editor has multiple issues such as #917 or #401.  While it is hard to build a WYSIWYG editor that automatically converts e.g., `*foo*` into emphasized text on input and then properly deals with the user editing the emphasized text or wanting it un-emphasized, it's much easier to write a "plain" editor input field that simply allows the user to input Markdown which is then rendered only on submission (or by using an explicit preview).

A toggle in the settings could be provided to select b/w WYSIWYG and non-WYSIWYG editing modes.

antfu:


patak-dev:
This is great! If we could add syntax highlighting to the markdown editor at one point, I may use it by default.

Some comments:
- Maybe we could avoid the "Markdown Editor" label because the icon button changing below could already be enough indication (this avoids the layout shift)
- There is a bug with code blocks. It works well from the rich editor to markdown, but if you go the other way around, the three backticks are converted to a single one.
- I wonder if we should support emojis using the `:heart:` notation in markdown. Currently, if you type that and swap to the Rich editor it won't be replaced.

zaidhaan:
> There is a bug with code blocks. It works well from the rich editor to markdown, but if you go the other way around, the three backticks are converted to a single one.

Just quickly glancing at the parsing code, the lack of `>` preceding the triple backticks fails the regex here, so removing it fixes it but I don't know if there's a reason for the `>` to begin with. 
<details>
<summary> View diff </summary>

```diff
diff --git a/composables/content-parse.ts b/composables/content-parse.ts
index 6e2d3383..f96e43ff 100644
--- a/composables/content-parse.ts
+++ b/composables/content-parse.ts
@@ -91,13 +91,13 @@ export function parseMastodonHTML(
   if (markdown) {
     // Handle code blocks
     html = html
-      .replace(/>(```|~~~)(\w*)([\s\S]+?)\1/g, (_1, _2, lang: string, raw: string) => {
+      .replace(/(```|~~~)(\w*)([\s\S]+?)\1/g, (_1, _2, lang: string, raw: string) => {
         const code = htmlToText(raw)
           .replace(/</g, '&lt;')
           .replace(/>/g, '&gt;')
           .replace(/`/g, '&#96;')
         const classes = lang ? ` class="language-${lang}"` : ''
-        return `><pre><code${classes}>${code}</code></pre>`
+        return `<pre><code${classes}>${code}</code></pre>`
       })
       .replace(/`([^`\n]*)`/g, (_1, raw) => {
         return raw ? `<code>${htmlToText(raw).replace(/</g, '&lt;').replace(/>/g, '&gt;')}</code>` : ''

```
</details>

antfu:
This is quite hard... It's because we have:

1. Plain markdown
2. Mastodon HTML-encoded markdown
3. Expanded HTML for TipTap

I have the logic of `2->3`, `2->1`, and `3->1`. But we are missing `1->2` since it's done on the mastodon server. That means we can't easily to `1->3` on the client side (the bug is caused that I am using `2->3` for `1->3`)

antfu:
To solve I think we need to introduce some real markdown to the HTML library and try to be close to how the Mastodon server is doing.

Issue #1890, opened 2023-03-13
Improve logo spacing/decrease width of elk


EddiesTech:
In my opinion, the logo is too wide. It looks weird next to other icons, especially as a PWA on Windows in the taskbar. The padding between the border of the SVG viewbox and the actual Elk icon. I have made a quick site that shows an idea of what I think it would be like:
https://eddiestech.github.io/elklogocompare (left is original, right is my mock-up)

Happy to get feedback and also look into guidance around spacing like this.

EddiesTech:
https://github.com/elk-zone/elk/pull/1896
I may have missed some logo files if they are hidden away somewhere. But I did all the files in `/public` and the SVG code in the vue template.

PR #1896, opened 2023-03-14
style: Change logo files with added left/right padding

EddiesTech:
Closes https://github.com/elk-zone/elk/issues/1890


EddiesTech:
For reference, here is the difference between the two logo svgs in Inkscape. This may not be perfect, and I may have missed files or made mistakes. Thanks for the great client!
bad left right padding
now in guides

EddiesTech:
Sorry for missing these folders. Will look into doing them tonight.
image

patak-dev:
Thanks for caring about these details ❤️ 

At least for me, when I see the comparison you linked
image

I have the feeling that it looks a little bit too compressed after the edit. But the one you linked here looks better. Did you change something? Or is it just that here I don't see them side by side. 
image

If they are the same, I wonder if we should go for something in the middle of the two.

EddiesTech:
@patak-dev Yes, the original comparison was between the original and a mock-up that I guessed from what I thought it should look like. The new version with the blue borders is a new version making the left/right padding ~ equal to the top/bottom padding of the logo, by measuring on Inkscape. I agree, the original mock-up was too squished in, but I feel the version with the blue borders looks equal and fits better. As I say, I will try changing the dev and staging files, so you can see the change in the Netlify demo.

patak-dev:
Yes, I agree 👍🏼 

I like the new version you did 🧡

EddiesTech:
@patak-dev Updated more files... let me know what you think :)

ayoayco (commented):
I was wondering about the full-body logo. Can we adjust those too? I imagine the head will not fit the body like before? 😄 See: https://docs.elk.zone/docs/

EddiesTech:
> The idea was to only use the head as branding moving forward. I think we should replace the one from the docs.

I can do this

> * they are now solid color instead of transparent. We talked with @userquin a while ago and I think this isn't an issue (what is more, it is better for them to be solid)

The pwa-* and favicon.ico files are transparent, as they are on the current main branch. The apple-touch-icon.png and maskable-icon.png images have a white background, again, as on the current main branch. I can make them all white if you'd like. In my opinion, the pwa-* images should be white background with small rounded corners, to fit in with apps on Windows, Mac, etc. (mobile will crop past the rounded corners anyway). I think the apple-touch-icon.png and maskable-icon.png should stay white background as they are - as the idea is that the devices/sites/etc. will do the cropping/masking.

> * but the new ones look blurry in the smallest versions (the old ones look a lot crispier). @userquin I think you generated them, what tool did you use?

Happy to use a different tool to render these. I can probably do it straight from Inkscape or Paint.net, but I'll wait until we have an idea of how we want to proceed with things like backgrounds, rounded corners, etc.

EddiesTech:
How about these rounded corners? @patak-dev @ayoayco 
logo rounded

ayoayco:
If anyone is viewing on light background, the rounded corners can only be seen on dark mode

Screenshot 2023-03-15 at 10 15 02 AM
Screenshot 2023-03-15 at 10 13 34 AM

EddiesTech:
Hi, again @patak-dev @ayoayco. Just made a new commit with better exports straight from Inkscape, rounded white background PWA icons, and I changed the logo in the docs. Let me know what you think :)

ayoayco:
Should the apple-touch-icon.png files be updated to have rounded corners too? Are those for iOS PWA? @userquin

EddiesTech:
Yes, they are for iOS PWA. But, I purposely didn't change it, as iOS takes the file and adds its own rounding (from my understanding. e.g. on your iPhone Home screen), so there is no need for us to round the corners.
[mathiasbynens.be/notes/touch-icons](https://mathiasbynens.be/notes/touch-icons#effects),

ayoayco:
Okay, cool! Looks good to me :)

ayoayco (approved):

userquin:
@EddiesTech can you optimize png files using for example https://compresspng.com/ or some other tool?

image

Tested PR preview on Android, new icons looks good to me, we need to test it on macOS and iOS.

userquin:
@cyberalien (sorry DM) is testing icons on macOS/iOS using the emulator.

userquin:
> Should the apple-touch-icon.png files be updated to have rounded corners too? Are those for iOS PWA? @userquin

I have no idea, latest changes (I don't use iOS) on the OS seems to change the behavior, but on beta 2 seems to work:
- https://twitter.com/userquin/status/1628361822695833601

cyberalien:
Looks good in iOS

Simulator Screen Shot - iPhone 14 Pro Max - 2023-03-18 at 14 57 45

cyberalien (approved):

cyberalien:
Actually... it does look way better on Android. Probably could use more padding for iOS.

userquin:
we should review the padding: I have no idea what icon is using macOS/iOS, here Android icon:

image

userquin:
I guess we can keep current ones (once optimized), if we reduce the padding/margin (safe are: https://web.dev/maskable-icon/, https://maskable.app/) we can break Android ones (too small).

I use https://realfavicongenerator.net/ to create `apple-touch-icon.png` to add padding/margin to the icon, but it seems this icon is not used anymore on macOS/iOS (using the slider on iOS section):

image

EddiesTech:
Hi @userquin and @cyberalien. Thanks for your help testing it out and giving feedback. I agree - the iOS version (which I think is using apple-touch-icon.png, as that has no padding) does not have enough padding, and I think the Android (which I think is using maskable-icon.png which does have padding) has too much padding (should be in line with Edge and Firefox, etc. in the screenshot above). Funnily enough, the maskable icon was generated with maskable.app with exactly the same padding as the Elk original. But I can change the padding and we can see what it looks like. And I will try adding padding to apple-touch-icon.png. And, yes, I will compress the images. But, I am not sure about the SVG. @userquin Do you have a tool for optimising them? Technically Inkscape has 'Optimised' and 'Compressed' options, but I am not sure how good they are
image

userquin:
@EddiesTech there is no need to optimize svg (we only reduce about the 2%), I use https://jakearchibald.github.io/svgomg/.

EddiesTech:
OK, I have compressed all the images using the website that @userquin sent. Note that it is lossy compression, so some quality will be lost. I have also changed the padding on the apple-touch-icon.png so it will hopefully look better on iOS devices. Would somebody be able to test that out? I have left Android - maskable-icon - alone for now, as according to the maskable.app website it is just within the safety zone, so it doesn't make sense to take it out of the safety zone. Any ideas?

cyberalien:
Screenshot from iOS:

Simulator Screen Shot - iPhone 14 Pro Max - 2023-03-21 at 08 56 10

EddiesTech:
Looks alright on iOS now to me. What do we think about Android? I still think it has a bit too much padding, but I don't want it to go out the safety zone for other uses of the maskable icon. @userquin

EddiesTech:
Mind you, compared with the new iOS, they look about the same

userquin (approved):

patak-dev:
Awesome work, thanks to all the reviewers and to @EddiesTech for implementing all the needed changes 🧡

PR #1897, opened 2023-03-16
feat: Include following count in hide follower count setting

### LINKED ISSUE #1878: Add "hide following count" to Wellbeing options
The "wellbeing" options in settings allows users to choose "hide follower count", but not "hide following count".

I think either it should be a new option, or the existing option should cover both cases, not least because I always forget which one is which and end up having to experimentally set options to see what they've done.

kkoyung:
Before changes: Enabling "Hide follower count" in the wellbeing setting can hide the follower count in the account profile page.

I changed that setting to "Hide following/follower count" which can hide both following count and follower count.

Close #1878

### DIFF
--- a/pages/settings/preferences/index.vue
+++ b/pages/settings/preferences/index.vue
@@ -78,10 +78,10 @@ const userSettings = useUserSettings()
       {{ $t('settings.preferences.hide_reply_count') }}
     </SettingsToggleItem>
     <SettingsToggleItem
-      :checked="getPreferences(userSettings, 'hideFollowerCount')"
-      @click="togglePreferences('hideFollowerCount')"
+      :checked="getPreferences(userSettings, 'hideFollowingFollowerCount')"
+      @click="togglePreferences('hideFollowingFollowerCount')"
     >
-      {{ $t('settings.preferences.hide_follower_count') }}
+      {{ $t('settings.preferences.hide_following_follower_count') }}
     </SettingsToggleItem>
     <SettingsToggleItem
       :checked="getPreferences(userSettings, 'hideUsernameEmojis')"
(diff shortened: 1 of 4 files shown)


pages/settings/preferences/index.vue
```diff
@@ -78,10 +78,10 @@ const userSettings = useUserSettings()
       {{ $t('settings.preferences.hide_reply_count') }}
     </SettingsToggleItem>
     <SettingsToggleItem
-      :checked="getPreferences(userSettings, 'hideFollowerCount')"
-      @click="togglePreferences('hideFollowerCount')"
+      :checked="getPreferences(userSettings, 'hideFollowingFollowerCount')"
+      @click="togglePreferences('hideFollowingFollowerCount')"
```
patak-dev, line 82: Thanks! I think we should leave the name of the preference as it was before though because if not, everyone will lose this preference and will need to re-enable it. We could add a backward-compatible scheme, but I think the previous name is good enough here.

kkoyung:
@patak-dev I reversed the name of preference to `hideFollowerCount` and the name of the displayed text to `hide_follower_count`. I kept the text displayed at the setting page as "Hide following/follower count" in order to precisely describe the setting.

Since the name of the displayed text remains unchanged and the non-English displayed texts have not yet updated in this PR. User who uses non-English interface may not notice the changes. Remember to update the language file for other language later.

patak-dev (approved):
Thanks for the changes! Let's merge this.

But I wonder if we should also only hide the followers/following counts in hover cards and then on profile still show "Followers" / "Following" without numbers but at least let the user click on them so they can explore. Checking who a person follows is a good way to learn about new good follows, and I think it doesn't go against the health preference.

kkoyung:
@patak-dev 
I prefer hiding the counts in both hover cards and full profile page, since we don't have hover cards on mobile browsers. If we only hide the counts on hover cards, it means mobile user will lose this hiding count function.

But I agree to keep showing "Followers" / "Following" label without numbers (in both hover cards and full profile page) so that user can click on them to see the following list and the follower list.

PR #1901, opened 2023-03-17
feat: allow user to view follow list in "Hide following/follower count" mode

kkoyung:
From @patak-dev 's idea (see [here](https://github.com/elk-zone/elk/pull/1897#pullrequestreview-1344639293)), I adjusted the wellbeing "Hide following/follower count" feature.

Before changes, when "Hide following/follower count" is enabled, the following count and the follower count are hidden as shown below.
screenshot_2023-03-17_23:13:49

After changes, when "Hide following/follower count" is enabled, only the numbers are hidden. The text "Following" and "Followers" are kept so that users are still able to click them to view the following list and the follower list.
screenshot_2023-03-17_23:15:03

### DIFF
--- a/components/account/AccountHoverCard.vue
+++ b/components/account/AccountHoverCard.vue
@@ -19,6 +19,6 @@ const relationship = $(useRelationship(account))
     <div v-if="account.note" max-h-100 overflow-y-auto>
       <ContentRich text-4 text-secondary :content="account.note" :emojis="account.emojis" />
     </div>
-    <AccountPostsFollowers text-sm :account="account" />
+    <AccountPostsFollowers text-sm :account="account" :is-hover-card="true" />
   </div>
 </template>
--- a/components/account/AccountPostsFollowers.vue
+++ b/components/account/AccountPostsFollowers.vue
@@ -26,7 +26,6 @@ const userSettings = useUserSettings()
       </template>
     </NuxtLink>
     <NuxtLink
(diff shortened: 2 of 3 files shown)


patak-dev:
Nice! Thanks for the PR 🙌🏼 

Two things I see:
- We could avoid showing Following/Followers when you are out of the profile (so it doesn't show when hovering on an avatar for example)
- I think it is better to have new keys for Following/Followers without a number or force it to use the plural form to avoid seeing `Follower` if there is only 1.

kkoyung:
I made changes on them.

- In the `AccountPostFollowers` component, I added a props `isHoverCard` which is set to be true by `AccountHoverCard` so that "Following" and "Followers" will be further hidden in hover card.

- In the `LocalizedNumber` component, force the plural form when the count is hidden, by `:plural="hideNumber ? 2 : count"`.

antfu:
I would also suggest we have separate keys for non-numbered names. The plural workaround works for English but is not necessary for other languages.

kkoyung:
I changed to re-use `account.following` and `account.followers` when counts are hidden. These two keys are used as html title after clicking "Following"/"Followers". I think it makes sense that the button and the title share same text.

Besides that, I found that when I click "XX Posts", "XX Following" and "XX Followers" back and forth, the html title only changes once for each one. I will create another issue about that later.

patak-dev (approved):
Sorry, it took us some time to merge this one. Looks great to me now! Thanks for implementing the feature.

PR #1921, opened 2023-03-23
feat: add account pinned tab

### LINKED ISSUE #305: missing info boxes in profile and pinned posts
Checking my own profile, it seems like a few things are inconsistent between elk and a Mastodon web-client.

Namely, I found the following:

- [ ] Some info boxes (Actually all except one) are missing.
- [ ] Pinned posts are not displayed.

Profile on Elk:
grafik

Profile on Mastodon:
grafik

lazzzis:
Try to resolve #305 

Add a new tab to show pinned posts.

Screenshot 2023-03-22 at 11 00 18 PM

### DIFF
--- a/components/account/AccountTabs.vue
+++ b/components/account/AccountTabs.vue
@@ -27,6 +27,15 @@ const tabs = $computed<CommonRouteTabOption[]>(() => [
     icon: 'i-ri:chat-1-line',
   },
   {
+    name: 'account-pinned',
+    to: {
+      name: 'account-pinned',
+      params: { server, account },
+    },
+    display: t('account.pinned'),
+    icon: 'i-ri:pushpin-2-line',
+  },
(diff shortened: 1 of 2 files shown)


sxzz:
Actually, I don't think it is a good design to add a new tab, because we already had pinned page.
image


Maybe we can add pinned statuses to the top of `Posts` tab like Mastodon official.

lazzzis:
> because we already had pinned page

oh. I see. I was not aware of this feature. It seems it only allows me to see my own pinned posts. I am still unable to see pinned posts from other accounts. Is it possible to open pinned page of another user?

> Maybe we can add pinned statuses to the top of Posts tab like Mastodon official.

Got it.

sxzz:
Seems no, we cannot open pinned page of other users.

PR #1937, opened 2023-04-01
fix: saving profile data breaks cache

### LINKED ISSUE #1879: Profile editor messes up User Picker
Saving profile in profile editor duplicates curently active user to other items in User Picker. 
[Video](https://imgur.com/a/UVJmAgV.mp4)

How to reproduce:
1. Select the 2nd account on User Picker
2. Go to own profile
3. `Edit profile`
4. Update profile then `Save`

Fixes itself by reloading the page. 
Also reproducible even if Settings>Prefferences>Experimental Features>User Picker is unticked. Difference is doesn't fix itself by reloading page. 
Not reproduceable when active user is the first item in User Picker.

Pardon if I'm doing something wrong, this is my first time submitting an Issue

userquin:
We need to add `@server` to the acct account response and override account cache, check:
- https://github.com/elk-zone/elk/blob/main/composables/users.ts#L147
- https://github.com/elk-zone/elk/blob/main/composables/users.ts#L172

closes #1879

### DIFF
--- a/pages/settings/profile/appearance.vue
+++ b/pages/settings/profile/appearance.vue
@@ -71,6 +71,12 @@ const { submit, submitting } = submitter(async ({ dirtyFields }) => {
     return
   }
 
+  const server = currentUser.value!.server
+
+  if (!res.account.acct.includes('@'))
+    res.account.acct = `${res.account.acct}@${server}`
+
+  cacheAccount(res.account, server, true)
   currentUser.value!.account = res.account
   reset()
 })


patak-dev (approved):
This makes sense with the current way we are handling user accounts, but for future reference @patak@m.webtoo.ls isn't valid. It should be @patak@webtoo.ls. We have this issue across the app right now and we have discussed with @userquin before about possible solutions. It isn't an easy task to fix it though.

PR #1938, opened 2023-04-01
refactor: remove `useHeadFixed` workaround

userquin:
The same as #1787 using latest version.

EDIT: We should fix the flash on page reload when not using default locale.

closes #1787

### DIFF
--- a/components/status/StatusDetails.vue
+++ b/components/status/StatusDetails.vue
@@ -20,7 +20,7 @@ const createdAt = useFormattedDateTime(status.createdAt)
 
 const { t } = useI18n()
 
-useHeadFixed({
+useHead({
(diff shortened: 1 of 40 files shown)


danielroe (approved):
The flash is a separate issue, I think. This looks good to merge for me (unless I'm misunderstanding what you're talking about with the flash).

userquin:
@danielroe this is my original PR https://github.com/elk-zone/elk/pull/1782, the problem with this PR (1938) is reloading the page we have text flash, `en-US` then `es-ES` (for example)

danielroe:
I believe we have that _anyway_ - `main.elk.zone` shows the same flash.

PR #1939, opened 2023-04-01
fix: remove flashing text on page reload

userquin:
This PR only fixes text flashing on page reload (F5) when using not default locale, we still have the problem with the tabs.

Issue #1776 shouldn't be closed.

supersedes #1782

### DIFF
--- a/app.vue
+++ b/app.vue
@@ -5,7 +5,7 @@ provideGlobalCommands()
 const route = useRoute()
 
 if (process.server && !route.path.startsWith('/settings')) {
-  useHydratedHead({
+  useHead({
     meta: [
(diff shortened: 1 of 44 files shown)


patak-dev (approved):
This works well, thanks @userquin! 
The flash is really bothersome for folks on translations. Let's merge it so we avoid it. (cc @danielroe so you are aware, I don't know if `useHydratedHead` is the best way to do this in Nuxt or if this should be offered out of the box by it).

Ideally, it would be great to have the text already in the HTML as we discussed at one point (using a table with all the entries for the sidebar for example). The text being stable would have a huge effect overall.

PR #1940, opened 2023-04-01
feat: filter for home timeline

### LINKED ISSUE #1931: Filter for Home timeline
As a user I want to filter my timeline to only show me content that is currently relevant to me.

I would like to gain a better UX by having a filter icon e.g. in the top left corner right to the profile icon while being in the Home timeline
When I click on that icon, a list with checkboxes will be shown to me where I can deselect some content
By default all "filters" are applied, so I can just deselect on demand

For the initial draft, I think this doesn't need to get persisted somewhere

Filters could be e.g.:
- post by bot
- post contains sensitive
- post is a repost
- post is from a mutual
- post is from a tag I follow

So by default everything is active, and when I deselect "post by bot", all post that where posted by a bot are just filtered out

This should be only a simple filter feature targeted for non-higher filter usages like using lists or e.g. #1540
The filter also will just be applied client side

It could be discussed if the filter should be swapped from everything is checked by default to "hide X" and then needs to be checked
like:
- hide posts by bot
- hide sensitive posts
- hide reposts
- hide mutual posts (this is unlikely 🤔)
- hide tagged posts (this is also unlucky 🤔)

As we can see, maybe the "hide" does not so much sense, and it is more like I want to deselect everything down to: "show me only from mutuals and tags" or something like that

This feature does NOT aim for higher filters like "tag contains X" or "boost count is higher then y" but instead should just be a quick accessible filter

Shinigami92:
close #1931


components/timeline/TimelineHomeFilter.vue
```diff
@@ -0,0 +1,28 @@
...
+      <div v-else i-ri:filter-2-fill />
+    </button>
+
+    <template #popper>
+      <CommonCheckbox v-model="homeFilter.bot" label="post by bot" />
+      <CommonCheckbox v-model="homeFilter.sensitive" label="post contains sensitive" />
+      <CommonCheckbox v-model="homeFilter.repost" label="post is a repost" />
+      <CommonCheckbox v-model="homeFilter.mutual" label="post is from a mutual" />
+      <CommonCheckbox v-model="homeFilter.tag" label="post is from a tag I follow" />
+    </template>
```
Shinigami92, line 26: TODO: This list needs styling

components/timeline/TimelineHomeFilter.vue
```diff
@@ -0,0 +1,28 @@
...
+      || !homeFilter.value.mutual
+      || !homeFilter.value.tag
+})
+</script>
+
+<template>
+  <VDropdown>
+    <button btn-text>
+      <div v-if="!isActive" i-ri:filter-2-line />
+      <div v-else i-ri:filter-2-fill />
```
Shinigami92, line 17: Is it possible to conditionally swap icons in a better way?

composables/timeline.ts
```diff
@@ -14,7 +14,34 @@ function removeFilteredItems(items: mastodon.v1.Status[], context: mastodon.v1.F
...
+      return false
+
+    if (!sensitive && item.sensitive)
+      return false
+
+    if (!repost && item.reblog != null)
+      return false
+
+    // if (!mutual && ??)
+    //   This would require a lookup of the user's followers
```
Shinigami92, line 34: TODO: find a global way to store a list of all mutual `account.acct` and access this

This could maybe also help in other future features

composables/timeline.ts
```diff
@@ -14,7 +14,34 @@ function removeFilteredItems(items: mastodon.v1.Status[], context: mastodon.v1.F
...
+    //   return false
+
+    // if (!tag && ??)
+    //   This would require a lookup of the user's tags
```
Shinigami92, line 38: TODO: find a global way to store a list of all followed `tag.name` and access this
Sadly the timeline api tags doesn't provide the `following` property 😞 

This could maybe also help in other future features

composables/timeline.ts
```diff
@@ -14,7 +14,34 @@ function removeFilteredItems(items: mastodon.v1.Status[], context: mastodon.v1.F
   const isFiltered = (item: mastodon.v1.Status) => (item.account.id === currentUser.value?.account.id) || !item.filtered?.find(isStrict)
   const isReblogFiltered = (item: mastodon.v1.Status) => !item.reblog?.filtered?.find(isStrict)
 
-  return [...items].filter(isFiltered).filter(isReblogFiltered)
+  const homeFilter = useHomeFilter()
+  const { bot, sensitive, repost, mutual, tag } = homeFilter.value
+
+  const isClientSideHomeFiltered = (item: mastodon.v1.Status) => {
```
Shinigami92, line 20: 1. Somehow the filter needs to be re-applied when the filter changes
2. Not sure right now if the filter is also applied to the "stream" (but for now this is not my prio...)

components/timeline/TimelineHomeFilter.vue
userquin: Cannot use `:class="isActive ? 'i-ri:filter-2-libe' : 'i-ri:filter-2-line'"`

components/timeline/TimelineHomeFilter.vue
userquin: And i18n ;)

components/timeline/TimelineHomeFilter.vue
Shinigami92: Ah 🤦
I forgot that these are just css classes
This is my first frontend js code I'm touching for one month... work (employer) sucks 😓

Shinigami92:
I misread the docs of https://vueuse.org/shared/computedWithControl/#computedwithcontrol 🤦 

Now the reactivity works 👌 but still not finished and the initial is buggy

PR #1966, opened 2023-04-11
chore(deps): replace esno with tsx

Shinigami92:
esno is not maintained anymore and just an alias to tsx


antfu:
I honestly don't see the value of this change (esno is an alias for a different naming, and it's not deprecated). But if you want, I don't mind having it either.

Shinigami92:
> I honestly don't see the value of this change (esno is an alias for a different naming, and it's not deprecated). But if you want, I don't mind having it either.

if it is only an alias, using the `tsx` package is more appropriated due to we get informed via renovate if it gets an update

PR #1968, opened 2023-04-11
feat(i18n): add Korean translation

liberaldev:
Added Korean translation.


patak-dev (approved):
Thanks a lot for the new translation! Looks great! Please stay around to keep it up to date if possible. You can join us at https://chat.elk.zone and look for the #i18n channel if you aren't already there.

PR #1969, opened 2023-04-11
feat: lazy load images

### LINKED ISSUE #1954: elk is unusably slow on long threads with many images
[here is an example post in a thread](https://mastodon.online/@alexxkittle/110152387841009521) from someone who posts work from an artist every day.

If you go to that message with elk, elk appears to load every image in the thread every time.

Then, if you try to respond to the most recent message, the app is so slow that you can't even type in the text box - it takes ~30s for a character to echo to the screen.

Shinigami92:
theoretically closes #1954

https://caniuse.com/intersectionobserver

### DIFF
--- a/components/common/CommonBlurhash.vue
+++ b/components/common/CommonBlurhash.vue
@@ -20,28 +20,26 @@ const placeholderSrc = $computed(() => {
   return getDataUrlFromArr(pixels, 32, 32)
 })
 
+const target = ref<HTMLImageElement | null>(null)
+const targetIsVisible = ref(false)
+
 function loadImage() {
-  const img = document.createElement('img')
-
-  img.onload = () => {
-    isLoaded.value = true
-  }
-
-  img.src = src
-
-  if (srcset)
-    img.srcset = srcset
-
-  setTimeout(() => {
-    isLoaded.value = true
-  }, 3_000)
+  isLoaded.value = true
 }
 
-onMounted(() => {
-  if (shouldLoadImage)
-    loadImage()
-})
+const { stop } = useIntersectionObserver(target, ([{ isIntersecting }]) => {
+  targetIsVisible.value = isIntersecting
 
+  if (isIntersecting) {
+    if (shouldLoadImage)
+      loadImage()
+
+    // Once the image is loaded, we don't need to observe it anymore
+    stop()
+  }
+}, { threshold: 0.2 })
+
+// If the loading is requested lazy from outside, we need to load the image
 watch(() => shouldLoadImage, () => {
   if (shouldLoadImage)
     loadImage()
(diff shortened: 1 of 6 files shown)


Shinigami92:
Works also on my iOS iPhone XS in Opera

antfu (approved):

antfu:
Why remove the `img.onload` hook? It's supposed to render the blurhash before the image is fully downloaded.

Shinigami92:
> Why remove the `img.onload` hook? It's supposed to render the blurhash before the image is fully downloaded.

Because it would load every image twice! (I tested it)
Because of the lazy it now always falls into the static 3 second timeout. So the programmatic onload got never called anymore.

antfu:
But then, with this change, you don't have the blur brush anymore. I guess to have the best of both world, we could have the blurbrush image behind the image so it still renders before the image it loaded. Then we could hook `@load=""` to the real image and remove the blurbrush once the image if fully loaded

Shinigami92:
Rewritten to use https://unlazy.byjohann.dev/integrations/nuxt.html

antfu (approved):
Looking good 👍

johannschopplich:
Hey @Shinigami92 – you were fast! I was about to open a PR with unlazy. Thank you for the effort.

Some notes:
I implemented the `lazyLoad` prop for the Nuxt `UnLazyImage` component today specifically for Elk, to make the `shouldLoadImage` prop compatible. The latter is used for the data saving mode. With the `lazyLoad` prop we can control if the BlurHash should be generated, but no image yet loaded:

```html
<UnLazyImage :lazy-load="shouldLoadImage" />
```

If `shouldLoadImage` is `false`, then only the BlurHash will be generated, but the image not loaded. This makes the `CommonBlurHash` component now compatible with data saving mode in Elk again. There was a bug in unlazy v0.8.2 but fixed in v0.8.4. Could you please update and test again? Thanks!

Shinigami92:
scrolled through the preview app and seems to work :ok_hand:

ayoayco:
@Shinigami92 check this weird behavior on the deployed preview. The page is reloaded when switching tabs. 
lazy-loading-elk

userquin:
@ayoayco can you check if images have width and height set?

ayoayco:
> @ayoayco can you check if images have width and height set?

@userquin I do see width and height properties on  elements that are set

Shinigami92:
If anybody wants to try to fix this tab behavior, feel free to take over the PR, because I don't know what the Unlazy component is doing behind the scene

johannschopplich:
Honestly, I'm unsure why this reload happens. Maybe because of some service worker magic under the hood in the preview environment? `UnLazyImage` should only load images and shouldn't trigger a whole newsfeed refresh. I will take a deeper look later.
@Shinigami92 I've made some minor perf improvements to unlazy today. Would you mind updating to v0.8.8? Thanks!

Shinigami92:
> Would you mind updating to v0.8.8?

Yeah will do later, need to rest currently :sleeping:

patak-dev:
I don't think this issue is related to this PR. I was also able to reproduce it with other PRs like https://github.com/elk-zone/elk/pull/2019. So maybe it is a regression in our test user. (also interesting, in your video you aren't always signed in to the Elk Dev Team Test user 🤔)

ayoayco:
I think @patak-dev saw this on other PRs' preview environments! Will also check later.

ayoayco:
> (also interesting, in your video you aren't always signed in to the Elk Dev Team Test user 🤔)

Yes it was signed out to make sure both environments were fetching from the same server, to compare behavior but was hard due to auto reload.

patak-dev (approved):
It works great for me! Awesome work @johannschopplich and @Shinigami92!

PR #1978, opened 2023-04-14
feat: improve personal notes

JohnyTheCarrot:
The basic `textarea` for the personal notes did not change its height to fit the content.
I made it do this now, but I had to hack it into doing it a bit. The `textarea` tag doesn't listen to you telling it to have its height fit the content, so here's what I did:

I wrapped the `textarea` in a parent div, and gave the `textarea` a sibling `div` that contains the same content as the textarea. This `div` **can** be told to set its height based on its content.
I made the `textarea` absolutely positioned and scaled to be 100% its parent's height.
This parent's height would fit its content which is, because the `textarea` is absolutely positioned, just the sibling `div`.
The result is that the textarea is the height it needs to be.

### DIFF
--- a/components/account/AccountHeader.vue
+++ b/components/account/AccountHeader.vue
@@ -81,6 +81,8 @@ watchEffect(() => {
   iconFields.value = icons
 })
 
+const personalNoteDraft = ref(relationship?.note ?? '')
+
 async function editNote(event: Event) {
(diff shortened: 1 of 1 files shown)


JohnyTheCarrot:
image

patak-dev (approved):

PR #1981, opened 2023-04-15
fix: use `purpose: any` for normal pwa icon

danielroe:
This adds the `any` purpose to our normal PWA icon and reserves `maskable` for just the maskable icon. It looks like this should improve rendering of the PWA icon.

See https://docs.pwabuilder.com/#/builder/manifest?id=icons.


userquin:
check the code below, we've 2 manifest, update also the pwa manifest below...

userquin (approved):

PR #1989, opened 2023-04-17
feat: Add clear button to search input field

manuq:


userquin:
can you focus the input once text deleted?

manuq:
> can you focus the input once text deleted?

ah yes, that was my intention. PR updated.

patak-dev:
I like this idea, but currently, it takes too much space from the visible text. We could have it if we push it to the right (it doesn't matter if the padding isn't the same as the other icon on the left). We could also make it the same color as the search icon.

manuq:
I agree with both things @patak-dev . I just updated the commit to remove some right padding and make the icon color consistent with the left icon.

components/search/SearchWidget.vue
```diff
@@ -64,7 +64,7 @@ function activate() {
 
 <template>
   <div ref="el" relative group>
-    <div bg-base border="~ base" h10 px-4 rounded-3 flex="~ row" items-center relative focus-within:box-shadow-outline gap-3>
+    <div bg-base border="~ base" h10 pl-4 rounded-3 flex="~ row" items-center relative focus-within:box-shadow-outline gap-3>
```
userquin, line 67: don't use padding/margin left/right, check this please, we have some ltr languages: https://github.com/elk-zone/elk/blob/main/CONTRIBUTING.md#rtl-support

manuq: Oh that is fantastic! Fixed.

Issue #1880, opened 2023-03-11
Link length not properly counted in posts


zaidhaan:
Not to be overly pedantic but the URL matching overshoots a bit when compared to mastodon. Specifically, in this example, mastodon would not capture the right brace `)`:
```
foo (https://news.ycombinator.com/)
```

Mastodon counts 29 characters (500 - 471):
![image](https://user-images.githubusercontent.com/5954994/224467406-f3fdd0a5-97e9-47c8-9ffb-ab61695f14a9.png)

Elk counts 28:
![image](https://user-images.githubusercontent.com/5954994/224467455-6e1aaa49-84b5-40be-a69a-d6b1cd04da44.png)

This is an issue since now it's possible to achieve a elk-safe 500 character limit and then receive a 500 character limit exceeded error by Mastodon after submitting which would confuse the user.

So somewhere down the line we might need to update this:
https://github.com/elk-zone/elk/blob/3732a2cc16b4de6cd60f4836151d4a3292c627fa/components/publish/PublishWidget.vue#L72

Since it includes the ending brace:
![image](https://user-images.githubusercontent.com/5954994/224467522-b548918b-9ceb-423e-92bd-6767f7167369.png)

Mastodon themselves have their URL regex defined here: https://github.com/mastodon/mastodon/blob/1ed12d/app/javascript/mastodon/features/compose/util/url_regex.js (which relies on [`twitter-text`](https://github.com/twitter/twitter-text/))

patak-dev:
Maybe we should use the exact same regex from Mastodon here? Including the use of twitter-text. What do you think?

zaidhaan:
Hmm, yes I think that's the best solution.

(Since sure, we _could_ modify the regex for this specific case to handle the ending by appending [`validUrlQueryEndingChars`](https://github.com/twitter/twitter-text/blob/30e2430d90cff3b46393ea54caf511441983c260/js/src/regexp/validUrlQueryEndingChars.js), but that would only address this specific issue, and would not address the underlying larger issue which is all the 23-char limit rules should apply on _valid_ URLs, and the current simple regex doesn't understand valid URLs so even `https://invalid_url_without_even_a_tld` gets truncated. So it's probably best to go with `twitter-text`.)

I think I'll leave this for someone else to pick up since I'm not sure how to get these imports to play nice with TypeScript:
![image](https://user-images.githubusercontent.com/5954994/224566294-ddf89e00-5fe7-4037-9e79-756ca8c5bfa1.png)

[...]

PR #1660, opened 2023-02-06
fix: only count 23 characters per link when composing

4350pChris:
Fix #1651 
limit length of links counting towards character limit to 23

Signed-off-by: Chris-Robin Ennen <chris-robin@ennen.dev>

PR #1990, opened 2023-04-17
fix: Use twitter-text package for URL character counts

### LINKED ISSUE #1977: bug: Incorrect character count on compose box
When composing a message, I had a character count of 496 characters out of my instance's 500 limit.
But when I went to publish it, I got a validation error that the character limit of 500 was exceeded.

When I posted the same text into the vanilla Mastodon web UI, it counted it as 501 characters, so the validation was correct. But the character count for the compose box was not.

image

alec-lefors:
Fixes #1977

Adds twitter-text as a dependency (Mastodon uses this for both their [javascript](https://github.com/mastodon/mastodon/blob/af578e8ce0aabdbe9c0cd3d72d6fa2cc30b7fc66/app/javascript/mastodon/features/compose/util/url_regex.js) and backend implementation) for URL character counts.

Example from the original issue that matches Mastodon's character count:

image


alec-lefors:
Converting this to a draft, still learning Typescript and two type definitions are missing from @type/twitter-text so I will need to write my own types and fix the import statements

danielroe:
Additionally, you should see the context. We're reluctant to include this dependency because of the complexity of the generated regular expression: https://github.com/elk-zone/elk/pull/1660#issuecomment-1419917158

alec-lefors:
> Additionally, you should see the context. We're reluctant to include this dependency because of the complexity of the generated regular expression: [#1660 (comment)](https://github.com/elk-zone/elk/pull/1660#issuecomment-1419917158) and #1880.

Thanks! Did not realize these discussions were already taking place so that’s my bad. I can see why there is reluctance to use the twitter-text regex—  feel free to close.

danielroe:
We felt that maybe we could use a modified/simplified version of the RE? Perhaps this PR could be adapted to (for example) remove some of the domain allow-listing going on?

alec-lefors:
Just fixed the Typescript errors by adding typing for the regexSupplant function and all the regex snippets in the package (@types/twitter-text has a typo for one of the snippets). @types/twitter-text is no longer needed but keeping it in the PR until we're satisfied with the end regex result.

Not sure what happened with the tauri-plugin-log-api failing tests

PR #1991, opened 2023-04-18
fix: Allow account handles to wrap in AccountHeader

### LINKED ISSUE #1902: users with long usernames getting the instance clipped off
this is especially relevant now that there is a live [nostr bridge](https://mostr.pub/) that makes usernames out of the public keys of the users because the second there are multiple bridges up it will be important to know which bridge you are following a nostr account from

image

account in screenshot is https://main.elk.zone/social.wake.st/@76c71aae3a491f1d9eec47cba17e229cda4113a0bbb6e6ae1776d7643e29cafa@mostr.pub

if you navigate to the remote url you get https://mostr.pub/users/76c71aae3a491f1d9eec47cba17e229cda4113a0bbb6e6ae1776d7643e29cafa
which only shows the json representation of the post at the moment.

alec-lefors:
Fixes #1902

### DIFF
--- a/components/account/AccountHeader.vue
+++ b/components/account/AccountHeader.vue
@@ -160,7 +160,7 @@ const isNotifiedOnPost = $computed(() => !!relationship?.notifying)
             <AccountDisplayName :account="account" font-bold sm:text-2xl text-xl />
             <AccountBotIndicator v-if="account.bot" show-label />
           </div>
-          <AccountHandle :account="account" />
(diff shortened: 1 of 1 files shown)


patak-dev:
Thanks @alec-lefors! 

As it stands these accounts don't look good when used as tags in the posts. Hoping there is something better in the future for the integration between the two networks.

patak-dev (approved):

PR #2002, opened 2023-04-22
fix: unicode emojis when editing

### LINKED ISSUE #1932: Editing a post with an emoji causes corruption, emoji conversions
If you have an emoji in the middle of a post, and you try editing it, most edits at the end will cause mangling of your changed text, with things turning into emojis.  Depending on the emoji, it also seems like sometimes it is getting split into separate unicode codepoints in which case it will be one invalid one and one emoji, the server will reject the message, and you have to delete the invalid one (and risk more conversion) to actually post the edit.

As you can see in the video, I discovered the emoji has to be in the middle of the post to trigger the error. I'm using Elk as a PWA "installed" using Chrome on Android.

https://user-images.githubusercontent.com/61129/228559145-b15063a2-f4e4-4015-884d-8a5c867a148e.mp4

### LINKED ISSUE #1994: Pressing enter when editing adds an emoji instead
When editing a post, if you press enter in/after a paragraph that contains an emoji, instead of adding a new line, it just adds that emoji again.

chrome_3O4iepDTVB

In this gif the only keys I press are enter (3x at a time) and ctrl+z.

patak-dev:
Fixes #1932
Fixes #1994

@antfu do you recall why we wanted to preserve unicode emojis here? It seems that converting them is the right thing to do so TipTap can properly handle them.

### DIFF
--- a/composables/content-parse.ts
+++ b/composables/content-parse.ts
@@ -145,7 +145,6 @@ export function convertMastodonHTML(html: string, customEmojis: Record<string, m
   const tree = parseMastodonHTML(html, {
     emojis: customEmojis,
     markdown: true,
-    replaceUnicodeEmoji: false,
     convertMentionLink: true,
   })
   return render(tree)


antfu:
I don't remember. We could give this a try

PR #2003, opened 2023-04-22
fix: Add zen mode to preferences, remove shortcut 

### LINKED ISSUE #1883: Make the "zen mode" button minimally findable
It took me a while to figure out why left side main menu bar (Home, Notifications, Conversations, etc.) started fading out when I'm not actively pointing at it, with no indication that there is anything available by pointing there while it is hidden, for no reason as far as I could tell as it just leaves that area of the window as unused wasted space -- but after searching for the missing docs, scouring all the preferences, and fiddling around with the window size, I started pointing at random parts of the window and I finally noticed the fine print nav on the other side that also only appears when you're pointing at it has buttons on it, and this is apparently a feature, not a bug: "Zen Mode".

Okayyyy.

Could you maybe:
- Make the Zen Mode setting also appear in the place in that people are rationally going to look for a setting like that, which is the Interface section of the Settings, like you already do for the Toggle Theme button right next to it
- Better yet make the things that disappear in Zen Mode when hidden show some minimal indication that there's something to find there

manuq:
Fixes #1883

### DIFF
--- a/plugins/magic-keys.client.ts
+++ b/plugins/magic-keys.client.ts
@@ -23,7 +23,6 @@ export default defineNuxtPlugin(({ $scrollToTop }) => {
   }
 
   whenever(logicAnd(notUsingInput, keys['?']), toggleKeyboardShortcuts)
-  whenever(logicAnd(notUsingInput, keys.z), () => userSettings.value.zenMode = !userSettings.value.zenMode)
 
   const defaultPublishDialog = () => {
     const current = keys.current
--- a/plugins/magic-keys.client.ts
+++ b/plugins/magic-keys.client.ts
@@ -2,7 +2,6 @@ import type { RouteLocationRaw } from 'vue-router'
 import { useMagicSequence } from '~/composables/magickeys'
 
 export default defineNuxtPlugin(({ $scrollToTop }) => {
-  const userSettings = useUserSettings()
   const keys = useMagicKeys()
   const router = useRouter()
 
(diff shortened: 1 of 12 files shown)


patak-dev:
Thanks @manuq! It works nicely!

There is only one issue I found while testing. Before you press the `Zen Mode` button or setting for the first time (so its value is still undefined in the settings), if you reload the page you'll notice the sidebars start in zen mode true, and they fade in. If you go into Zen Mode, then toggle back (so the setting is now zenMode: false), this doesn't happen.

manuq:
Good catch @patak-dev ! It turns out that all preferences started as `undefined`, and then the storage was in charge of setting them. Thus the fade-in effect that happened right away. I have appended a commit to fix it.

plugins/magic-keys.client.ts
```diff
@@ -23,7 +23,6 @@ export default defineNuxtPlugin(({ $scrollToTop }) => {
   }
 
   whenever(logicAnd(notUsingInput, keys['?']), toggleKeyboardShortcuts)
-  whenever(logicAnd(notUsingInput, keys.z), () => userSettings.value.zenMode = !userSettings.value.zenMode)
```
patak-dev, line 26: We should remove the `useUserSettings` in this file too

patak-dev:
The flash was reduced but it is still there. You can compare reloading against elk.zone to see it.

userquin:
@manuq starting dev server from scratch using your branch using zen mode by default (`ni && nr dev`)

userquin:
> The flash was reduced but it is still there. You can compare reloading against elk.zone to see it.

On my local there is no flash (also using PR preview), maybe missing transition? I see a fade-out transition when activating zen mode while deactivating, it is almost instant.

patak-dev:
Ok, I can't reproduce locally. Let's try it out, and we fix it if not later one

PR #2007, opened 2023-04-23
fix: emoji sizing

### LINKED ISSUE #1983: Odd custom emoji rendering for @forestbot.
See: https://hellsite.site/@forestbot/110208083719403771

As seen on hellsite.site

As seen in Elk

patak-dev:
Fixes #1983

@userquin @cyberalien maybe you could check this one? I don't know why we were using different sizes for custom emojis and for normal ones, and why we were using max-width/max-height instead of direct width/height.

The forest repro looks good after this PR
image

### DIFF
--- a/styles/global.css
+++ b/styles/global.css
@@ -34,25 +34,17 @@ body {
   display: none;
 }
 
-.custom-emoji {
-  display: inline-block;
-  vertical-align: text-bottom;
-}
-
 .content-editor img.custom-emoji,
-.custom-emoji img {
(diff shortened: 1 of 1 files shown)


antfu:
I think that was because some of the custom emojis are not squared? We may move on until we see another case

cyberalien:
Tested, works correctly.

Twitter emojis are built for 18x18 grid and are used with that size. Custom emojis don't have fixed size. This fix makes them look identical, so behaviour now matches Mastodon client.

PR #2024, opened 2023-04-26
fix: detect Takahe hashtags correctly

chdorner:
Takahē hashtag links always end with a trailing slash which Elk's regex didn't recognize. With this change hashtag URLs are allowed to have an optional trailing slash.

i.e. `https://takahe.social/tags/takahe/`


patak-dev (approved):

patak-dev:
Thanks!

PR #2035, opened 2023-04-30
chore(deps): update to Nuxt 3.4.3

patak-dev:
Testing to remove the context-related console error.


patak-dev:
@danielroe I may let you update Nuxt. Could you tell me later what you do? Do you run nuxi upgrade? (here I did it by hand because this one https://github.com/elk-zone/elk/pull/1944 is also failing)

Issue #967, opened 2023-01-11
language selection improvements


severak:
Hi, as person who is switching between two languages (Czech and English) on my account I have two minor ideas how to improve post language selector:

- it should remember last selected language and for new posts use this
- it should be seen which language is currently chosen without going to dropdown (I would use language code for this purpose - e.g. show CZ or EN next to language chooser)

alper:
Not to mention #1550

kimusan:
I agree with this. Maybe even make it possible to limit the list of languages to just the one the use would use. I switch between 3–4 languages and really don't need African or Asian languages in my list.

zordsdavini:
there should be possibility to add manual code as all languages never will be in the list. For example I write in Samogitian (sgs)

PR #2042, opened 2023-05-01
feat: prefer en language as top

Shinigami92:
This is a really small but effective change!

Issues like https://github.com/elk-zone/elk/issues/967 are asking for improvements for the language selection. I'm also affected by this myself, because my native language is german but I write posts in english very often.

So english is (I assume) the most common language in the world and many folks write post in english beside their native language. That's why now this PR moves `en` to the top if nothing was searched for manually.

I think this is the most easiest and smallest solution for now we could think of. If we want to improve it further with any smart stuff, we could enhance and iterate later in further PRs.

### DIFF
--- a/components/publish/PublishLanguagePicker.vue
+++ b/components/publish/PublishLanguagePicker.vue
@@ -20,6 +20,10 @@ const languages = $computed(() =>
     ? fuse.search(languageKeyword).map(r => r.item)
     : [...languagesNameList].filter(entry => !userSettings.value.disabledTranslationLanguages.includes(entry.code))
         .sort(({ code: a }, { code: b }) => {
+          // Prefer code en over everything else
+          if (a === 'en')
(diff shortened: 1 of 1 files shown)


components/publish/PublishLanguagePicker.vue
```diff
@@ -20,6 +20,10 @@ const languages = $computed(() =>
     ? fuse.search(languageKeyword).map(r => r.item)
     : [...languagesNameList].filter(entry => !userSettings.value.disabledTranslationLanguages.includes(entry.code))
         .sort(({ code: a }, { code: b }) => {
+          // Prefer code en over everything else
```
antfu, line 23:
```suggestion
          // Put English on the top
```

antfu (approved):
Sounds reasonable to me!

PR #2045, opened 2023-05-02
fix: notification templates for admin reports

### LINKED ISSUE #1809: Add template case for admin.report notifications
Similar to `notification.type === 'admin.sign_up'` in `NotificationCard.vue`, Masto 4.0's notification for new reports should be displayed.

Notification type is `admin.report`, currently an error is shown:
image

kytta:
This PR adds the notification template case for admin.report notifications. Closes #1809

-----
<a href="https://stackblitz.com/~/github.com/kytta/elk/tree/kytta%2Fissue1809">Review PR in StackBlitz Codeflow</a> _Submitted with [StackBlitz Codeflow](https://stackblitz.com/~/github.com/kytta/elk/tree/kytta%2Fissue1809)._

### DIFF
--- a/components/notification/NotificationCard.vue
+++ b/components/notification/NotificationCard.vue
@@ -39,6 +39,22 @@ const { notification } = defineProps<{
         <span>{{ $t("notification.signed_up") }}</span>
       </div>
     </template>
+    <template v-else-if="notification.type === 'admin.report'">
+      <NuxtLink :to="getReportRoute(notification.report.id)">
+        <div flex p3 items-center bg-shaded>
+          <div i-ri:admin-fill me-1 color-purple />
+          <AccountDisplayName
+            :account="notification.account"
+            text-purple me-1 font-bold line-clamp-1 ws-pre-wrap break-all
+          />
+          <span>{{ $t("notification.reported") }}</span>
+          <AccountDisplayName
+            :account="notification.report.target_account"
+            text-purple me-1 font-bold line-clamp-1 ws-pre-wrap break-all
+          />
+        </div>
+      </NuxtLink>
+    </template>
     <template v-else-if="notification.type === 'follow_request'">
       <div flex ms-4 items-center class="-top-2.5" absolute inset-ie-2 px-2>
         <div i-ri:user-follow-fill text-xl me-1 />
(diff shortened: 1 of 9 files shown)


kytta:
This is not ready, I need some help from the maintainers.

First, what is the best way to return the Route to the actual Mastodon instance? Is there a Router entity that would achieve this for me?

Second, I have problems with `masto`'s `Notification` type, as it does not include the optional `report` attribute ([which is present in the actual entity](https://docs.joinmastodon.org/entities/Notification/))

Third, I have only added the English localization for now. In other languages (like German), the translation would require two words (_$USER hat $TARGET_USER gemeldet_). Is there a way to author i18n messages with templates? I haven't got much experience with the library.

Apart from that, I think this is enough for the issue to be fixed.

kytta:
I'm happy with the end result so far. The only problem I'm having right now is the spacing between "reported" and the target account name:

image

No idea how to fix it :/

components/notification/NotificationCard.vue
```diff
@@ -39,6 +39,23 @@ const { notification } = defineProps<{
...
+        <div flex p3 items-center bg-shaded>
+          <div i-ri:flag-fill me-1 color-purple />
+          <i18n-t keypath="notification.reported">
+            <AccountDisplayName
+              :account="notification.account"
+              text-purple me-1 font-bold line-clamp-1 ws-pre-wrap break-all
+            />
+            <AccountDisplayName
+              :account="notification.report?.targetAccount!"
+              text-purple me-1 font-bold line-clamp-1 ws-pre-wrap break-all
```
patak-dev, line 53: This fixes the spacing issue
```suggestion
              text-purple ms-1 font-bold line-clamp-1 ws-pre-wrap break-all
```

patak-dev (approved):

patak-dev:
Thanks for the PR!

composables/masto/routes.ts
```diff
@@ -29,6 +29,10 @@ export function getAccountFollowersRoute(account: mastodon.v1.Account) {
   })
 }
 
+export function getReportRoute(id: string | number) {
+  return `https://${currentUser.value?.server}/admin/reports/${id}`
```
danielroe, line 33:
```suggestion
  return `https://${currentUser.value?.server}/admin/reports/${encodeURIComponent(id)}`
```

PR #2046, opened 2023-05-02
chore: upgrade to nuxt v3.4.3

danielroe:
After updating version constraints in `package.json`, I ran:

```sh
pnpm up -rL nuxt
pnpm dedupe
```


patak-dev (approved):
Thanks for updating it Daniel! I tested it a bit and looks good

PR #2047, opened 2023-05-02
fix: patch pwa plugin

userquin:
Check docker build using Vite instead Rollup to build service worker


patak-dev:
Maybe we no longer need this after https://github.com/elk-zone/elk/pull/2046

userquin:
it works, closing it, we're changing pwa plugin...

PR #2064, opened 2023-05-05
feat: recover Search/Explore pages separation

### LINKED ISSUE #2063: New well-being preference: Hide trends posts on Explore page by default
This is a similar idea to #1965.

## Description of the problem
In the widescreen view, we can search a keyword by typing the search box on the sidebar without viewing the trend post tab (`/explore`). However, in the small screen view and mobile view, we have to move to the Explore page and are forced to see the trends posts. That sometimes distracting since we cannot control which posts can be shown.

Personally, this was a pain point to use Elk on mobile devices after the Explore page had been merged into the search tab a while ago.

## Suggested solution
Add a new well-being option to make the Explore page not show the trends post tab by default.

shuuji3:
This resolves #2063

# Example screenshots
**Before** (previous `/explore` and new `/explore/posts` page)
Screenshot 2023-05-05 at 16 32 22

**After** (new `/explore` page)
Screenshot 2023-05-05 at 16 31 52


patak-dev:
I don't think we need a new option. We wanted to revert the change that merged explore and search but never got to it. See https://github.com/elk-zone/elk/pull/1301#issuecomment-1416690067. There is an open PR but I think it isn't working right https://github.com/elk-zone/elk/pull/1704.

@shuuji3 could you remove the new option and directly revert what was done regarding the merge of explore/search in #1301 (there are other things too in that PR)? Or maybe yours already work if you remove the option.

shuuji3:
Ah, thanks for pointing out. I missed that discussion. I'm going to remove the option and hide the trends post by default at first as it's easier to modify.

But it would be slightly different from the previous state so after that, I'll check #1704 and update the code to restore the previous behavior later.

shuuji3:
[44d1763](https://github.com/elk-zone/elk/pull/2064/commits/44d1763b4138e22fdc5215b4107d7f31b7d3b869) is an incomplete change so it's better to adopt #2075.

PR #2065, opened 2023-05-05
feat: respect Media Display preferences

### LINKED ISSUE #1002: "Always expand content warnings" and "always show media" settings do not transfer and are not findable
### Pre-Checks

### Describe the bug

I've never submitted a bug report here before. I copied this format from a couple other issues. Please be gentle if I messed up!

I have the "Always expand posts marked with content warnings" and "Always show media" settings enabled on disabled.social.

When I logged into Elk, posts marked with content warnings and posts with just media marked as sensitive both stayed hidden, the default setting on Mastodon. I looked for a way to turn these settings back on and could not find it.

### Suggested solution

Make it so these settings transfer into Elk when someone logs in, like how my filters transferred just fine.

### Alternative

Add an easy-to-find way to re-enable these settings after logging into Elk.

### Additional context

If it's relevant, I'm on disabled.social, and I'm using Firefox on desktop. I've also attached a screenshot of one of the posts with a content warning that didn't expand, since the pre-check said to add a screenshot, though I'm not sure how helpful that is in this case.

Thank you for making this client!

image

### Additional Context

_No response_

ayoayco:
fixes #1002 

This PR will respect two preferences with regards to media display: `show all`, or `hide all`
1. `show all` will expand all media hidden under a spoiler UI
2. `hide all` will put all media (even non-sensitive ones) under a spoiler UI

Also added a new translation string to mark if media is hidden, same with Mastodon web. Example below is a non-sensitive post but hidden to respect the "hide all" preference:
| Mastodon | Elk |
|--------|--------|
| Screenshot 2023-05-05 at 11 22 44 AM | Screenshot 2023-05-05 at 11 22 54 AM |

### DIFF
--- a/components/status/StatusContent.vue
+++ b/components/status/StatusContent.vue
@@ -22,7 +22,7 @@ const isFiltered = $computed(() => status.account.id !== currentUser.value?.acco
 // check spoiler text or media attachment
 // needed to handle accounts that mark all their posts as sensitive
 const hasSpoilerOrSensitiveMedia = $computed(() => !!status.spoilerText || (status.sensitive && !!status.mediaAttachments.length))
-const unfilteredSensitive = !isFiltered && status.sensitive && !status.spoilerText
+const unfilteredSensitive = computed(() => !isFiltered && status.sensitive && !status.spoilerText)
 const hideAllMedia = computed(
   () => {
     return currentUser.value ? (getHideMediaByDefault(currentUser.value.account) && !!status.mediaAttachments.length) : false
--- a/components/status/StatusContent.vue
+++ b/components/status/StatusContent.vue
@@ -38,7 +38,7 @@ const hideAllMedia = computed(
       'ms--3.5 mt--1 ms--1': isDM && context !== 'details',
     }"
   >
-    <StatusBody v-if="isSensitiveNonSpoiler || hideAllMedia" :status="status" :newer="newer" :with-action="!isDetails" :class="isDetails ? 'text-xl' : ''" />
+    <StatusBody v-if="(!isFiltered && isSensitiveNonSpoiler) || hideAllMedia" :status="status" :newer="newer" :with-action="!isDetails" :class="isDetails ? 'text-xl' : ''" />
     <StatusSpoiler :enabled="hasSpoilerOrSensitiveMedia || isFiltered" :filter="isFiltered" :sensitive-non-spoiler="isSensitiveNonSpoiler || hideAllMedia" :is-d-m="isDM">
       <template v-if="status.spoilerText" #spoiler>
         <p>{{ status.spoilerText }}</p>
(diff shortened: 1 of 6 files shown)


ayoayco:
Ah sorry missed some need to fix typechecks...

patak-dev:
ya, better to remove the lock file changes entirely for this PR

components/status/StatusContent.vue
```diff
@@ -22,6 +22,12 @@ const isFiltered = $computed(() => status.account.id !== currentUser.value?.acco
 // check spoiler text or media attachment
 // needed to handle accounts that mark all their posts as sensitive
 const hasSpoilerOrSensitiveMedia = $computed(() => !!status.spoilerText || (status.sensitive && !!status.mediaAttachments.length))
+const unfilteredSensitive = !isFiltered && status.sensitive && !status.spoilerText
```
patak-dev, line 25: This should be a computed

ayoayco: pushed a change to make it computed

components/status/StatusContent.vue
```diff
@@ -32,15 +38,15 @@ const hasSpoilerOrSensitiveMedia = $computed(() => !!status.spoilerText || (stat
...
+    <StatusBody v-if="unfilteredSensitive || hideAllMedia" :status="status" :newer="newer" :with-action="!isDetails" :class="isDetails ? 'text-xl' : ''" />
+    <StatusSpoiler :enabled="hasSpoilerOrSensitiveMedia || isFiltered" :filter="isFiltered" :unfiltered-sensitive="unfilteredSensitive || hideAllMedia" :is-d-m="isDM">
       <template v-if="status.spoilerText" #spoiler>
         <p>{{ status.spoilerText }}</p>
       </template>
       <template v-else-if="filterPhrase" #spoiler>
         <p>{{ `${$t('status.filter_hidden_phrase')}: ${filterPhrase}` }}</p>
       </template>
-      <StatusBody v-if="!status.sensitive || status.spoilerText" :status="status" :newer="newer" :with-action="!isDetails" :class="isDetails ? 'text-xl' : ''" />
+      <StatusBody v-if="!(unfilteredSensitive || hideAllMedia)" :status="status" :newer="newer" :with-action="!isDetails" :class="isDetails ? 'text-xl' : ''" />
```
patak-dev, line 49: We weren't checking for `isFiltered` before here, is it ok to do it now?

ayoayco: good catch, pushed a change for this.

Though I think for further improvement we need to have a different handling for hiding media. Maybe using the blurhash like in Mastodon. With what we have now, there's a complex scenario that Elk cannot support, which is if the post has a CW and a sensitive media, and the user viewing set "expand all CWs" but have "hide all media".

patak-dev (approved):

PR #2067, opened 2023-05-05
feat: add 'use star favorite icon' preference

### LINKED ISSUE #2058: Option to use a star icon instead of a heart icon for favourites
Give users the option to change the "heart" icon for favourites to a "star" icon to be more aligned with Mastodon's official design system.

Reference: https://elk.zone/mastodon.uno/@amministratore/110304288599350691

shuuji3:
resolves #2058

This will add an additional option to change the favorite icon.

This will change the color of the icon from rose to yellow but this causes the same color usage as the bookmark icon.

Screenshot 2023-05-05 at 20 06 20


shuuji3:
That's true. I think we can distinguish the behavior with the help of tooltips and the user will be intentional to enable this feature. Similarly, we already have the same heart icons between the favorite and right-bottom donations but there is no confusion for them.

Screenshot 2023-05-05 at 22 49 06
Screenshot 2023-05-05 at 22 49 27

patak-dev:
Nice! Two details, we could use red for bookmark so it isnt yellow like the star. So swap the colors if the new option is enabled.

And we should add a new env variable to let people hosting the app which should be the default. At least there is one server that would like star by default. We could do this in another PR thought

shuuji3:
> Nice! Two details, we could use red for bookmark so it isnt yellow like the star. So swap the colors if the new option is enabled.

That's a good idea! 🙂  I'll update the PR soon.

As for the environment variable, maybe I can work on that. But I'm not familiar with those settings yet, so I'll leave that to another PR.

shuuji3:
Updated the bookmark icon color.

Screenshot 2023-05-06 at 1 30 38

Shinigami92 (commented):
If `getPreferences(userSettings, 'useStarFavoriteIcon')` is used multiple times in a component, should it be extracted into a computed to safe performance?

patak-dev:
I don't think it will affect performance, but I agree it is better to extract it 👍🏼

shuuji3:
Thanks, I tried the `usePreferences()` to get `computed` for `useStarFavoriteIcon`. This also could simplify some templates.

components/status/StatusActionsMore.vue
```diff
@@ -149,8 +150,8 @@ function showFavoritedAndBoostedBy() {
 
           <CommonDropdownItem
             :text="status.favourited ? $t('action.favourited') : $t('action.favourite')"
-            :icon="status.favourited ? 'i-ri:heart-3-fill' : 'i-ri:heart-3-line'"
-            :class="status.favourited ? 'text-rose' : ''"
+            :icon="`${useStarFavoriteIcon ? 'i-ri:star-' : 'i-ri:heart-3-'}${status.favourited ? 'fill' : 'line'}`"
```
Shinigami92, line 153: You should try to always have the classes as fully written out and not dynamically calculated and concatinated because this results in issues

try something like

```ts
useStarFavoriteIcon
  ? status.favourited ? 'i-ri:star-fill'    : 'i-ri:star-line'
  : status.favourited ? 'i-ri:heart-3-fill' : 'i-ri:heart-3-line'
```

this also applies to lines below

AstroOrbis (approved):

components/status/StatusActionsMore.vue
shuuji3: OK, this was fixed by a new commit. 👍🏻

Shinigami92 (approved):
Another one should also approve :slightly_smiling_face:

patak-dev (approved):
Awesome 🌟

PR #2070, opened 2023-05-05
feat: show followers from account's instance

Shinigami92:
@danielroe here you can see how it would be possible to additionally load the rest of the paginator if the account is not part of the instance you are currently have your own account

Questions:
- how could it be possible to merge the paginators?
- should we cache other mastodon clients or create them always on demand?
- should we try somehow to merge the paginators? (this could also help for some other features if we find a way to merge paginators in general)
- or should we add e.g. a button at the bottom to load them if wanted?
- maybe we don't need to load the paginator of the instance from your own account at all but just directly use the account's instance :thinking: :eyes: (does they include always all?)

For fast communication: https://discord.com/channels/1044887051155292200/1104389271877587004


Shinigami92:
Still interested in this feature, but leaving it to other contributors

PR #2071, opened 2023-05-06
feat: search back to the right

patak-dev:
It doesn't seems we'll be adding things on the right sidebar for a while more, and moving the search widget to the left was a controversial move from the start.

I'm thinking we should move it back to the right, at least until we have more things in that sidebar to justify the move.

image

### DIFF
--- a/components/nav/NavSide.vue
+++ b/components/nav/NavSide.vue
@@ -7,7 +7,6 @@ const { notifications } = useNotifications()
 
 <template>
   <nav sm:px3 flex="~ col gap2" shrink text-size-base leading-normal md:text-lg h-full mt-1 overflow-y-auto>
-    <SearchWidget lg:ms-1 lg:me-5 hidden xl:block />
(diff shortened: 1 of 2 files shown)


antfu (approved):

Shinigami92:
Wait!!!

Can this be an option? Please?

I personally much more prefer the search at the left and it drives my UX much more and I use it there much more. When it was on the left, I sometimes even did not noticed it, even if it is obvious and still there.

patak-dev:
I would prefer to avoid the option as it may make things difficult as the ui evolves. If there are a lot of people wants it, we could evaluate later

patak-dev:
Merging so we see how folks using https://main.elk.zone react to this change. We can keep discussing the option for later, but we should find strong reasons for it.

Shinigami92:
Made a poll: https://main.elk.zone/mas.to/@Shini92/110323293448460642

PR #2072, opened 2023-05-06
fix: regression with media hidden

ayoayco:
This PR will fix a regression on my last PR where if a post is sensitive but does not have a media attached, it shows the spoiler UI. Also sneaked in a fix for a typo in the Tagalog translation


ayoayco:
Sorry the commits are messy, they're going to get squashed when merged right?

patak-dev (approved):

PR #2075, opened 2023-05-06
feat: separate search and explore

patak-dev:
Closes #2064
Closes #1704

See @TodePond post for reference https://github.com/elk-zone/elk/pull/1301#issuecomment-1416490216
 
Partial revert of #1301

@shuuji3 it was easier to do the revert, please test this one. Thanks!

### DIFF
--- a/components/nav/NavBottom.vue
+++ b/components/nav/NavBottom.vue
@@ -16,7 +16,7 @@ const { notifications } = useNotifications()
       <NuxtLink to="/home" :aria-label="$t('nav.home')" :active-class="moreMenuVisible ? '' : 'text-primary'" flex flex-row items-center place-content-center h-full flex-1 class="coarse-pointer:select-none" @click="$scrollToTop">
         <div i-ri:home-5-line />
       </NuxtLink>
-      <NuxtLink :to="isHydrated ? `/${currentServer}/explore` : '/explore'" :aria-label="$t('nav.explore')" :active-class="moreMenuVisible ? '' : 'text-primary'" flex flex-row items-center place-content-center h-full flex-1 class="coarse-pointer:select-none" @click="$scrollToTop">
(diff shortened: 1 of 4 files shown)


TodePond:
My hero! This is a great digital well-being improvement for me! My prediction is that some people might not like this 🤔 Ideally we could put this behind an option, but I understand that that would come with its own cost!

patak-dev:
Yes, lets add the option if there is a lot of demand. Even then I would push back to try to avoid it though.

shuuji3 (approved):
@patak-dev This is exactly what I was planning to change on #2064. Thanks for helping 🙂

components/nav/NavBottom.vue
```diff
@@ -16,7 +16,7 @@ const { notifications } = useNotifications()
       <NuxtLink to="/home" :aria-label="$t('nav.home')" :active-class="moreMenuVisible ? '' : 'text-primary'" flex flex-row items-center place-content-center h-full flex-1 class="coarse-pointer:select-none" @click="$scrollToTop">
         <div i-ri:home-5-line />
       </NuxtLink>
-      <NuxtLink :to="isHydrated ? `/${currentServer}/explore` : '/explore'" :aria-label="$t('nav.explore')" :active-class="moreMenuVisible ? '' : 'text-primary'" flex flex-row items-center place-content-center h-full flex-1 class="coarse-pointer:select-none" @click="$scrollToTop">
+      <NuxtLink to="/search" :aria-label="$t('nav.explore')" :active-class="moreMenuVisible ? '' : 'text-primary'" flex flex-row items-center place-content-center h-full flex-1 class="coarse-pointer:select-none" @click="$scrollToTop">
```
shuuji3, line 19: It looks like the `aria-label` was not changed here:

```suggestion
      <NuxtLink to="/search" :aria-label="$t('nav.search')" :active-class="moreMenuVisible ? '' : 'text-primary'" flex flex-row items-center place-content-center h-full flex-1 class="coarse-pointer:select-none" @click="$scrollToTop">
```

patak-dev: Good catch!

PR #2076, opened 2023-05-06
feat: review compose buttons

patak-dev:
We received feedback that:
- It is confusing to have two Compose buttons + the upper publish widget (after adding the one at the top)
- We add the compose button because it was difficult to spot "compose" after we added it to the sidebar as a route instead of a button.

This PR removes the new compose button at the top and instead makes the "compose" on the sidebar more prominent and different from the rest:
- a new spacer above it
- only one using a fill icon, and the icon is always primary (like the button above)
- text is bold

I think this is a good middle ground to try. I would still prefer to avoid using a big orange Compose button on the sidebar. I think it is too distracting (the reason that pushes us in the route direction)

image

### DIFF
--- a/components/nav/NavSide.vue
+++ b/components/nav/NavSide.vue
@@ -27,7 +27,7 @@ const useStarFavoriteIcon = usePreferences('useStarFavoriteIcon')
     <NavSideItem :text="$t('nav.bookmarks')" to="/bookmarks" icon="i-ri:bookmark-line" user-only :command="command" />
 
     <div class="spacer" shrink hidden sm:block />
-    <NavSideItem :text="$t('action.compose')" to="/compose" user-only :command="command" font-bold>
+    <NavSideItem :text="$t('action.compose')" to="/compose" icon="i-ri:quill-pen-fill" user-only :command="command" font-bold>
       <template #icon>
         <div class="i-ri:quill-pen-fill" text-xl text-primary />
       </template>
--- a/components/nav/NavSide.vue
+++ b/components/nav/NavSide.vue
@@ -27,7 +27,7 @@ const useStarFavoriteIcon = usePreferences('useStarFavoriteIcon')
     <NavSideItem :text="$t('nav.bookmarks')" to="/bookmarks" icon="i-ri:bookmark-line" user-only :command="command" />
 
     <div class="spacer" shrink hidden sm:block />
-    <NavSideItem :text="$t('action.compose')" to="/compose" icon="i-ri:quill-pen-fill" user-only :command="command" font-bold />
+    <NavSideItem :text="$t('action.compose')" to="/compose" icon="i-ri:quill-pen-line" user-only :command="command" />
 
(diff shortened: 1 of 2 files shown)


Shinigami92:
looks good, I think the bold might not be needed

patak-dev:
I tried without bold and it still felt it needed more pop

ZMYaro:
What about, instead of the bold and filled icon (which are associated with it being the selected navigation option in a lot of design languages, and may confuse people), make it look like a button (rounded rectangle with accent color background)?

antfu:
Now I am lost the initial motivation - is it because ppl say it does not look like a button or so? Honestly, the current design gives me a feeling that we are over-emphasize the compose entry, which feels a bit necessary and easy to confuse with the active state. I'd suggest we could keep the very original design but increasing the upper space to make it a separate group.

patak-dev:
@ZMYaro it looked like a button for a long time, it is how it looks on Twitter. But we wanted to avoid a big orange button on the sidebar. I still think that is a good idea, but we could review that in the future

@antfu, (1) this is how it looks only with the spacing
image

(2) Here is with spacing + font-bold
image

(3) Same + fill icon
image

The spacing only already helps a lot. Maybe it is enough. I'm fine going with (1), or if you like (2) or (3) also do that to make it a bit more visible.

antfu:
My concern is why we need to treat the Compose tab specially? We design a standalone page to make it more like a tab rather than an action, which to me we should either treat it as equal as the rest tabs, or make it special enough like the previous button on the top.

patak-dev:
I think we need to treat Compose especially because it is the main action of the app. I like the idea of not having a flashy button but it seems users are having issues finding the current Compose tab. I'm not that happy about the new small compose button we added next to the back arrow. So the idea here was to have a middle ground.

Maybe now that the search is back to the right and there is less noise on the left sidebar, just adding the space as you suggest is enough.

antfu:
Thanks! I am open to making it more button (maybe a FAB) in the future if we find a good place to put it.

ZMYaro:
> I'd suggest we could keep the very original design but increasing the upper space to make it a separate group.

That has the same problem the added button was presumably intended to solve—that it looks like a navigation section and not an _action_.

> the current design gives me a feeling that we are over-emphasize the compose entry, which feels a bit necessary and easy to confuse with the active state. 

> I think we need to treat Compose especially because it is the main action of the app. I like the idea of not having a flashy button but it seems users are having issues finding the current Compose tab.

I agree we need it to make it look more like a button.  Obviously that is easier said than done with Elk's ultra-minimalist UI design language.  One option is we could change the active nav item to look more like its hover state (neutral text, light translucent background) and make the compose button the accent color (like a hyperlink), but that still has the potential to confuse a lot of people.  I think it is also fine to leave it as is and come back to it later if any better ideas come up.

One small thing we could do is move it to the top now that the search box has been moved.  That way at least it is the first thing on the list people see.

PR #2088, opened 2023-05-10
fix: user acct not respecting domain

patak-dev:
We were constructing the `acct` for the active user as `handle@server` instead of `handle@domain` (for example `vite@m.webtoo.ls` instead of `vite@webtoo.ls`). You can reproduce by hovering over your account on the user picker, and if you click on it in the URL you'll see a URL like `https://elk.zone/m.webtoo.ls/@elk@m.webtoo.ls` instead of `https://elk.zone/m.webtoo.ls/@elk`.

@userquin, in the end, we had everything in place to respect the instance domain (using `uri` in Mastodon and also supporting GoToSocial). This was also used as the key in local storage so there is backward compat code.

### DIFF
--- a/composables/cache.ts
+++ b/composables/cache.ts
@@ -59,12 +59,12 @@ export function fetchAccountById(id?: string | null): Promise<mastodon.v1.Accoun
 export async function fetchAccountByHandle(acct: string): Promise<mastodon.v1.Account> {
   const server = currentServer.value
   const userId = currentUser.value?.account.id
-  const userAcct = acct.endsWith(`@${server}`) ? acct.slice(0, -server.length - 1) : acct
+  const domain = currentInstance.value ? getInstanceDomain(currentInstance.value) : undefined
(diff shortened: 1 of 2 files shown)


userquin:
We've some race condition, I've not checked code yet:

https://streamable.com/cwilv8

userquin:
Still some race condition:
- sign-in with webtoo.ls account
- find a webtoo.ls notification
- go to previous notification
- switch to Elk Dev user

imagen

userquin:
Problem present in main (https://github.com/elk-zone/elk/pull/2087), error page then redirect when switching accounts:

imagen

patak-dev:
Thanks @userquin, looks like fixing that errors will require a refactoring that is out of the scope of this PR.

@danielroe I think we can move forward with this PR. At one point in the future, we should probably have a single `currentUser` ref and compute everything in sync when switching (currentUser.account, currentUser.server, currentUser.instance).

danielroe (approved):

userquin (approved):

PR #2111, opened 2023-05-20
feat: poll creation

### LINKED ISSUE #1668: Editing a post with a poll results in the poll being deleted
Bug originally identified when [this post](https://mstdn.science/@flodebarre/109819299827220428) was edited using Elk.

Steps to Reproduce:

1. Create poll using native Mastodon website

   image

2. Edit post using Elk

   image

3. Edited post no longer contains poll

   image

### LINKED ISSUE #1769: Support creating polls
Creating a new issue for tracking, @wheatjs said in Discord he may be looking into implementing this feature.

JohnyTheCarrot:
A PR for the much requested poll creation feature! 🥳 
Since this modifies the status creation process, it's probably best to do rigorous testing for bugs.

https://github.com/elk-zone/elk/assets/24935987/253263f3-5d6b-4a78-9fd4-f1dff6ed748f

closes #1769
closes #1668

### DIFF
--- a/components/common/CommonCheckbox.vue
+++ b/components/common/CommonCheckbox.vue
@@ -1,10 +1,12 @@
 <script setup lang="ts">
 defineProps<{
-  label: string
+  label?: string
   hover?: boolean
+  iconChecked?: string
+  iconUnchecked?: string
 }>()
 const { modelValue } = defineModels<{
-  modelValue?: boolean
+  modelValue?: boolean | null
 }>()
 </script>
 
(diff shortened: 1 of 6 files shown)


patak-dev:
This is awesome! I like the UI for the poll options but I just played in mobile and only with icons it will be hard for people to understand. Maybe we should have the options as text plus icon checkboxes above the post toolbar.

JohnyTheCarrot:
Hmmm, I also don't know how this fairs with editing.

JohnyTheCarrot:
msedge_kiC5xBqTtV

@patak-dev how about this?

patak-dev:
Looks good, maybe the icons to the left to avoid the layout shift

JohnyTheCarrot:
> Looks good, maybe the icons to the left to avoid the layout shift

Unsure what you mean

patak-dev (approved):
I mean that we could have the icon on the left for these two CommonCheckbox, but the layout shift will still be there. We could have the width of the pop over to be bigger so it stay the same when the text change.

It seems the polls options isn't preserved in the draft (if you go to other screen and then back or reload). And I think maybe we should force the poll to have at least two options? 🤔  

But let's merge this and iterate! So we get some testing from folks in main.elk.zone for a bit.

PR #2113, opened 2023-05-21
chore: replace deprecated fetchV1Instance with V2

JohnyTheCarrot:
fetchV1Instance [is deprecated](https://docs.joinmastodon.org/methods/instance/#v1).
This PR replaces it with its V2 equivalent

### DIFF
--- a/composables/users.ts
+++ b/composables/users.ts
@@ -1,4 +1,3 @@
-import { withoutProtocol } from 'ufo'
 import type { mastodon } from 'masto'
 import type { EffectScope, Ref } from 'vue'
 import type { MaybeRefOrGetter, RemovableRef } from '@vueuse/core'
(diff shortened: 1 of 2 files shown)


danielroe (approved):

composables/users.ts
```diff
@@ -70,7 +68,7 @@ const publicInstance = ref<ElkInstance | null>(null)
 export const currentInstance = computed<null | ElkInstance>(() => currentUser.value ? instanceStorage.value[currentUser.value.server] ?? null : publicInstance.value)
 
 export function getInstanceDomain(instance: ElkInstance) {
-  return instance.accountDomain || withoutProtocol(instance.uri)
+  return instance.accountDomain ?? instance.domain
```
patak-dev, line 71: We use the stored instance from local storage before fetching. We probably need to support where we can reading from the old instance format from local storage. I just got an error when checking the PR the first time I open it, then it worked as the new config was cached. Maybe this is ok, but we should review there is no chance of messing with the user config. I think the safest is to still support `instance.uri` as a fallback here, and check other places too.

patak-dev:
https://docs.joinmastodon.org/methods/instance/#v2 is only available from Mastodon 4.0.0. I don't remember if we ever supported pre v4. @danielroe do you know if we did? Looks like we currently use `v2` only in some features (search, suggestions), I think most of Elk still works in Mastodon 3.x 🤔

I think we should document that Elk only support v4 and move on though, most servers should have already been updated, and we should try to reduce maintenance burden for the project. Someone could deploy an older version of Elk if that is important to them.

PR #2114, opened 2023-05-21
feat: adhere to and display poll limits

JohnyTheCarrot:
msedge_zi6xS9NFIE

### DIFF
--- a/components/publish/PublishWidget.vue
+++ b/components/publish/PublishWidget.vue
@@ -63,14 +63,25 @@ const { editor } = useTiptap({
   onPaste: handlePaste,
 })
 
+function trimPollOptions() {
+  const indexLastNonEmpty = draft.params.poll!.options.findLastIndex(option => option.trim().length > 0)
+  const trimmedOptions = draft.params.poll!.options.slice(0, indexLastNonEmpty + 1)
+
+  if (currentInstance.value?.configuration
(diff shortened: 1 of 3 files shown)


patak-dev:
Looks great! So this PR requires https://github.com/elk-zone/elk/pull/2113, no?

JohnyTheCarrot:
> Looks great! So this PR requires https://github.com/elk-zone/elk/pull/2113, no?

It doesn't require it. The fields here are the same.

components/publish/PublishWidget.vue
```diff
@@ -63,14 +63,25 @@ const { editor } = useTiptap({
...
+function trimPollOptions() {
+  const indexLastNonEmpty = draft.params.poll!.options.findLastIndex(option => option.trim().length > 0)
+  const trimmedOptions = draft.params.poll!.options.slice(0, indexLastNonEmpty + 1)
+
+  if (currentInstance.value?.configuration
+      && trimmedOptions.length >= currentInstance.value?.configuration?.polls.maxOptions)
+    draft.params.poll!.options = trimmedOptions
+  else
+    draft.params.poll!.options = [...trimmedOptions, '']
+}
```
patak-dev, line 75: I think this is ok for now, but maybe later we could leave the extra options with the `5/4` in red so the user can decide what options to trim? It should be rare to have this issue though.

patak-dev (approved):

PR #2005, opened 2023-04-23
feat(i18n): Add Italian Localisation

katullo11:
supersedes #1891
supersedes #1892

PR #2116, opened 2023-05-21
feat(i18n): Add Vietnamese locale

mastoduy:
Hope this will help!

### DIFF
--- a/config/i18n.ts
+++ b/config/i18n.ts
@@ -213,6 +213,11 @@ const locales: LocaleObjectData[] = [
     file: 'tl-PH.json',
     name: 'Tagalog',
   },
+  {
+    code: 'vi-VN',
+    file: 'vi-VN.json',
+    name: 'Vietnamese',
+  },
 ]
 
 function buildLocales() {
--- a/config/i18n.ts
+++ b/config/i18n.ts
@@ -216,7 +216,7 @@ const locales: LocaleObjectData[] = [
   {
     code: 'vi-VN',
     file: 'vi-VN.json',
-    name: 'Vietnamese',
+    name: 'Tiếng Việt',
   },
 ]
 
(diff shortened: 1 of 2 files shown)


mastoduy:
> Thanks! You also need to add the locale to `config/i18n.ts` (see https://github.com/elk-zone/elk/pull/2005/files). And there are lint errors, which you can fix with `pnpm run lint --fix`

I updated i18n.ts and failed check, too

patak-dev:
We'll take care of this once we get a chance, don't worry. Thanks!

userquin:
> > Thanks! You also need to add the locale to `config/i18n.ts` (see https://github.com/elk-zone/elk/pull/2005/files). And there are lint errors, which you can fix with `pnpm run lint --fix`
> 
> I updated i18n.ts and failed check, too. It was a little bit hard for a translator like me.

Why do you sent a new PR to update the i19n module? I fixed it.

#2119 should be closed

config/i18n.ts
```diff
@@ -213,6 +213,11 @@ const locales: LocaleObjectData[] = [
     file: 'tl-PH.json',
     name: 'Tagalog',
   },
+  {
+    code: 'vi-VN',
+    file: 'vi-VN.json',
+    name: 'Vietnamese',
```
userquin, line 219: shouldn't `Vitenamese` also be translated here?

userquin: @mastoduy `Tiếng Việt`

userquin: update to use first one:

imagen

patak-dev:
Thanks again for the translation @mastoduy! And @userquin for preparing things to be merged!

mastoduy:
Many thanks all of you for helping. Love Elk Team!

PR #2122, opened 2023-05-22
feat: add "Dim" colorMode

DataDrivenMD:
This PR adds a third color mode (and an acccompanying color theme) called "Dim". This is meant for users that are looking for a color mode that isn't as bright as the default "light" mode, but doesn't have the high-contrast attributes of the default "dark" mode.

This PR also:
- [X] explicitly defines the color mode logic for first-time users
- [X] disables a function call that would regenerate the theme source code with every build. Even if this PR isn't merged, this should be disabled as it's a superflouous step AFAICT. It also overwrites theme customizations without warning during the build step, which could lead to data loss if the build step is invoke prior to diff stashing or staging.
- [X] includes tooltip localization for nearly all of the languages currently supported


patak-dev:
I think we should keep showing the current theme icon in the toggle button (as it works in main). Currently, it shows the next one after the PR.

About the new theme, maybe the normal text color could be dimmed a bit too.

@antfu what do you think? At first, my reaction was to keep only dark/light and make dim an option but it may be easier to have different themes as this PR proposes. The toggle may work a bit strange if more are added though.

antfu:
I think it makes sense to keep the toggle binary. For dim mode, we could have it in the setting to customize how dark mode be like. So when you enable dim mode, it toggles between light and dim mode. Later we could also include a black mode that increases the contrast.

antfu:
We could do something similar to GitHub:
https://github.com/settings/appearance

image

antfu:
I don't mind to keep it simple for now, just throwing some ideas :)

patak-dev:
@DataDrivenMD that sounds good to me 👍🏼 
One change, for the value name of `dim`, maybe we could call it `dark-dim` so we have an easy way, later on, to know if a theme is light or dark. Later we could have `dark-high-contrast`, `light-solarized`, etc.

PR #2125, opened 2023-05-25
feat: expand Mastodon links as preview cards

DataDrivenMD:
This PR repurposes uses the `card` property from a Mastodon-compatible `Status` object to generate an interactive, styled preview card.

For `Status` objects that contain only text, the result looks like this:

open_more_actions_menu

For `Status` objects that contain attachments, the result looks like this:

preview_with_attachments

When the body of the preview card is selected (e.g. mouse click, keyboard command), the original status is opened in a new tab. When the "More actions" menu (top right) is selected, the following options are presented:

expanded_mastodon_preview_card

Note that the option to view the post directly is not available. That's intetional as this implementation specifically avoids making additional API calls. In order to enable the `NuxtLink` element that would handle native navigation, we would need to first lookup the `Status` to ensure that it's been federated to the host server. For the same reason, the option to "Mention" the original author is not available.

ref: https://github.com/elk-zone/elk/pull/456
cc: @ayoayco @patak-dev

### DIFF
--- a/components/status/StatusActionsMore.vue
+++ b/components/status/StatusActionsMore.vue
@@ -5,6 +5,8 @@ const props = defineProps<{
   status: mastodon.v1.Status
   details?: boolean
   command?: boolean
+  hideFavoritedAndBoostedBy?: boolean
+  hideMentionAccount?: boolean
 }>()
 
 const emit = defineEmits<{
@@ -176,6 +178,7 @@ function showFavoritedAndBoostedBy() {
         </template>
 
         <CommonDropdownItem
+          v-if="!hideFavoritedAndBoostedBy"
           :text="$t('menu.show_favourited_and_boosted_by')"
           icon="i-ri:hearts-line"
           :command="command"
@@ -254,12 +257,14 @@ function showFavoritedAndBoostedBy() {
             />
           </template>
           <template v-else>
-            <CommonDropdownItem
-              :text="$t('menu.mention_account', [`@${status.account.acct}`])"
-              icon="i-ri:at-line"
-              :command="command"
-              @click="mentionUser(status.account)"
-            />
+            <template v-if="!hideMentionAccount ?? false">
+              <CommonDropdownItem
+                :text="$t('menu.mention_account', [`@${status.account.acct}`])"
+                icon="i-ri:at-line"
+                :command="command"
+                @click="mentionUser(status.account)"
+              />
+            </template>
           </template>
         </template>
       </div>
--- a/components/status/StatusActionsMore.vue
+++ b/components/status/StatusActionsMore.vue
@@ -5,6 +5,7 @@ const props = defineProps<{
   status: mastodon.v1.Status
   details?: boolean
   command?: boolean
+  hideCopyLinkToPost?: boolean
   hideFavoritedAndBoostedBy?: boolean
   hideMentionAccount?: boolean
 }>()
@@ -186,6 +187,7 @@ function showFavoritedAndBoostedBy() {
         />
 
         <CommonDropdownItem
+          v-if="!hideCopyLinkToPost"
           :text="$t('menu.copy_link_to_post')"
           icon="i-ri:link"
           :command="command"
--- a/components/status/StatusContent.vue
+++ b/components/status/StatusContent.vue
@@ -9,6 +9,43 @@ const { status, context } = defineProps<{
   inNotification?: boolean
 }>()
 
+const linkToStatus = $computed(() => {
+  if (!status)
+    return undefined
+  const uriMatches = status.content.matchAll(/(?:href=\"(https[^"]+)\" rel)/mgi) // (https:\/\/[a-z0-9]+(?:[-a-z0-9()@:%_\+.~#?&\/=]+)+)/gi)
+  for (const mm of uriMatches) {
+    if (mm[0] !== null) {
+      // see https://github.com/mastodon/mastodon/blob/c1e70a20720741c33ac740242a8a7082fab23557/config/routes.rb#L128
+      // see also https://github.com/mastodon/mastodon/blob/c1e70a20720741c33ac740242a8a7082fab23557/app/models/account.rb#L65
+      const shortAccountStatus = mm[1].search(/[\/]@[a-z0-9_]+([a-z0-9_\.-]+[a-z0-9_]+)?[\/][0-9]+[\/]?$/gi)
+      if (shortAccountStatus !== -1) {
+        try {
+          const url = new URL(mm[1].replace(/[\/]$/gi, ''))
+          return url
+        }
+        catch (e) {
+          return undefined
+        }
+      }
+
+      // see https://github.com/mastodon/mastodon/blob/c1e70a20720741c33ac740242a8a7082fab23557/config/routes.rb#L89
+      const longAccountStatus = mm[1].search(/[\/]users[\/][a-z0-9_]+([a-z0-9_\.-]+[a-z0-9_]+)?[\/]statuses[\/][0-9]+[\/]?$/gi)
+      if (longAccountStatus !== -1) {
+        try {
+          const url = new URL(mm[1].replace(/[\/]$/gi, ''))
+          const p = url.pathname.replaceAll(/([\/]users[\/])|([\/]statuses)/gi, '')
+          url.pathname = `/@${p}`
+          return url
+        }
+        catch (e) {
+          return undefined
+        }
+      }
+      return undefined
+    }
+  }
+})
+
 const isDM = $computed(() => status.visibility === 'direct')
 const isDetails = $computed(() => context === 'details')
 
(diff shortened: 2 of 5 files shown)


components/status/StatusContent.vue
```diff
@@ -34,6 +71,7 @@ const hideAllMedia = computed(
 <template>
   <div
     space-y-3
+    my-4
```
patak-dev, line 74: I think this is having an unintended effect on every post
image

components/status/StatusPreviewMastodon.vue
```diff
@@ -0,0 +1,137 @@
+<script setup lang="ts">
```
patak-dev, line 1: running `pnpm run lint --fix` should format the files. I think it should have done it automatically 🤔

components/status/StatusPreviewMastodon.vue
```diff
@@ -0,0 +1,137 @@
...
+          <StatusBody v-if="derivedStatus" :status="derivedStatus" :with-action="false" class="font-light" />
+        </NuxtLink>
+      </div>
+      <!-- Attachments -->
+      <div v-if="card.image" space-y-3 my-2>
+        <StatusMedia
+          v-if="derivedStatus.mediaAttachments?.length > 0"
+          :status="derivedStatus"
+          :is-preview="true"
+        />
```
patak-dev, line 128: I saw a jump on [this one](https://deploy-preview-2125--elk-zone.netlify.app/universeodon.com/@atomicpoet@calckey.social/110421838157818549), the space isn't reserved until the image is loaded. @ayoayco said we have the width/height of the image: https://github.com/elk-zone/elk/pull/456#issuecomment-1374027756, so it looks like we could avoid this layout shift.

components/status/StatusPreviewCard.vue
```diff
@@ -2,20 +2,23 @@
...
 
 const gitHubCards = $(usePreferences('experimentalGitHubCards'))
 </script>
 
 <template>
-  <LazyStatusPreviewGitHub v-if="gitHubCards && providerName === 'GitHub'" :card="card" />
-  <LazyStatusPreviewStackBlitz v-else-if="gitHubCards && providerName === 'StackBlitz'" :card="card" :small-picture-only="smallPictureOnly" :root="root" />
+  <StatusPreviewMastodon v-if="linkToStatus" :link-to-status="linkToStatus" :source-status="status" :card="card" />
+  <LazyStatusPreviewGitHub v-else-if="gitHubCards && providerName.toLowerCase().includes('github')" :card="card" />
+  <LazyStatusPreviewStackBlitz v-else-if="gitHubCards && providerName.toLowerCase().includes('stackblitz')" :card="card" :small-picture-only="smallPictureOnly" :root="root" />
```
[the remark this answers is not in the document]
DataDrivenMD: I didn't make any change with regard to lazy vs. strict. The changes I made were to: 
- [X] Add the `StatusPreviewMastodon` component
- [X] clean up the business logic to make sure that we're capturing spelling variations or other changes that happen at the source. For example, if `Github` changes their provider name to `GitHub` or `github`, the existing logic would fail.

components/status/StatusPreviewMastodon.vue
DataDrivenMD: I can hard-code the image tag like @ayoayco had initially proposed. The drawback is that we lose the interactivity (e.g. click to open lightbox) that is currently enabled by running it through the `StatusMedia` pipeline.

FWIW, it's actually pretty hard to find posts with links to other posts. It's rarer, still, to find posts with links to other posts that include media attachments. When scrolling through a timeline, the jump is virtually imperceptible b/c the loading nearly always happens before the status makes it into the view. Then, once it's loaded the first time, going from timeline -> detail view is seamless* (*on my computer).

^^ That's my pitch for keeping it as-is, but I'm happy to hard-code the attachment if you prefer.

patak-dev: I like the idea of going through the `StatusMedia` pipeline, but you could pass the width and height as optional props to the component to avoid the jump, no?

components/status/StatusPreviewCard.vue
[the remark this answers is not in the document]
DataDrivenMD: Ah, gotcha. I thought you were referring to the Lazy loading. I'll make the changes.

components/status/StatusPreviewMastodon.vue
DataDrivenMD: AFAICT we have 2 options: we can either tell Elk to treat the attachment as a full-sized image/video by settting `fullSize === true` *or* we can mock the attachment's meta tags programmatically:

code

For some reason, I forgot to set `fullSize` so, I'll do that before resubmitting.

FTR: I initially went down the route of mocking the meta tags, but changed my mind once I realized that I would need to create custom mocking logic for each attachment type:

masto_media_meta


That's *a lot* of code to write *just* to avoid making a single Masto API call to fetch the original status, which would get us the actual attachments in addition to all the other attributes that are currently being mocked to make this work without additional API calls.

components/status/StatusPreviewMastodon.vue
DataDrivenMD: It's running that command every time I commit, but it didn't pick this one up. Weird. Anywho, I manually deleted the superfluous variable assignment.

components/status/StatusPreviewMastodon.vue
DataDrivenMD: I suppose another option would be to add a div element with  `flex flex-initial basis-[Xpx]` where X === `card.height` but that requires a level of tinkering with `flex` attributes that I'm not prepared to do ATM lol

The jumping bothers me, too, even if it's not always noticeable. I'd like to get these changes done, and revisit in a week or two once my brain is a little less fried than it currently is -- frontend dev work is not my forté lol

components/status/StatusContent.vue
DataDrivenMD: Oops! I'll revert the change.

components/status/StatusActionsMore.vue
```diff
@@ -5,6 +5,8 @@ const props = defineProps<{
   status: mastodon.v1.Status
   details?: boolean
   command?: boolean
```
DataDrivenMD, line 7: Adding this because I just noticed that it's possible to throw a fatal error when someone tries to manually fetch a status that has not federated to their instance.

DataDrivenMD: @patak-dev this PR is ready for re-review. This is only *new* change that I made, which is a bug fix to handle the scenario where the linked status has not federated to the instance where the post with the embedded link was posted. In this scenario, the dropdown previously included the option to "Copy link to this post" but when that link was subsequently entered into a browser's address bar, the user would hit an error page because the status `id` would be incorrect or missing altogether.

Here's an example of the fixed version using the preview environment: 
https://deploy-preview-2125--elk-zone.netlify.app/calckey.social/notes/9f6y642l66

DataDrivenMD:
@patak-dev I've been at this for over a week, and it seems to me that the cause of the skipping/flickering is rooted in the DOM tree itself. The use of 100dvh near the top of the DOM tree in combination with the calculated container height that wraps the slot that `<MainContent>` populates doesn't play nice with the 3+ layers of nested components within the main content slot that inherit `overscroll-y: auto`.

Regardless of what I do on my end to avoid skipping/shifting/flickering of preview cards with media attachments, it still happens. Using DevTools I found that every time the user scrolls vertically, the current DOM structure triggers multiple re-renders of the same `StatusCard` element a mere milliseconds apart. As these reactive elements are based on promises that are not consistently `await`ed, the mounted node reflects the first fulfilled promise, and as the subsequent promises are fulfilled, they are interpreted as updates to the node with the same/corresponding `:key`.

TL; DR: The DOM tree produced by the current layout yields the UX equivalent of a race condition. I'm pretty close to figuring out how to keep the re-rendering to a minimum, and will circle back with a separate PR to address that issue. I'll leave this PR as-is for now, as there's nothing that I can add to it that would resolve the outstanding issue.

components/status/StatusPreviewMastodon.vue
```diff
@@ -0,0 +1,137 @@
...
+            </div>
+          </div>
+        </div>
+        <div flex-auto />
+        <StatusActionsMore v-if="derivedStatus" :status="derivedStatus" :details="false" :command="false" :hide-copy-link-to-post="true" :hide-favorited-and-boosted-by="true" :hide-mention-account="true" me--2 />
+      </div>
+      <!-- Content -->
+      <div space-y-3 my-2>
+        <NuxtLink v-if="derivedStatus.url" :to="derivedStatus.url" external target="_blank">
+          <StatusBody v-if="derivedStatus" :status="derivedStatus" :with-action="false" class="font-light" />
```
bleonard252, line 118: I'll be honest, `font-light` looks weird here.
image

shuuji3:
It's been a while but Mastodon finally supported the quote feature and Elk implemented (see #387) 🙂

PR #2160, opened 2023-06-11
feat: greater privacy until user is logged in

DataDrivenMD:
Community (open) Elk instances can be source of data leakage to bad actors and to web scrapers who seek to circumvent Mastodon's discoverability settings.

This PR seeks to mitigate the risk of inadvertent and/or unauthorized disclosure by requiring users to log in before:
- [X] running a search
- [X] navigating to account profiles
- [X] navigating to status details

This PR preserves the existing permalink structure. If a user knows the path to a public post or profile, then they will be served the corresponding page. If, however, the user seeks to navigate from that page to another post or profile, they will be prompted to log-in:

- [X] On public posts, unauthenticated users will no longer be able to see who interacted with the post, nor will they be able to copy the permalink. This will make unauthorized crawling of social graphs more challenging
- [X] On public profile pages, unauthenticated users will no longer see follower, following, or total posts (nor be able to explore relationships)

### DIFF
--- a/components/account/AccountHoverCard.vue
+++ b/components/account/AccountHoverCard.vue
@@ -8,8 +8,8 @@ const { account } = defineProps<{
 const relationship = $(useRelationship(account))
 </script>
 
-<template>
-  <div v-show="relationship" flex="~ col gap2" rounded min-w-90 max-w-120 z-100 overflow-hidden p-4>
+<template v-if="currentUser && account">
+  <div v-if="relationship" flex="~ col gap2" rounded min-w-90 max-w-120 z-100 overflow-hidden p-4>
     <div flex="~ gap2" items-center>
       <NuxtLink :to="getAccountRoute(account)" flex-auto rounded-full hover:bg-active transition-100 pe5 me-a>
         <AccountInfo :account="account" />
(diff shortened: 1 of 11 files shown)


DataDrivenMD:
@danielroe I use [Puppeteer](https://github.com/puppeteer/puppeteer) to crawl SPAs, and I've detected programmatic request patterns in my Mastodon server logs so it's definitely possible and may already be happening.

While it's true that the default Mastodon configuration translate to wide-open CORS settings for the API endpoints, many instance admins (myself included) harden the default CORS configuration in order to accommodate higher rate limits from trusted third-party clients.

All that being said, I totally understand if the cost/benefit analysis of merging this PR to Elk core doesn't pan out. I can leave the PR open to serve as a point-of-reference for other instance admins, or close it to minimize clutter in the PR queue clutter. No worries either way.

patak-dev:
I'm with @danielroe here. If you go to a Mastodon server: https://fosstodon.org/, you can do all the operations that would be restricted after this PR without being logged in. So I don't see a big difference with what Elk is doing. Also, sending an elk.zone link to a profile or a status is an important feature. You can share these links in a forum and you know people are going to be able to see the content without yet having an account (and they may create one after seeing that there is good content in Mastodon)

DataDrivenMD:
@danielroe @patak-dev I hear your concerns, and will close the PR with this final comment. I do want to state FTR that this doesn't change at all with this PR:

> Also, sending an elk.zone link to a profile or a status is an important feature. You can share these links in a forum and you know people are going to be able to see the content without yet having an account (and they may create one after seeing that there is good content in Mastodon)

What does change is that, from that profile or status, the user (or bot as the case may be) is prompted to log in if they want to interact further with features and/or pages that could include `noindex` meta tags if visited via the official Mastodon web client.

I'll keep the branch around for awhile just in case other instance admins want to apply this to their fork.

Issue #1522, opened 2023-01-29
CORS issue on fetching `/nodeinfo/2.0`


jibsaramnim:
One of the earlier automatic requests made to the server you're looking at through Elk fails due to its CORS policy. I noticed this with my locally running version, but can confirm this is also the case on `main.elk.zone`.

```log
Cross-Origin Request Blocked: The Same Origin Policy disallows reading the remote resource at https://m.webtoo.ls/nodeinfo/2.0. (Reason: CORS header ‘Access-Control-Allow-Origin’ missing). Status code: 200.
```

The endpoint seems mostly just informative regarding what the selected server is running in terms of version and whatnot, but as Elk does try to read from it, I thought I'd report it here.

Shinigami92:
I can confirm this, but just didn't had time to open an issue for this 👍

ayoayco:
@Shinigami92 did you see a JS error being thrown due to this?

Shinigami92:
I had an issue in the browser console. Need to check again when I'm back to a PC

Shinigami92:
@ayoayco 

![image](https://user-images.githubusercontent.com/7195563/227205832-d71d64b5-2865-4c51-a83c-c512ab426e3c.png)

ayoayco:
Thanks!! For additional context, this is the endpoint "used for advertising software name and version, protocols, usage statistics, and whether registrations are open." https://docs.joinmastodon.org/dev/routes/#nodeinfo

sxzz:
See https://github.com/mastodon/mastodon/issues/23135

Shinigami92:
https://github.com/mastodon/mastodon/commit/23f8e93c6405277515d018d4ff2d550fe64ac159 was merged
I assume it will land in Mastodon v4.2.2

I will close this PR as done

PR #2161, opened 2023-06-11
fix: CORS issue

DataDrivenMD:
Closes: https://github.com/elk-zone/elk/issues/1522

Notes for whomever reviews this PR:
- Mastodon API-compatible instances remap *but do not redirect* `/nodeinfo/2.0` to `/.well-known/nodeinfo` ([see Mastodon documentation](https://docs.joinmastodon.org/dev/routes/#nodeinfo)). The `/.well-known/*` routes are wide open by default ([see Mastodon repo](https://github.com/mastodon/mastodon/blob/dfaf59d99a2aa1f913608faee978bc3e1586a3ce/config/initializers/cors.rb#L12)) and should be sufficient to fix this issue for Mastodon instances. That said, in the wilde I've encountered a wide range customizations that would cause this fix to fail as a standalone approach (e.g. instance-wide rate-limiting, CORS-hardened configurations, CORS-stripped headers), which is why I included a fail-safe call to the same endpoint using the `no-cors` fetch option.
- The original code mentions GoToSocial, and I am not familiar with the differences in the API spec, so I essentially did the same thing as above, but with the `/nodeinfo/2.0` endpoint just in case.

### DIFF
--- a/composables/users.ts
+++ b/composables/users.ts
@@ -120,15 +120,47 @@ export function useSelfAccount(user: MaybeRefOrGetter<mastodon.v1.Account | unde
 
 export const characterLimit = computed(() => currentInstance.value?.configuration?.statuses.maxCharacters ?? DEFAULT_POST_CHARS_LIMIT)
 
+async function fetchMastodonNodeinfo(url: string, mode: RequestMode = 'cors') {
+  return fetch(`${url}/.well-known/nodeinfo`, { mode })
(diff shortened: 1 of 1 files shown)


Shinigami92:
@DataDrivenMD What does all this cloudflare stuff to do with the cors nodeinfo fix?
Do you maybe need to rebase this PR?

Issue #1794, opened 2023-02-18
acct is wrongly including subdomain for certain servers (like m.webtoo.ls)


patak-dev:
@userquin @danielroe, I'm creating an issue to dump what I see so far as I'm a bit confused. I thought this would be an easy regression but it looks like the code evolved quite a bit with this bug present. It is causing many issues in the app currently, and it looks like it may be a regression introduced in several steps in the past month (or even more).

For a server like `https://m.webtoo.ls`, acct handles should look like `patak@webtoo.ls` without the subdomain. This is because we configure the server to use the domain as the handle.

If you hover in the timeline avatars, everything works well (even for your own avatar). The `account.acct` will correctly be `patak@webtoo.ls` and clicking on it will lead to `/m.webtoo.ls/@patak`. 

But if you hover on the Account Switcher or Picker on the bottom right, then the `acct` will be `patak@m.webtoo.ls`. Even for other accounts or in other servers, the subdomain is also there. Clicking on it will lead you to `/m.webtoo.ls/@patak@m.webtoo.ls` because it can't remove the server and if you share this route, the social preview will not work. Switching users has other issues that I think are related to the same.

What is even more problematic is that the `acct` is also wrong in both local storage and in IndexedDB. 

For some reason, we are getting a good value for `acct` in the timeline. It looks like `status.account.acct` is correct. But the way we get the user account given an id may be wrong, maybe masto.js should give us a proper `acct` and it isn't and this is why we end up with code like [this](https://github.com/elk-zone/elk/blob/41175b02ca7163e37a95cc0cd302835d2685f276/composables/users.ts#L170) that is wrong because we should be ignoring the subdomain for the server (but I think we can't know at this point).

I'm a little bit puzzled here, I hope it is a regression. IIRC, all this was working at one point. But checking the code, we now have 5-6 utils where we are adding the `@{server}` at the end and all of these should be removed IIUC. Maybe I'm missing something? We may need to rework the local storage and IndexedDB once this is solved for everyone in servers like webtoo.ls.

userquin:
I Will try to review 2 modules and the usage of account extraction functions, It seems a call to the wrong function from sfc.

userquin:
The logic is on cache.ts and users.ts composables, I'll check history in cache.ts, it should remove the server suffix on sign-in from users.ts (IIRC the Last function to cache the account, there is a slice)

patak-dev:
[...]

Issue #1636, opened 2023-02-05
Respect Mastodon's LOCAL_DOMAIN and WEB_DOMAIN environment variables


pro-sumer:
Elk displays my account as `@username@mastodon.domain.com`(not what I want) in the account switcher in the bottom left corner, but as `@username@domain.com` (which I want) in my profile. This is probably because I have set `LOCAL_DOMAIN` to `domain.com` and `WEB_DOMAIN` to `mastodon.domain.com`:

https://docs.joinmastodon.org/admin/config/#local_domain
https://docs.joinmastodon.org/admin/config/#web_domain

It would be great if Elk respects these settings and displays my account as `@rob@prosumer.dev` everywhere.

Issue #888, opened 2023-01-08
Link copied from address bar is not working for people hosted in a different mastodon instance


edimitchel:
Someone shared me the following link: 
`https://elk.zone/m.webtoo.ls/@feditips@mstdn.social`

For context, my account is hosted on `mamot.fr`, which is different than `webtoo.ls` used for getting that link.

If the exact link is used in incognito mode, it works.
If the link is `https://elk.zone/mstdn.social/@feditips`, it works.

So I think there is a small issue on how we handle link from another host.

PR #2167, opened 2023-06-13
fix: account resolution, subdomains, routing logic

DataDrivenMD:
This PR closes several related issues by standardizing and cleaning up the Mastodon account resolution logic with a particular emphasis on centralizing the account ID and account URI parsing logic. It also handles edge cases that were previously missing (e.g. attempting to lookup tags using the Mastodon API accounts lookup endpoint. The end result fixes the following open issues

- [X] [Exposing subdomain in handles](https://github.com/elk-zone/elk/issues/1794)
- [X] [Permalink failure due to errors during remote account lookup and resolution](https://github.com/elk-zone/elk/issues/888)

Closes: https://github.com/elk-zone/elk/issues/888
Closes: https://github.com/elk-zone/elk/issues/1636
Closes: https://github.com/elk-zone/elk/issues/1794

### DIFF
--- a/components/account/AccountHoverWrapper.vue
+++ b/components/account/AccountHoverWrapper.vue
@@ -7,11 +7,16 @@ const props = defineProps<{
   disabled?: boolean
 }>()
 
-const account = computed(() => props.account || (props.handle ? useAccountByHandle(props.handle!) : undefined))
+const account = ref<mastodon.v1.Account | null>(null)
+
 const userSettings = useUserSettings()
 
 defineOptions({
   inheritAttrs: false,
+})
+
+onMounted(async () => {
+  account.value = props.account || await fetchAccountByHandle(props.handle)
 })
 </script>
 
--- a/components/account/AccountMoreButton.vue
+++ b/components/account/AccountMoreButton.vue
@@ -18,6 +18,7 @@ const { t } = useI18n()
 const { client } = $(useMasto())
 const useStarFavoriteIcon = usePreferences('useStarFavoriteIcon')
 
+const accountServerName = $computed(() => getServerName(account))
 async function toggleMute() {
   if (!relationship!.muting && await openConfirmDialog({
     title: t('confirm.mute_account.title', [account.acct]),
@@ -48,14 +49,14 @@ async function toggleBlockUser() {
 
 async function toggleBlockDomain() {
   if (!relationship!.domainBlocking && await openConfirmDialog({
-    title: t('confirm.block_domain.title', [getServerName(account)]),
+    title: t('confirm.block_domain.title', [accountServerName]),
     confirm: t('confirm.block_domain.confirm'),
     cancel: t('confirm.block_domain.cancel'),
   }) !== 'confirm')
     return
 
   relationship!.domainBlocking = !relationship!.domainBlocking
-  await client.v1.domainBlocks[relationship!.domainBlocking ? 'block' : 'unblock'](getServerName(account))
+  await client.v1.domainBlocks[relationship!.domainBlocking ? 'block' : 'unblock'](accountServerName)
 }
 
 async function toggleReblogs() {
@@ -176,17 +177,17 @@ async function removeUserNote() {
             @click="toggleBlockUser()"
           />
 
-          <template v-if="getServerName(account) !== currentServer">
+          <template v-if="accountServerName !== currentServer">
             <CommonDropdownItem
               v-if="!relationship?.domainBlocking"
-              :text="$t('menu.block_domain', [getServerName(account)])"
+              :text="$t('menu.block_domain', [accountServerName])"
               icon="i-ri:shut-down-line"
               :command="command"
               @click="toggleBlockDomain()"
             />
             <CommonDropdownItem
               v-else
-              :text="$t('menu.unblock_domain', [getServerName(account)])"
+              :text="$t('menu.unblock_domain', [accountServerName])"
               icon="i-ri:restart-line"
               :command="command"
               @click="toggleBlockDomain()"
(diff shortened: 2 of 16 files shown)


patak-dev:
For reference, both https://github.com/elk-zone/elk/issues/1636 and https://github.com/elk-zone/elk/issues/1794 should already be closed by:
- https://github.com/elk-zone/elk/pull/2088

This is a big change. I don't know how are we going to properly test it 👀

As far as I understand, the approach in #2088 was ok. I think you may not be getting the proper `acct` for a server like `m.webtoo.ls`, or I may not understand properly the new logic. For `m.webtoo.ls` we should have `patak@webtoo.ls` (using the WEB_DOMAIN) and I now see a lot of `patak@m.webtoo.ls` when adding some logs in the code. This may be the reason for some bugs I encountered. It is strange though because in the UI I properly see `patak@webtoo.ls` everywhere 
- Settings were not respected when changing users (the theme color always stays the same even if users have different colors). But I see m.webtoo.ls in the result of webfinger and I see that acct is overridden with this value.
- Logged out of my account and re-logged and there was an error, reloading showed the account was not there anymore. Then later I couldn't reproduce. Maybe something related to old settings?
- The url when clicking my user avatar to go to the profile is `/m.webtoo.ls/@patak@webtoo.ls` and it should be `/m.webtoo.ls/@patak`.

DataDrivenMD:
You're right in terms of the navigation behavior, and I can fix that but I don't think that (2088)[https://github.com/elk-zone/elk/pull/2088]. Also, it seems like there's been a regression since 2088 was merged and 1636 and 1794 are still outstanding-- this is what I see for your profile on (https://elk.zone/m.webtoo.ls/@patak@webtoo.ls)[https://elk.zone/m.webtoo.ls/@patak@webtoo.ls]

elk_dot_zone

This what I see on the local timeline for elk.zone, which redirects to (https://elk.zone/m.webtoo.ls/public/local)[https://elk.zone/m.webtoo.ls/public/local]

Screenshot 2023-06-13 at 2 24 48 PM

Big picture: at the core of many of these issues is a desperate need for a refactor of account and post resolution. For example: I counted 4 different definitions of "handle" including two definitions of "short handle" -- and that's not even counting the string interpolation versions that are sprinkled throughout the codebase. The reason you see `m.webtoo.ls` in some logging is because the timeline paginator code (not my the stuff I touched) does some side-channel account resolution by pulling from the `status.account` and the `status.mentions` fields, when they are populated. That particular part of the status rendering lifecycle is particularly error-prone because it assumes that there's a relationship between the logged-in user and the author of a post, which isn't true most of the time-- particularly when reblogs and replies are involved. That's where a lot of the hydration errors stem from, including a bug that's live right now: hover cards fail if the user hovers over a mention in a user profile or post that they do not already have a relationship with. 

In light of all of this, it honestly would be easiest if you gave me the specs you want/need for the account resolution logic, and I just write this PR to those specs. That being said, if now is not the time for a refactor of this magnitude, then that's OK, too. It's honestly fine by me either way.

This (+ the other PRs I created in the last 48 hrs) are my way of contributing back to the project on issues that I encounter as I encounter them. I have no expectation that you'll merge any of them, and they're meant to serve future reference. If closing the PR makes more sense at this time, I'm happy to do that and will keep the branch around indefinitely as a point-of-reference - just let me know either way

patak-dev:
> You're right in terms of the navigation behavior, and I can fix that. It seems like there's been a regression since 2088 was merged and 1636 and 1794 are still outstanding

I see the issues in your screenshots only when I'm not logged in, and only for accounts in `m.webtoo.ls`. For other servers (m.codeflow.club, mastodon.roe.dev) or when I'm logged in, I see handles correctly respecting the webdomain. So it seems these are diff issues from what I was trying to solve (maybe related). There is probably a smaller patch that could be applied for these particular bugs. Although as you say, a good refactor to clean up all the places where handles are processed as you propose would be the best long term.

> This (+ the other PRs I created in the last 48 hrs) are my way of contributing back to the project on issues that I encounter as I encounter them. I have no expectation that you'll merge any of them, and they're meant to serve future reference. If closing the PR makes more sense at this time, I'm happy to do that and will keep the branch around indefinitely as a point-of-reference - just let me know either way

These are great PRs by all accounts, thanks again for them! We should keep them open and let others chime in. I just don't have the bandwidth myself right now, but wanted to comment on this one as I worked on the issues you linked here. Hopefully, others will have the time to dig deeper here.

DataDrivenMD:
@patak-dev just for posterity's sake, I fixed the issue you described in a way that should be agnostic to personal preference. Regardless of whether you try hitting `/m.webtoo.ls/@patak@m.webtoo.ls` or `/m.webtoo.ls/@patak@webtoo.ls` or `/webtoo.ls/@patak@webtoo.ls` you should get to the correct profile. And, as it turns out, that particular issue you highlighted was directly related to the bug causing the hover cards to throw errors. And it was also related to another issue I hadn't yet noticed which surfaced the subdomain in handles rendered by account hover cards.

The issue I didn't fix because it's entirely unrelated AFAICT is the observation regarding color/theme preferences not being persisted when changing users. Best I can tell that particular preference isn't persisted on a per-user basis, so fixing that would be out-of-scope for this particular PR.

Anywho, I'll close this PR now but leave the branch around for others to reference.

patak-dev:
Thanks for continuing with the PR and leaving a better reference here @DataDrivenMD.

About the color/theme preference, it is set by user but I now see the same issue in elk.zone. Sorry about missreporting it. It seems that when a user doesn't have a theme selected (and has the default orange) then it breaks when switching to it from a user that has a theme selected. So this is unrelated to this PR as you say.

PR #2168, opened 2023-06-13
fix: repeatedly fetching posts, improve federation logic

DataDrivenMD:
The post/status caching logic is written in a way that triggers multiple fetch requests for the same post, which becomes a problem when rate-limits come into play as well as when there are network bottlenecks (such as those [illustrated here](https://github.com/elk-zone/elk/issues/1374)). The issue is further exacerbated when the post is authored by someone that has been blocked by the user (or vice versa), or when the remote host is down.

This PR improves fetching logic for posts by caching error response codes, and using that information to decide whether to retry fetch attempts. This PR also does a better job of handling requests for remote posts that:
- reference `status.uri` instead of local `status.id` (erroneously) or as a side effect of preview card rendering logic, and
- when permalinks request remoted posts that have not yet been federated to the host instance (or that were purged as part of regular maintenance)

### DIFF
--- a/composables/cache.ts
+++ b/composables/cache.ts
@@ -1,8 +1,16 @@
 import { LRUCache } from 'lru-cache'
 import type { mastodon } from 'masto'
 
+// expire in an hour
 const cache = new LRUCache<string, any>({
   max: 1000,
+  ttl: 3600000,
+  ttlAutopurge: true,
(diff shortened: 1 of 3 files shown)


ryanccn:
You have commits from #1882 here :p

Issue #1082, opened 2023-01-13
Web app flickers when scrolling


forestsystem:
This is an issue I've had both with regular Mastodon's web app and Elk's web app, so maybe it's a problem with my phone (iPhone SE 2022, less than a month old, iOS 16.2). There doesn't seem to be much logic for when this starts happening, but it seems to happen more often once I've been using the web app for a while. I would like it to stop flickering.

This is what it looks like:

https://user-images.githubusercontent.com/122533181/212343668-29eb98aa-955a-478b-8053-db82b5e48a66.mov

PR #2171, opened 2023-06-15
fix: scrolling and random layout shifts

DataDrivenMD:
This re-works the default layout to address random layout shifts during loading and routing. It also unwraps the nested containers within the MainContent component, which is responsible for the mid-flight updates that cause flickering while scrolling.

By reworking the layout in this way, we're able to preserve the timeline state as the user navigates into-and-out-of status details and as they go from one timeline to another. It also isolates the third column, which would enable features such as remote timelines or a notification drawer.

Closes: https://github.com/elk-zone/elk/issues/1082

### DIFF
--- a/components/drawer/DrawerContent.vue
+++ b/components/drawer/DrawerContent.vue
@@ -1,0 +1,17 @@
+<script setup lang="ts">
+const info = useBuildInfo()
+</script>
+
+<template>
+  <nav>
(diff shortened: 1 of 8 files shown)


Shinigami92:
There is one big issue right now I directly identified: when you open https://deploy-preview-2171--elk-zone.netlify.app/home and place your mouse cursor left of the nav, you cant scroll the timeline
Also the nav is a bit shifted to the right and does not visually align anymore with the elk logo / header (but this could be due to the `Preview deploy` container in lower right corner 🤔)

PR #2179, opened 2023-06-18
Allow a user to cancel a follow request

### LINKED ISSUE #2162: Can't cancel a follow request
If a remote account hasn't accepted or rejected your follow request yet, there's no way to cancel the request. The "Requested" button just disables itself on hover, and the menu next to it doesn't include a cancel option.

I've mostly run into this when trying to interact with platforms that don't federate well with each other - the current version of GoToSocial can send a follow request to Lemmy, for instance, but the response never makes it back. (That's totally separate from this issue, but an easy way to reproduce it.) The same problem used to happen between GTS and Bookwyrm, but once that was fixed, I had no way through Elk to cancel the request so I could re-send it after the servers updated. (I ended up using Pinafore.)

schultyy:
Resolves #2162

### DIFF
--- a/components/account/AccountFollowButton.vue
+++ b/components/account/AccountFollowButton.vue
@@ -15,7 +15,7 @@ const relationship = $computed(() => props.relationship || useRelationship(accou
 
 const { client } = $(useMasto())
 async function toggleFollow() {
-  if (relationship!.following) {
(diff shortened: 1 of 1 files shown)


sxzz:
Thanks for the PR, but we should also handle the text on follow button

PR #2184, opened 2023-06-22
feat: Report posts

aaga:
Adds a modal dialog flow for reporting posts. Mostly mimics the stock Mastodon web app implementation.

Note: this PR rolls in the changes from an existing open PR for menu UI: #1673. If this PR is merged, #1673 can be closed as well.

### DIFF
--- a/components/report/ReportModal.vue
+++ b/components/report/ReportModal.vue
@@ -13,27 +13,60 @@ const emit = defineEmits<{
 
 const { client } = useMasto()
 
-const step = ref('choose')
+const step = ref('selectCategory')
 const serverRules = ref((await client.value.v2.instance.fetch()).rules || [])
 const reportReason = ref('')
 const selectedRuleIds = ref([])
+const availableStatuses = ref(status ? [status] : [])
+const selectedStatusIds = ref(status ? [status.id] : [])
 const additionalComments = ref('')
 const forwardReport = ref(false)
 
 const dismissButton = ref<HTMLDivElement>()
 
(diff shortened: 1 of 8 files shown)


components/report/ReportModal.vue
```diff
@@ -0,0 +1,267 @@
...
+  account: mastodon.v1.Account
+  status?: mastodon.v1.Status
+}>()
+
+const emit = defineEmits<{
+  (event: 'close'): void
+}>()
+
+const { client } = useMasto()
+const { t } = useI18n()
```
patak-dev, line 15: t is unused here
```suggestion
```

patak-dev:
Looks good to me @aaga! And great to see Mozilla interested in Elk and in contributing.

I don't know how to properly test this one without reporting someone, so maybe we should merge so we get it deployed to https://main.elk.zone and let others use the reporting tool to check it out.

Something I think we could change before merging is making the dialog wider on desktop. The Next button was after the fold in some cases with the current width 
image

aaga:
Thanks for the feedback @patak-dev. I increased the max-width of the dialog a bit; ultimately the number and length of server rules will determine how far down the Next button is, but hopefully this new width helps make things feel a little less cramped.

And yes, merging for testing on [https://main.elk.zone](https://main.elk.zone) sounds good!

patak-dev (approved):
A lot better now! I'll approve it and leave some time for others to check it out. If we don't see other comments, we could merge tomorrow and test it for a bit on https://main.elk.zone.

Issue #2192, opened 2023-07-03
Routing to post URLs regression


patak-dev:
Example of URL working in elk.zone: https://elk.zone/mastodon.social/@TodePond/109749045181798484
Regression on main.elk.zone: https://main.elk.zone/mastodon.social/@TodePond/109749045181798484

danielroe:
I think this should now be resolved.

PR #2195, opened 2023-07-03
fix: downgrade nuxt to 3.5.3 due to routes regression

horvbalint:
Fix for #2192

I tracked the issue down to the PR https://github.com/elk-zone/elk/pull/2163.
The problem is introduced with the nuxt `3.6.0` version, `3.5.3` works just fine. In this PR I revert this dependency update, but I don't know if we were using anything new from 3.6.0 so the solution might not be this easy.


patak-dev:
As discussed, I don't think we should merge this, but it is good info to know it seems to be an issue after upgrading Nuxt.

danielroe:
I think this should now be resolved.

Issue #2193, opened 2023-07-03
mixed replies' content in thread view 


patak-dev:
Issue reported by @TodePond ([here](https://discord.com/channels/1044887051155292200/1125312801741738024/1125312801741738024)) and by Wakest [here](https://discord.com/channels/1044887051155292200/1046769315405365288/1114892269032591473)

> Some replies are missing text and showing the incorrect media
> 
> Compare the two screenshots. The first is what is displayed when looking at the toot as a reply of the start of the thread. The second is what is displayed after clicking on the reply.
> 
> Start of thread: 
> https://main.elk.zone/mastodon.social/@TodePond/109749045181798484
> 
> Reply in question: 
> https://main.elk.zone/m.webtoo.ls/@patak/109749150499911818

It can't be tested using the links because there is another regression, URL links are no longer working: https://github.com/elk-zone/elk/issues/2192

I tested by searching Tode and checking their pinned post, then scrolling to my answer. It can be reproduced without virtual scrolling. We thought the issue was fixed by @danielroe [here](https://github.com/elk-zone/elk/commit/209013af09e75a5515886d869b8559147b4adc3e). At least the above is a good way to reproduce it now.

PR #2196, opened 2023-07-03
fix: status card confusion

horvbalint:
Fix for #2193

Adding a `key` attribute to the StatusCard seems to fix the issue.


patak-dev (approved):
This seems to work. It looks like the `key-field` we are passing to `DynamicScroller` has no effect. See [here](https://github.com/Akryum/vue-virtual-scroller/pull/732#issuecomment-1282045091). It is always `id`. Let's merge it for now. cc @Akryum, in case you want to check it too. It is strange that this started failing after updating Vue or Nuxt so maybe it is a regression.

We also have other `DynamicScroller`, we may need an id there too if this is indeed required.

PR #2208, opened 2023-07-08
fix: Update command key in keyboard shortcuts modal

aaronransley:
Replaces incorrect reference to `cmd` key in `MagickeysKeyboardShortcuts` with proper platform key on non-macOS devices.

Old display (see _Actions -> Command mode_):

image

New display

image


patak-dev (approved):
Nice one! Thanks!

PR #2209, opened 2023-07-08
fix(ui): stop question mark propagation

### LINKED ISSUE #1910: Compose a post with a question mark prompts shortcut modal on publish
This is happening to me frequently.

## Description

When I compose a post that includes a question mark, i'm having to select the **Publish** button twice.
The first time I select it, the keyboard shortcut modal displays.
I have to dismiss the modal, then select the **Publish** button a second time to get the post to publish.

## Steps to reproduce

1. In the Compose window on a desktop, type a post with a question mark

   I'm not sure yet if the question mark has to be the last character or anywhere in the post.

1. Select the **Publish** button

   Keyboard shortcut modal displays (since the `?` is a shortcut to display it).

1. Dismiss modal
1. Select the **Publish** button again

   Post publishes.

## What should happen?

Post should publish the first time I select the **Publish** button without displaying the shortcut modal.

## System Info

- Elk version v0.7.5 · [c76e8e8](https://github.com/elk-zone/elk/commit/c76e8e8f5eb27d3f24e0359a532809adaeb463ac)
- Host: elk.vmst.io
- OS: Pop!_OS 22.04
- Browser: Firefox 111.0

userquin:
closes #1910

### DIFF
--- a/components/publish/PublishWidget.vue
+++ b/components/publish/PublishWidget.vue
@@ -200,6 +200,11 @@ defineExpose({
   },
 })
 
+function stopQuestionMarkPropagation(e: KeyboardEvent) {
+  if (e.key === '?')
+    e.stopImmediatePropagation()
+}
+
 onDeactivated(() => {
   clearEmptyDrafts()
 })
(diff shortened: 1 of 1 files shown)


johnsonm:
This has resolved the problem for me at least on Firefox 114 on Linux

patak-dev (approved):
I think we need a more general solution if we later have other shortcuts, but let's merge this for now as this bug is affecting a lot of users.

PR #2211, opened 2023-07-08
feat: view-transition-api

horvbalint:
In this PR I created a demo for the (very) new [View Transitions API](https://developer.mozilla.org/en-US/docs/Web/API/View_Transitions_API).

This API allows web applications to animate complex layout shifts with relative ease and achieve a very native like look. This PR is only a rough demo of the feature, it can be improved a lot if this is something that the core Elk maintaners and the community would like to see.

The API is currently only available in desktop and android Chrome, for other browsers everything should work like before.

### How does it look?
Currently only the change between the feed and a specific post is animated.

https://github.com/elk-zone/elk/assets/40771359/ed5d3f4f-2599-45fa-831c-81facdb1368d

### Hacks/Changes/Problems:
These hacks and problems can be eliminated with some more work (in theory).

- There is a 100ms delay when opening a post or when returning from it to the feed
- The forced refresh of the status cache when opening a status had to be removed
- The reduce motion user settings is currently ignored (very easy to implement)
- Z-indexes can cause trouble (posts can hover over the header-nav while being animated)
- There are some visual glitches/there is room for the improvement of the animation

More information on this API can be found here: https://developer.chrome.com/docs/web-platform/view-transitions/

### DIFF
--- a/components/account/AccountDisplayName.vue
+++ b/components/account/AccountDisplayName.vue
@@ -1,14 +1,25 @@
 <script setup lang="ts">
 import type { mastodon } from 'masto'
 
-const { account, hideEmojis = false } = defineProps<{
+const { account, hideEmojis = false, status } = defineProps<{
   account: mastodon.v1.Account
   hideEmojis?: boolean
+  status?: mastodon.v1.Status
 }>()
+
+const viewTransitionStyle = computed(() => {
+  if (!status)
+    return
+
+  const targets = getViewTransitionTargets().value
+  if (targets.statusId === status.id && targets.accountId === account.id)
+    return { 'view-transition-name': 'account-display-name' }
+})
 </script>
 
 <template>
   <ContentRich
+    :style="viewTransitionStyle"
     :content="getDisplayName(account, { rich: true })"
     :emojis="account.emojis"
     :hide-emojis="hideEmojis"
--- a/composables/viewTransition.ts
+++ b/composables/viewTransition.ts
@@ -1,0 +1,11 @@
+interface ViewTransitionTargets {
+  statusId: null | string
+  accountId: null | string
+}
+
+export function getViewTransitionTargets() {
+  return useState('viewTransitionTargets', () => ({
+    statusId: null,
+    accountId: null,
+  } as ViewTransitionTargets))
+}
--- a/components/account/AccountAvatar.vue
+++ b/components/account/AccountAvatar.vue
@@ -1,18 +1,29 @@
 <script setup lang="ts">
 import type { mastodon } from 'masto'
 
-defineProps<{
+const { account, status } = defineProps<{
   account: mastodon.v1.Account
   square?: boolean
+  status?: mastodon.v1.Status
 }>()
 
 const loaded = $ref(false)
 const error = $ref(false)
+
+const viewTransitionStyle = computed(() => {
+  if (!status)
+    return
+
+  const targets = getViewTransitionTargets().value
+  if (targets.statusId === status.id && targets.accountId === account.id)
+    return { 'view-transition-name': 'account-avatar' }
+})
 </script>
 
 <template>
   <img
     :key="account.avatar"
+    v-bind="$attrs"
     width="400"
     height="400"
     select-none
@@ -21,8 +32,7 @@ const error = $ref(false)
     loading="lazy"
     class="account-avatar"
     :class="(loaded ? 'bg-base' : 'bg-gray:10') + (square ? ' ' : ' rounded-full')"
-    :style="{ 'clip-path': square ? `url(#avatar-mask)` : 'none' }"
-    v-bind="$attrs"
+    :style="{ 'clip-path': square ? `url(#avatar-mask)` : 'none', ...viewTransitionStyle }"
     @load="loaded = true"
     @error="error = true"
   >
--- a/components/account/AccountBigAvatar.vue
+++ b/components/account/AccountBigAvatar.vue
@@ -6,12 +6,13 @@ import type { mastodon } from 'masto'
 
 defineProps<{
   account: mastodon.v1.Account
+  status?: mastodon.v1.Status
   square?: boolean
 }>()
 </script>
 
 <template>
   <div :key="account.avatar" v-bind="$attrs" :style="{ 'clip-path': square ? `url(#avatar-mask)` : 'none' }" :class="{ 'rounded-full': !square }" bg-base w-54px h-54px flex items-center justify-center>
-    <AccountAvatar :account="account" w-48px h-48px :square="square" />
+    <AccountAvatar :account="account" :status="status" w-48px h-48px :square="square" />
   </div>
 </template>
--- a/components/account/AccountHandle.vue
+++ b/components/account/AccountHandle.vue
@@ -1,15 +1,24 @@
 <script setup lang="ts">
 import type { mastodon } from 'masto'
 
-const { account } = defineProps<{
+const { account, status } = defineProps<{
   account: mastodon.v1.Account
+  status?: mastodon.v1.Status
 }>()
-
 const serverName = $computed(() => getServerName(account))
+
+const viewTransitionStyle = computed(() => {
+  if (!status)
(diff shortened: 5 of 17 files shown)


danielroe:
We have support for this built in to Nuxt, which can be enabled with:

```ts
export default defineNuxtConfig({
  experimental: {
    viewTransition: true
  }
})
```

Of course, it still needs to be used with CSS.

horvbalint:
@danielroe Wow, that is awesome! I will look into it. That should eliminate the 100ms hack I think 😄

patak-dev:
This looks amazing @horvbalint!

composables/viewTransition.ts
```diff
@@ -0,0 +1,11 @@
+interface ViewTransitionTargets {
```
userquin, line 1: use kebap-case: `view-transition.ts`

userquin:
It would be nice if also adding the transition when clicking in the toot account.

horvbalint:
That would be great indeed 😍 I will work on this in the upcoming days, hopefully it will take shape slowly.

horvbalint:
I excluded the account-hover-card from the transition and that apparently solved the occasional animation error 🎉
I will now work on the 'reduce motion' setting and maybe tweak the animation a bit. Then it is ready to be reviewd I think 🤔

danielroe:
sorry for the long delay in reviewing 🙏

horvbalint:
No problem😄 I converted it back into draft today, since I found some bugs + I want to add a user preference for the animation now that elk is using nuxt 3.10.

horvbalint:
@danielroe This is ready for review now 🥳
I added an entry in the experimental section of the preferences, so to try it out one have to enable it (it defaults to false). I think this is needed, because on phones and lower end devices the animation might be too heavy. I have no test for this, but collecting some feedback during experimental would be worth it imo.

There is one change I am not sure is ok, that is the only unchecked list item in the original PR comment.

.vscode/settings.json
```diff
@@ -6,7 +6,7 @@
     "unstorage"
   ],
   "editor.codeActionsOnSave": {
-    "source.fixAll.eslint": true
+    "source.fixAll.eslint": "explicit"
```
userquin, line 9: revert this change

pnpm-lock.yaml
```diff
@@ -1,4 +1,4 @@
-lockfileVersion: '6.0'
+lockfileVersion: '6.1'
```
userquin, line 1: what is this? Have you updated you local pnpm via corepack?

userquin:
cannot see any transition on PR preview on my local...

userquin:
@horvbalint can you run lint script on your local? it seems you need to fix the dts

pnpm-lock.yaml
horvbalint: Oops, this was not intentional, I will revert this and the settings.json :)

userquin:
can you update your main branch and merge changes in view-transition-api branch?

horvbalint:
> cannot see any transition on PR preview on my local...

They are off by default, you can enable them in sttings/preferences/experimental. + the transition is quite choppy in dev mode, but smooth in production builds.

userquin:
nice, the account transition also included (awesome)

horvbalint:
I still have some plans to fix/improve things, like the transition to a status page where there are many replies have a big stutter currently when the replies are rendered. 
I think that should be fixable by waiting for the transition to end before rendering the replies, but this PR is getting somewhat big, so maybe those are better in a subsequent PR. (this is why I think this should be in experimental settings for now)

userquin:
forgot the ci error, will take 6 hours to finish, some vitest worker cannot be terminated, there is a pending PR to fix it

userquin:
@horvbalint https://github.com/elk-zone/elk/pull/2692

horvbalint:
Thanks @userquin 🎉 I will work on this in the upcoming days.

components/account/AccountHeader.vue
```diff
@@ -10,7 +10,9 @@ const { client } = useMasto()
 
 const { t } = useI18n()
 
-const createdAt = useFormattedDateTime(() => account.createdAt, {
+provide(viewTransitionAccountInjectionKey, account)
+
+const createdAt = useFormattedDateTime(() => account.creaedAt, { // TODO: typo
```
userquin, line 15: fix this @horvbalint

composables/view-transition.ts
```diff
@@ -1,4 +1,5 @@
 import type { mastodon } from 'masto'
+import type { ViewTransition } from '#app/plugins/view-transitions.client'
```
userquin, line 2: cannot use `#imports`?

userquin: remove it, not being used :sweat_smile:

horvbalint:
Waiting for the transition to end before rendering the context resulted in a pretty big improvement 🥳
Before:

https://github.com/elk-zone/elk/assets/40771359/48463418-b4c5-471e-afa1-aae0d4dae7b5

After:

https://github.com/elk-zone/elk/assets/40771359/f1ef1c94-150b-4331-a396-a465743c80be

I also fixed the context loading skeletons, those were shifted out of the view by a big margin from the status.

composables/view-transition.ts
horvbalint: Oooops :D At least they were used in a previous attempt :P

danielroe:
Beautiful 😍

horvbalint:
Maybe we should indicate it in 'Preferences' if the ViewTransition API is not supported by the browser? (It is only available in chromium now)

horvbalint:
I am not sure how to signal not available functionality the proper way. In the last commit I made the settings row disabled when ViewTransitions are not supported by the browser, is this enough? Won't it be confusing why it does nothing? Or maybe the row should be hidden altogether? Should we put an explainatory text somewhere?
If the latter, then I would love some design guidence, beacuse I am not good with design :P

shuuji3:
I think disabling the row looks good. In addition to that, how about adding a help message to the disabled row, for example, "This browser currently does not support [View Transition API](https://caniuse.com/view-transitions)"? So that users can understand why this setting is not available on their browsers.

locales/en.json
```diff
@@ -552,6 +552,8 @@
       "use_star_favorite_icon": "Use star favorite icon",
       "user_picker": "User Picker",
       "user_picker_description": "Displays all avatars of logged accounts in the bottom-left so you can switch quickly between them.",
+      "view_transitions": "View Transitions",
+      "view_transitions_description": "Navigate pages while transitioning the common elements between them.",
```
userquin, line 556: This should be more user oriented, it's very technical, something like  `😍 Enable beutiful animations  when navigating between pages` (the emoji shouldn't be included) would be better

userquin: /cc @patak-dev any suggestion here?

userquin:
On desktop there is some flickering if the toot is outside the destination viewport...

locales/en.json
patak-dev: I think maybe we should just say:
```suggestion
      "view_transitions_description": "Animate transitions.",
```

components/account/AccountDisplayName.vue
```diff
@@ -5,10 +5,13 @@ const { account, hideEmojis = false } = defineProps<{
   account: mastodon.v1.Account
   hideEmojis?: boolean
 }>()
+
+const viewTransitionStype = getViewTransitionStyles('account-display-name', { account })
 </script>
 
 <template>
   <ContentRich
+    :style="viewTransitionStype"
```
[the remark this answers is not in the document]
horvbalint: Thanks!

horvbalint:
> On desktop there is some flickering if the toot is outside the destination viewport...
> 
> Go to elkdev notifications and click on this toot (it seems the scroll):
> 
> imagen

This happens, because the context of a status has to be fetched when opened, while the status itself is reused from the timeline. So they appear later and shift the layout. I am not really sure what to do about this, but will try to figure something out.

Issue #2147, opened 2023-06-05
Selected text selection is not when visible `experimental-web-platform-features` flag is set


tomayac:
![Screenshot 2023-06-05 at 14 08 18](https://github.com/elk-zone/elk/assets/145676/347a4e52-c991-40e5-9315-7af819a05050)

The CSS color variable `--c-bg-selection` is not defined. The reason is outlined in this blog post: https://www.azabani.com/2022/09/01/meet-the-css-highlight-pseudos.html#accessing-global-constants, including some workarounds.

shuuji3:
I think this was already fixed by #2222

PR #2222, opened 2023-07-10
fix: display background when text is selected

fEyebrow:
https://github.com/elk-zone/elk/issues/2147

When writing tweets or entering text into an input field, there is no feedback when selecting text using a keyboard shortcut.


patak-dev (approved):
Thanks!

PR #2226, opened 2023-07-12
fix: Show lock indicator with the correct state

Kowlin:
Currently on Canary there is a issue whereby the locked account indicator shows up when a account is marked as a bot. This PR resolves that issue by modifying the state to check if the account is locked.


patak-dev (approved):

patak-dev:
Thanks!

PR #2247, opened 2023-07-21
fix(a11y): make the bottom menu button focusable and navigable with a11y label

### LINKED ISSUE #1422: Accessibility: Unlabeled links

shuuji3:
fix #1422 

Hi, this is the follow-up PR of #1867

This changes the `<label>` to `<button>` so that users can focus on it and also can open the menu panel by Enter/Space key. I think this doesn't change anything visually.

I'm not sure why the original invisible `<input>` was needed for the click event (I guess it could be attached to `<label>` too?). I believe the same functionality is available now via the `<button>`'s click event so removed it.

**Before**
Screenshot 2023-07-21 at 16 47 27

**After**
Screenshot 2023-07-21 at 16 46 58

Ideally, we can improve the menu items by following the appropriate ARIA menu role coding guide (ARIA: menu role - Accessibility | MDN - https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/menu_role) so that those who use accessibility tool can operate the menu more easily. But it requires more detailed work so I'll leave it for future work.

### DIFF
--- a/components/nav/NavBottom.vue
+++ b/components/nav/NavBottom.vue
@@ -43,14 +43,15 @@ const { notifications } = useNotifications()
       </NuxtLink>
     </template>
     <NavBottomMoreMenu v-slot="{ toggleVisible, show }" v-model="moreMenuVisible" flex flex-row items-center place-content-center h-full flex-1 cursor-pointer>
-      <label
+      <button
         flex items-center place-content-center h-full flex-1 class="select-none"
         :class="show ? '!text-primary' : ''"
+        aria-label="More menu"
+        @click="toggleVisible"
       >
-        <input type="checkbox" z="-1" absolute inset-0 opacity-0 @click="toggleVisible">
         <span v-show="show" i-ri:close-fill />
         <span v-show="!show" i-ri:more-fill />
-      </label>
+      </button>
     </NavBottomMoreMenu>
   </nav>
 </template>
--- a/components/nav/NavBottom.vue
+++ b/components/nav/NavBottom.vue
@@ -49,8 +49,7 @@ const { notifications } = useNotifications()
         aria-label="More menu"
         @click="toggleVisible"
       >
-        <span v-show="show" i-ri:close-fill />
-        <span v-show="!show" i-ri:more-fill />
+        <span :class="show ? 'i-ri:close-fill' : 'i-ri:more-fill' />"
       </button>
     </NavBottomMoreMenu>
   </nav>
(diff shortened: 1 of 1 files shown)


components/nav/NavBottom.vue
```diff
@@ -43,14 +43,15 @@ const { notifications } = useNotifications()
...
-      <label
+      <button
         flex items-center place-content-center h-full flex-1 class="select-none"
         :class="show ? '!text-primary' : ''"
+        aria-label="More menu"
+        @click="toggleVisible"
       >
-        <input type="checkbox" z="-1" absolute inset-0 opacity-0 @click="toggleVisible">
         <span v-show="show" i-ri:close-fill />
         <span v-show="!show" i-ri:more-fill />
```
userquin, line 53: can you replace spans with `<span :class="show ? 'i-ri:close-fill' : 'i-ri:more-fill' />"`

shuuji3: Thanks, this is a good refactoring. Applied now.

userquin: remove trailing double quote

components/nav/NavBottom.vue
```diff
@@ -43,14 +43,14 @@ const { notifications } = useNotifications()
...
         flex items-center place-content-center h-full flex-1 class="select-none"
         :class="show ? '!text-primary' : ''"
+        aria-label="More menu"
+        @click="toggleVisible"
       >
-        <input type="checkbox" z="-1" absolute inset-0 opacity-0 @click="toggleVisible">
-        <span v-show="show" i-ri:close-fill />
-        <span v-show="!show" i-ri:more-fill />
-      </label>
+        <span :class="show ? 'i-ri:close-fill' : 'i-ri:more-fill' />"
```
userquin, line 52: `<span :class="show ? 'i-ri:close-fill' : 'i-ri:more-fill'" />`

The trailing `"`  is outside the span tag

userquin: My snippet was wrong, sorry man

shuuji3: Ah, that's why.😄 No problem, it was helpful and should be fixed now.

userquin (approved):

Issue #2039, opened 2023-04-30
The bottom bar buttons are offset upwards from their displayed position om Firefox Android


Cragsand:
Screen recording shows clicking on the buttons clicks through them, but clicking above them registers properly.

Directly after refreshing a page, the buttons register at their correct locations. It seems to get worse and worse offset the more I scroll on a page but I'm not sure. 

https://user-images.githubusercontent.com/49168045/235375913-1aafbb59-60b6-48c3-9bdf-f2d18604511e.mp4

Example sketched in orange where I have to click a bit above buttons.

![Screenshot_20230430_222520_Firefox](https://user-images.githubusercontent.com/49168045/235376052-4f16c118-ad87-4c1f-b1d5-5268b147a074.jpg)

Firefox Android version updated yesterday:
112.2.0 (Build #2015946691)

Disabling all extensions doesn't appear to make any difference.

Best regards
Keep up the great work!

zspotter:
This is my only issue with the web app :) Seems like the buttons are offset by the size of the browsers bottom toolbar, which hides when you scroll. Likely has to do with how Firefox mobile reports the window size differently than Chrome.

audunmb:
I got this in the newest version as well.

PR #2252, opened 2023-07-22
fix: Position NavBottom with "fixed" instead of "sticky"

zspotter:
This fixes a Firefox Android bug (#2039) in which the bottom bar is rendered offset from the actual button location.

Apparently there's a browser bug with bottom-hugging sticky elements. When the browser's nav bar hides, the sticky element looks like it's positioned correctly, but the pointer events pass through it.

It's the same as this issue: https://stackoverflow.com/questions/70893056/mozilla-android-addressbar-conflict-with-sticky-button-on-bottom

But replacing `position: sticky` with `position: fixed` is an easy fix.

These screenshots from the remote debugger show the issue as the browser nav bar hides!

### DIFF
--- a/layouts/default.vue
+++ b/layouts/default.vue
@@ -54,7 +54,7 @@ const isGrayscale = usePreferences('grayscaleMode')
         <div min-h="[calc(100vh-3.5rem)]" sm:min-h-screen>
           <slot />
         </div>
-        <div sticky left-0 right-0 bottom-0 z-10 bg-base pb="[env(safe-area-inset-bottom)]" transition="padding 20">
+        <div style="position: fixed" left-0 right-0 bottom-0 z-10 bg-base pb="[env(safe-area-inset-bottom)]" transition="padding 20">
           <CommonOfflineChecker v-if="isHydrated" />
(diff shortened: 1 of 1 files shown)


layouts/default.vue
```diff
@@ -54,7 +54,7 @@ const isGrayscale = usePreferences('grayscaleMode')
         <div min-h="[calc(100vh-3.5rem)]" sm:min-h-screen>
           <slot />
         </div>
-        <div sticky left-0 right-0 bottom-0 z-10 bg-base pb="[env(safe-area-inset-bottom)]" transition="padding 20">
+        <div style="position: fixed" left-0 right-0 bottom-0 z-10 bg-base pb="[env(safe-area-inset-bottom)]" transition="padding 20">
```
patak-dev, line 57: This should do the same, no?
```suggestion
        <div fixed left-0 right-0 bottom-0 z-10 bg-base pb="[env(safe-area-inset-bottom)]" transition="padding 20">
```

patak-dev:
It seems we tried this fix here:
- https://github.com/elk-zone/elk/pull/660

but we had to revert it, so there should be more work to do to fix this problem.

PR #2253, opened 2023-07-22
fix: prevent the lost of focus on compose icons

### LINKED ISSUE #1679: Keyboard accessibility/focus issues on desktop/desktop apps.
The keyboard navigation on the desktop is inconsistent between Firefox, Chrome, Edge, Safari, and the Edge and Chrome apps. Once the user tabs to the "Add images, video, audio" button, the focus is lost and if the user tries to SHIFT + TAB back, the focus is trapped in the "footer" links (About, Privacy, Mastodon, etc.).

The web interface for Elk when creating a post.

This may also be a separate issue more than likely but the focus order is out of order because of the trapping of focus in the footer.

shuuji3:
fix #1679

This was caused by autofocusing the tooltip during keyboard navigation.

A similar issue was reported in the original `floating-vue` repository (https://github.com/Akryum/floating-vue/issues/872), which is the component we use to implement the `CommonTooltip`, and the `no-auto-fucus` option was implemented as mitigation.

I found the same issue on the "created at" time label of the post.

## How to test
1. Click the compose textarea
2. Use the `Tab` key several times to focus on the compose buttons and "created at" time.
3. Confirm if the focus won't be lost so the application is navigable with the keyboard.

### DIFF
--- a/components/publish/PublishWidget.vue
+++ b/components/publish/PublishWidget.vue
@@ -378,26 +378,26 @@ onDeactivated(() => {
             </button>
           </PublishEmojiPicker>
 
-          <CommonTooltip v-if="draft.params.poll === undefined" placement="top" :content="$t('tooltip.add_media')">
+          <CommonTooltip v-if="draft.params.poll === undefined" placement="top" :content="$t('tooltip.add_media')" no-auto-focus>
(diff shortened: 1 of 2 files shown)


patak-dev (approved):
This was a big one. Thanks!

Issue #2056, opened 2023-05-03
Label moderators and owners


biloti:
Moderators and owners of instances should be recognized by a clear label in their profiles. This is important to users in need of actions concerning moderation.

PR #2265, opened 2023-07-25
feat: add roles into account header and info

iruldi:
## Add roles into account header and account info

Linked to issue: https://github.com/elk-zone/elk/issues/2056

Added `AccountRoleIndicator` which is based on `AccountBotIndicator` so it has the same design. I use the color and name of the Role object to display it on the profile.

Screenshot 2023-07-25 at 16 29 11

Screenshot 2023-07-25 at 16 29 21

### DIFF
--- a/components/account/AccountHeader.vue
+++ b/components/account/AccountHeader.vue
@@ -167,6 +167,7 @@ const personalNoteMaxLength = 2000
         <div flex="~ col gap1" pt2>
           <div flex gap2 items-center>
             <AccountDisplayName :account="account" font-bold sm:text-2xl text-xl />
+            <AccountRoleIndicator v-for="role in account.roles" :key="role.id" :role="role" />
             <AccountLockIndicator v-if="account.locked" show-label />
             <AccountBotIndicator v-if="account.bot" show-label />
           </div>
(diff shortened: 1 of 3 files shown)


EddiesTech:
Is it intentional that when there is no specific role, their is an empty box? I am assuming not. I've attached an image from Elk Canary.
@iruldi @patak-dev 
Screenshot_20230728-135006

patak-dev:
Maybe that user has a role that is an empty string? You could send a PR to guard against that. I don't see the empty box in  https://main.elk.zone/ for any of the users I follow.

EddiesTech:
Hmm, ok. I'll do a bit of testing and try and figure out why that's happened then.

patak-dev:
@iruldi, @EddiesTech is right, it seems there is an empty box when you hover over users, and also on the user switcher on the bottom left. PR welcome to fix this one if you have some time
image

iruldi:
@patak-dev Thanks, I'll do that soon!

iruldi:
Didn't know why it was showing up some times but now it should not be the case anymore and it's much better this way. 

PR: https://github.com/elk-zone/elk/pull/2333

PR #2266, opened 2023-07-25
feat: role labels wrap

iruldi:
## Feat role labels wrap

Sorry @patak-dev I've not seen you've merged https://github.com/elk-zone/elk/pull/2265 and I've already done some little improvement because I've seen that with a lot of labels it became really ugly. This PR should limit these edge cases.


patak-dev (approved):
No problem, thanks for the follow up!

PR #2277, opened 2023-07-28
Add basic user-agent to all mastodon-bound requests

untitaker:
Default UA "undici" is not very nice, and I blocked it on my instance to
block a certain botwave.

### DIFF
--- a/server/api/[server]/oauth/[origin].ts
+++ b/server/api/[server]/oauth/[origin].ts
@@ -1,5 +1,7 @@
 import { stringifyQuery } from 'ufo'
 
+import { APP_NAME } from '~/constants'
+
 export default defineEventHandler(async (event) => {
   let { server, origin } = getRouterParams(event)
   server = server.toLocaleLowerCase().trim()
@@ -24,6 +26,9 @@ export default defineEventHandler(async (event) => {
   try {
     const result: any = await $fetch(`https://${server}/oauth/token`, {
       method: 'POST',
+      headers: {
+        'user-agent': APP_NAME,
+      },
       body: {
         client_id: app.client_id,
         client_secret: app.client_secret,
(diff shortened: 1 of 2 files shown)


server/api/[server]/oauth/[origin].ts
```diff
@@ -24,6 +26,9 @@ export default defineEventHandler(async (event) => {
   try {
     const result: any = await $fetch(`https://${server}/oauth/token`, {
       method: 'POST',
+      headers: {
+        'user-agent': APP_NAME,
```
Comment (antfu) [inline server/api/[server]/oauth/[origin].ts:30]: I guess we should append the version as well, `Elk/0.x.x`. Could put in the constants module.

I am not sure if we should also include other UA for compatibility, otherwise it LGTM

Comment (untitaker) [reply server/api/[server]/oauth/[origin].ts:30]: what changes need to be done to the release process so that this constant gets updated?

Comment (antfu) [reply server/api/[server]/oauth/[origin].ts:30]: You can get the version here https://github.com/elk-zone/elk/blob/57814915d655acecb578718a99d602218e0c3fe7/config/env.ts#L4

Comment (untitaker) [reply server/api/[server]/oauth/[origin].ts:30]: ah okay, ty. i thought you were suggesting to add a new constant. will fix soon

Comment (untitaker) [reply server/api/[server]/oauth/[origin].ts:30]: > I am not sure if we should also include other UA for compatibility, otherwise it LGTM

I believe this should not be necessary, considering you got away with literally hte string `"undici"` so far

untitaker:
@antfu can you take another look?

shuuji3 (approved):
Thank you for updating the PR!🙂

I confirmed both fetch requests (`/api/v1/apps` and `/oauth/token`) now have the user-agent value `Elk/0.15.1` in headers on the Elk server side.

I haven't checked the received request on the server side (Mastodon) by myself but that should be fine. Have you already tested on your server?

untitaker:
yes, IIRC I tested it fully. but it's been a long time.

shuuji3:
OK, recent commits only modified user-agent string so it should work as expected. Let's see how it goes on Canary version (https://main.elk.zone) after merging.

PR #2280, opened 2023-07-29
feat(i18n): sync zh-CN with en

alexzhang1030:


alexzhang1030:
close it cause duplicated with #2248

PR #2281, opened 2023-07-29
fix: cannot fullscreen due to `Permission-Policy`

### LINKED ISSUE #1942: video fullscreen button is disabled
截屏2023-04-02 09 59 23

alexzhang1030:
closes #1942

### DIFF
--- a/nuxt.config.ts
+++ b/nuxt.config.ts
@@ -242,6 +242,9 @@ export default defineNuxtConfig({
         'style-src': ['\'self\'', '\'unsafe-inline\''],
         'upgrade-insecure-requests': true,
       },
+      permissionsPolicy: {
+        fullscreen: ['\'self\'', 'https:', 'http:'],
+      },
     },
     rateLimiter: false,
   },
--- a/components/status/StatusAttachment.vue
+++ b/components/status/StatusAttachment.vue
@@ -131,7 +131,7 @@ watch(shouldLoadAttachment, () => {
           :poster="videoThumbnail"
           muted
           loop
-          playsinline
+          :playsinline="isSmallScreen"
           :controls="shouldLoadAttachment"
           rounded-lg
           object-cover
--- a/components/status/StatusAttachment.vue
+++ b/components/status/StatusAttachment.vue
@@ -131,7 +131,7 @@ watch(shouldLoadAttachment, () => {
           :poster="videoThumbnail"
           muted
           loop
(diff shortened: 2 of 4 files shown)


patak-dev:
Do you see the fullscreen button enabled in the preview? I still get it disabled for a large screen https://deploy-preview-2281--elk-zone.netlify.app/universeodon.com/@CraigWeekend@mstdn.ca/110794193399419745

alexzhang1030:
@patak-dev After a considerable amount of debugging, I've finally discovered that this issue originates from the [Permission-Policy](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Permissions-Policy/fullscreen), and it only occurs in Chrome 😅.

Already revert the last changes. You can merge it now!

alexzhang1030:
And why does the preview version not change? 🤔

nuxt.config.ts
```diff
@@ -242,6 +242,9 @@ export default defineNuxtConfig({
         'style-src': ['\'self\'', '\'unsafe-inline\''],
         'upgrade-insecure-requests': true,
       },
+      permissionsPolicy: {
+        fullscreen: ['\'self\'', 'https:', 'http:'],
```
patak-dev, line 246: Does it need to be so permissive? I'll let others review this one in that case.

alexzhang1030: FYI：it's syncs with contentSecurityPolicy.media-src

patak-dev: Ah, I see.

patak-dev:
In MacOS/Chrome, if I press the fullscreen on a video, then press exit, the app goes back to the top of the timeline. I don't know if we should merge this as it may confuse folks 🤔

alexzhang1030:
> In MacOS/Chrome, if I press the fullscreen on a video, then press exit, the app goes back to the top of the timeline. I don't know if we should merge this as it may confuse folks 🤔

Ah, it's indeed a problem, I'll see if I can do something...

alexzhang1030:
@patak-dev It works! Can't believe that the issue, I can't believe this is what caused the problem. Chrome sucks 🤪

alexzhang1030:
Ah, netlify does not update the preview

patak-dev (approved):
Nice! Interesting that the `height` had such an effect. Let's try it out in main.elk.zone. I tested in Android and desktop and using min-height looks good.

PR #2283, opened 2023-07-29
feat: hide original post area in the edit widget

### LINKED ISSUE #2276: Consider flipping the order of the Edit UI
Thanks for your work!

Just a suggestion - whenever I edit I instinctively try to edit the top box, which is actually showing the existing post, not the bottom box which is the Edit UI.

I think this makes sense as:

a) When you compose a new post, the UI is at the top
b) Newer posts in timelines appear at the top
c) Editing is your primary task here, not seeing the original post

But others may disagree!

shuuji3:
resolve #2276

You can also the alternative implementation (original flipping layout) for comparison: #2284


patak-dev (approved):
Thanks for creating the other PR for reference. I think both are good. Let's try hiding and we can change it if we receive further feedback.

PR #2284, opened 2023-07-29
feat: flip two post areas in the edit widget

shuuji3:
see #2283

This is just a preview PR for reference.


patak-dev:
We merged https://github.com/elk-zone/elk/pull/2283

PR #2296, opened 2023-08-01
feat: show delete confirmation dialog when clicking "Delete & re-draft" menu

### LINKED ISSUE #2291: Confirmation dialog for delete-and-redraft
There is a confirmation for deleting a toot, but not for deleting and redrafting one. Yeah, it's a less destructive operation because you have the toot content there again in the popup, but it can still break threads, lose boosts etc so I think there ought to be a confirmation.

shuuji3:
resolve #2291 

This change ensures to show the same dialog message as the "Delete" menu. I'm wondering if we'd better change the message but basically, both actions are warning about the same delete action so I assume the same message can be acceptable.

Screenshot 2023-08-02 at 2 05 58


shuuji3:
Ah, it seems that the CI error is not related to this change. It started from this commit: https://github.com/elk-zone/elk/commit/e6172ad38b16a6dad5ee9c0656fee56b607df068 in the `main` branch.

userquin:
What's the default button focused when shown (if any)? We should focus the cancel button.

Can you add `Delete post` title?
Can you also avoid center the mesaage?

Maybe the dialog component requires a review.

antfu (approved):

antfu:
@userquin Yeah I agree we should revise the confirm dialog. It's not blocking this PR tho.

shuuji3:
Thank you both! I also agree on the above suggestions will improve the dialog UI. I'm going to try to make a separate PR to adjust the layout and add a title to the dialog later. 🙂

PR #2307, opened 2023-08-06
feat: improve title and layout of the confirm dialog

shuuji3:
This is follow-up PR for #2296.

Actually the confirm dialog component has had both title description so all I had to is just using `description` prop in addition to `title` prop. Currently, long test is used as `title` that was caused centered bold text.

So I moved the previous `title` to `description` and added a new short `title` like `Delete post` instead.

I also tried to implement the button focus but it was not strightfoward so I leave it for the future issue.

## Before
Screenshot 2023-08-02 at 2 05 58

## After
Screenshot 2023-08-02 at 20 01 58


components/status/StatusActionsMore.vue
```diff
@@ -83,18 +84,12 @@ async function deleteStatus() {
...
   }) !== 'confirm')
     return
 
-  if (process.dev) {
-    // eslint-disable-next-line no-alert
-    const result = confirm('[DEV] Are you sure you want to delete and re-draft this post?')
-    if (!result)
-      return
-  }
-
```
shuuji3, line 97: We will be asked by confirm dialog with this update so this duplicated alert won't be needed.

composables/masto/relationship.ts
```diff
@@ -41,6 +41,7 @@ export async function toggleFollowAccount(relationship: mastodon.v1.Relationship
   if (unfollow) {
     if (await openConfirmDialog({
       title: i18n.t('confirm.unfollow.title'),
+      description: i18n.t('confirm.unfollow.description', [`@${account.acct}`]),
```
shuuji3, line 44: Unlike the other confirm dialog, this message didn't have an account name place holder so I added one for a better dialog message.

locales/en.json
```diff
@@ -127,27 +129,32 @@
...
       "confirm": "Show",
-      "title": "Are you sure you want to show boosts from {0}？"
+      "description": "Are you sure you want to show boosts from {0}?",
+      "title": "Show boosts"
     },
     "unfollow": {
       "cancel": "Cancel",
       "confirm": "Unfollow",
-      "title": "Are you sure you want to unfollow?"
+      "description": "Are you sure you want to unfollow {0}?",
```
shuuji3, line 156: As explained in the above comment, I added `{0}` for `en.json` but couldn't add other languages. Is it OK to keep other messages without `{0}`?

userquin:
don't center the title, it should be in left or right (depending on lang ltr or rtl), check also the message and the buttons alignment:
[RTL Support](https://github.com/elk-zone/elk/blob/main/CONTRIBUTING.md#rtl-support) in contributing guide

shuuji3:
Thanks for the feedback and resource. Deleted the title centering and check the RTL language. Is it the expected layout?

Screenshot 2023-08-08 at 3 06 11

Screenshot 2023-08-07 at 2 35 24

components/status/StatusActionsMore.vue
userquin: Keep this confirm, it is for dev only

userquin:
LGTM we can add focus and maybe focus trap later, we should add also a11y (aria roles and proper markup)

components/status/StatusActionsMore.vue
shuuji3: @userquin All right, reverted this just now.

locales/en.json
userquin: Can you add a new one entry and use `te` to check if present?

userquin: If missing new one use current otherwise use the new one with `{0}`

userquin: IIRC `te` is for checking entry.

userquin: upps, forgot it, we'll need to add the title, I'll add the title for spanish and try to add focus trap

userquin: rn all confirm dialog titles will be using `en`  entry

userquin (approved):

ryanccn (approved):

patak-dev:
@shuuji3 this looks good to me, thanks! If you resolve the conflicts, we can merge it.

shuuji3:
@patak-dev OK, I've just resolved the code conflict. Thanks!

patak-dev (approved):

Issue #1855, opened 2023-03-05
feat: keyboard shortcuts (v2)


hartmut-co-uk:
note: follow up issue of https://github.com/elk-zone/elk/issues/39

TODO: plan next steps / shortcuts

PR #2324, opened 2023-08-10
fix: some keyboard shortcuts not working in non-english

Sma11X:
This PR fix that the `f` and `b` keyboard shortcuts could not be triggered if user language is not English.

And I've noticed that Elk is lacking some shortcuts, should it be synced as mastodon or is there a better solution?


patak-dev (approved):

patak-dev:
Awesome! Thanks 🙌🏼

patak-dev:
See the discussion here https://github.com/elk-zone/elk/issues/1855, there were ideas for a lot more shortcuts. What we got was the first MVP. See the linked issues at the end of that issue for an exploration another contributor was doing.

PR #2326, opened 2023-08-11
feat: disabled boost button

lazzzis:
Show a disabled boost button when the toot can't be boosted.
1. lower opacity
2. replace with a repeat-off icon

Screenshot 2023-08-10 at 11 36 19 PM


EddiesTech:
Nice work! :)

antfu (approved):

PR #2333, opened 2023-08-12
fix: role indicator

iruldi:
## Fix role indicator

Linked to discussion on: https://github.com/elk-zone/elk/pull/2265

Fixing the box from showing up even if the user didn't have a role.


antfu:
Why the component get removed?

iruldi:
@antfu It's because it's an old one which is not used anymore. We use AccountRolesIndicator instead of AccountRoleIndicator

antfu (approved):

Issue #2272, opened 2023-07-26
Text input area too small - often need to scroll to review


heaths:
When publish a new post using the mobile site, as I approach about 300 words a scroll bar is added making it harder review my text before posting. It'd be great if the area was larger accommodating at least the 500 char limit many instances seem to impose. Showing a bunch of existing replies below the input area just seems like a waste of screen real estate, and eliminating that would leave plenty of room for input area.

![Screenshot_20230726_060633_Edge](https://github.com/elk-zone/elk/assets/1532486/3d02692a-14f1-422c-a3a4-027ed3a3b3af)

PR #2334, opened 2023-08-12
fix: publish height

iruldi:
## Fix publish textarea height size

fix: https://github.com/elk-zone/elk/issues/2272

It was a wrong use of tailwind breakpoints


antfu (approved):

patak-dev:
I reverted this PR at https://github.com/elk-zone/elk/commit/0343898146bfdef9babe368d36edbcb2ac7b962d because it was causing issues (see #2363 and #2416)

Issue #709, opened 2023-01-02
Follow Requests in Notifications


vwchu:
I set up my Mastodon account to require follow requests. So, I can manually control who can follow me by approving follow requests. I also set up my notifications to include follow requests. However, they don't seem to display correctly within the Notifications page nor is there a Follow Requests page or a Follow Requests tab within the Notifications page.

hendrikmitk:
I just stumbled across this issue myself and had some serious trouble figuring out what the actual problem is. It wasn't until I looked at the Notification view in the Mastodon default web GUI that I realized this is about follow requests.

Here is a screenshot here showing the broken display.

![Screen Shot 2023-01-19](https://user-images.githubusercontent.com/61987146/213507611-a5ae29c4-f2e6-407b-83df-d542311ec7d9.jpg){:width="50%" height="auto"}

windhamdavid:
I also just came across this issue.

stephen304:
Also noticed this, the way it is right now, it just looks like that person's profile card is in your notifications for no reason, and the only action you can take is following / unfollowing them, instead of being able to accept or reject the follow request

SDK4rma:
Indeed, it seems a "follow requests" menu/tab is missing

PR #2339, opened 2023-08-14
feat: handle follow request

Sma11X:
- Allow locked users handle follow requests at notifications and use hover-card instead of the tips above the button
- Add tips at the top of requested user card if user has requested a follow

# Before
image

# After
image

# Tips
image

And I'm not sure if this is the appropriate style for the buttons, if has better ideas please let me know.


components/notification/NotificationCard.vue
```diff
@@ -60,12 +60,14 @@ const { notification } = defineProps<{
       </NuxtLink>
     </template>
     <template v-else-if="notification.type === 'follow_request'">
-      <div flex ms-4 items-center class="-top-2.5" absolute inset-ie-2 px-2>
+      <!-- <div flex ms-4 items-center class="-top-2.5" absolute inset-ie-2 px-2>
         <div i-ri:user-shared-fill text-xl me-1 />
         <AccountInlineInfo :account="notification.account" me1 />
-      </div>
+      </div> -->
```
patak-dev, line 66: Let's remove this commented part

patak-dev:
Thanks! This is way better than before, we can merge it and keep improving later on. One thing that would be good is to have "Accept follow request" or similar in the notification. Maybe use a similar icon from the follow notification: 
image

Sma11X:
> Thanks! This is way better than before, we can merge it and keep improving later on. One thing that would be good is to have "Accept follow request" or similar in the notification. 

I used the same color as follow info and icon used before. And it changed as:
image

components/notification/NotificationCard.vue
Sma11X: Removed it.

patak-dev (approved):
Awesome, thanks!

waldyrious:
Does this affect #709?

Issue #2199, opened 2023-07-04
Suggestion: click on user’s @[username]@[server] to copy it


MIfoodie:
<img width="1024" alt="image" src="https://github.com/elk-zone/elk/assets/94649676/8eef6fc0-00c1-40fc-84ba-fd790ec4ac90">
This would be useful becuase madtodon usually has some pretty long usernames and it would be a nice QOL change.

ryanccn:
This could possibly lead to some confusing behavior since the whole avatar-name-username group is a click target IMO

MIfoodie:
> This could possibly lead to some confusing behavior since the whole avatar-name-username group is a click target IMO

Maybe then you could use a different button (like middle mouse button).

patak-dev:
One issue we also have is that we need to do a review of the padding around all things clickable, so probably a good idea to avoid having another thing to click in that region as @ryanccn suggested.

And a different button would be hard to discover. Maybe we could add this functionality to the three dots menu.

miclgael:
@MIfoodie I am curious what are you usually doing with the **@[username]@[server]** after copying it? 

The **@[username]@[server]** on an individual 'Toot links to the user's profile. 

The same in the users profile is currently plain-text. Would it be useful to have this copying functionality on the profile itself?

![a screen shot of a toot and the profile, visually demonstrating the above text](https://github.com/elk-zone/elk/assets/4301358/3ea526d5-1146-43c3-a95a-47eac3f2896e)

shuuji3:
I think it would be better to have this feature on the profile page or/and a three-dot menu to avoid confusion.

Here's a case study from two front-end implementations.

In Mastodon, when clicking the user id, the entire text is selected by JavaScript to help copy and paste (without copying it automatically). This may not be obvious to all users.

https://mastodon.social/@mastodon

<img width="903" alt="Screenshot 2023-08-06 at 14 43 38" src="https://github.com/elk-zone/elk/assets/1425259/d23735f4-cb1c-4324-b091-7a8044d2931a">


Takahē has a copy button on the profile page next to the user id. Someone may think it's not beautiful to put a button on every id, but easy to recognize and use.

https://takahe.social/@takahe@jointakahe.org/

<img width="897" alt="Screenshot 2023-08-06 at 14 41 59" src="https://github.com/elk-zone/elk/assets/1425259/3d9c4574-fd34-4423-84c2-c0158c492ac1">

miclgael:
[...]

PR #2347, opened 2023-08-16
feat: put account name copy button

shuuji3:
This is one option to solve #2199. Please see the issue for details and other options.

### DIFF
--- a/components/account/AccountHeader.vue
+++ b/components/account/AccountHeader.vue
@@ -22,6 +22,7 @@ const namedFields = ref<mastodon.v1.AccountField[]>([])
 const iconFields = ref<mastodon.v1.AccountField[]>([])
 const isEditingPersonalNote = ref<boolean>(false)
 const hasHeader = $computed(() => !account.header.endsWith('/original/missing.png'))
+const isCopied = ref<boolean>(false)
 
 function getFieldIconTitle(fieldName: string) {
   return fieldName === 'Joined' ? t('account.joined') : fieldName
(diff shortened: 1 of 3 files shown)


components/account/AccountHeader.vue
```diff
@@ -171,7 +183,15 @@ const personalNoteMaxLength = 2000
             <AccountLockIndicator v-if="account.locked" show-label />
             <AccountBotIndicator v-if="account.bot" show-label />
           </div>
-          <AccountHandle :account="account" overflow-unset line-clamp-unset />
+
+          <div flex items-center gap-1>
+            <AccountHandle :account="account" overflow-unset line-clamp-unset />
+            <CommonTooltip placement="bottom" :content="$t('account.copy_account_name')" no-auto-focus flex>
+              <button text-secondary-light text-sm i-ri:file-copy-line @click="copyAccountName">
```
ryanccn, line 190: Would be great if the button had some sort of response, like changing the icon to a checkmark and changing back after a few seconds.

ryanccn (changes requested):

components/account/AccountHeader.vue
shuuji3: Thank you for the review.🙂 That's a good suggestion. I modified it to show a check mark for 2 seconds.

shuuji3:
New check mark after clicking:

https://github.com/elk-zone/elk/assets/1425259/7faa4354-2136-42d8-9c00-9e79c26d5807

ryanccn (approved):

patak-dev (approved):

PR #2365, opened 2023-08-25
feat: add keyboard shortcut helper link or button

shuuji3:
This adds an explicit link/button to show the keyboard shortcut helper panel. I tried to implement both but that's just to evaluate which one is better (will revert one of the commits later).

I saw some people commented somewhere (sorry, couldn't find the original comment) that they didn't notice the keyboard shortcut is supported by Elk. This link/button can help to find the feature.

Screenshot 2023-08-26 at 2 15 03

Screenshot 2023-08-26 at 2 14 54

### DIFF
--- a/components/nav/NavFooter.vue
+++ b/components/nav/NavFooter.vue
@@ -92,5 +92,10 @@ function toggleDark() {
         GitHub
       </NuxtLink>
     </div>
+    <div>
+      <button hover:underline @click="toggleKeyboardShortcuts">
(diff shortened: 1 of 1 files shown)


userquin (approved):

antfu:
Could we put it between `Zen Mode` and `Spsonsors`? I'd prefer to have the light/dark switch always on the left

antfu (approved):

PR #2371, opened 2023-08-31
feat: filter notifications by type

ayoayco:
add capability to filter notifications by all types as supported by mastodon: 

This PR will 
1. create new dynamic notifications filter page
2. extend the CommonRouteTabs component to have more options in a dropdown
3. add new constant array for notification types
4. add translation strings
5. add typeguards for checking valid Notification / Notification Filters

notifs_filter


iruldi:
I would have used a filter icon instead of the down arrow so the users can see where to filter. It's not really intuitive here.

ayoayco:
> I would have used a filter icon instead of the down arrow so the users can see where to filter. It's not really intuitive here.

Good one. I extended the CommonRouteTabs component to take more options, which is a new pattern for the app. Initially I decided to make it more generic (with just the caret icon) if we want to reuse it somewhere else, that is, for non-fiter actions.

I pushed another commit to extend the config we pass to the moreOptions prop so we can have custom icons for it, now displaying a filter icon for this Notifications page.

Also made it to look to have similar design to the dropdowns when writing/editing posts:

Screenshot 2023-09-01 at 13 22 01

How dropdowns look for editing posts:

Screenshot 2023-09-01 at 13 42 43

userquin:
@ayoayco I'll check it in 1 hour or so, I'm releasing PWA plugin stuff...

userquin:
@ayoayco I'll fix a few things...

userquin:
@ayoayco @patak-dev we  need to review StatusCard SFC, I'm getting this error:

imagen

ayoayco:
@userquin sounds good

userquin:
@ayoayco 

imagen

userquin:
Added some reactivity to router filter param, must be reactive to the params and the hydration state.

We also need to change a few things for a11y, the dropdown menu key traversing is wrong, cannot use tab key, should use key down/up (with rewind, check a simple `<select>`)

userquin:
I'll add Spanish traslation for the new keys, you can add the local storage here later...

PR LGTM

ayoayco:
@userquin regarding the tooltip being dynamic, I’m thinking later we have to remove the selected one if/when we support multiple selection. Otherwise it would grow long. But let’s just handle later and iterate on it

userquin:
uhmm, if you want this behavior, you cannot use dynamic route, we should add a custom page where we can filter the entries,so it would be better to add a filter in All and Mentions page.

ayoayco:
Hmm yeah you’re right. That’s not going to be in this dynamic filter page.

Don’t know yet what is better, this is good for now then let’s see

ayoayco:
Will also work on localstorage in next PR

userquin (approved):
@ayoayco review my changes, we also need to review StatusCard SFC to remove the error (another PR)

ayoayco:
Checking...

ayoayco:
@userquin looks good to me, but I can't approve as the author of the PR 😄

pages/notifications.vue
```diff
@@ -22,17 +27,64 @@ const tabs = $computed<CommonRouteTabOption[]>(() => [
...
 
+const filter = $computed(() => {
+  if (!isHydrated.value)
+    return undefined
+
+  const rawFilter = route.params?.filter
+  if (!rawFilter)
+    return undefined
+
+  if (NOTIFICATION_TYPES.includes(Array.isArray(rawFilter) ? rawFilter[0] : rawFilter))
```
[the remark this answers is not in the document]
userquin: sure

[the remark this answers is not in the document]
userquin: done

ayoayco:
hey @userquin sorry for the change again... I pushed a change to support the two admin notification types (sign_up & report), because I saw we actually support them in the NotificationCard component

ayoayco:
Also we might want to make the icons in our dropdown here the same as the icons used in `NotificationCard.vue`? What do you think, @userquin should we update that component too?

userquin:
ok

userquin:
@ayoayco we need to review the button in narrow screens:

imagen

userquin:
imagen

pnpm-lock.yaml
```diff
@@ -1,4 +1,4 @@
-lockfileVersion: '6.0'
+lockfileVersion: '6.1'
```
userquin, line 1: @ayoayco upps, you should update `pnpm` via `corepack`, this version is wrong, can you run `pnpm --version`  in your local?

I'll try to fix it tmr.

userquin: I don't check it since I always run `pnpm install --frozen-lockfile`, we should exclude the lockfile in this PR.

ayoayco: I pushed a revert for this lock file

pages/notifications.vue
```diff
@@ -20,30 +26,95 @@ const tabs = $computed<CommonRouteTabOption[]>(() => [
...
+    case 'admin.sign_up':
+      acc[name] = 'i-ri:user-add-line'
+      break
+  }
+  return acc
+}, {} as Record<string, string>)
+
+const filterText = $computed(() => (`${t('tab.notifications_more_tooltip')}${filter ? `: ${t(`tab.notifications_${filter}`)}` : ''}`))
+
+const more = $computed<CommonRouteTabOption[]>(() => NOTIFICATION_TYPES.filter(type => type !== 'mention').map(
```
ayoayco, line 79: If we can determine that currentUser is not an admin, we should add a condition to not show the two notification types that are for admins only. But that could also be for another PR.

ayoayco: I'll create a separate task for this one... I cannot find it yet

ayoayco:
How it looks now :)
Screenshot 2023-09-02 at 10 52 18

userquin:
Follow request and sign up entries using the same icon, change one of them

ayoayco:
Did some clean up and added typeguards to check for valid Notification / Notification Filters :)

ayoayco:
@userquin @patak-dev this feature is all cleaned up and ready!

danielroe (approved):
beautiful

patak-dev (approved):
🧡

PR #2372, opened 2023-09-01
feat: add verified icon to verified links

### LINKED ISSUE #155: Verified website, github, twitter in header

Thomas-Philippot:
## Add 'check' icons on verified links

Not sure if it's the right way to do it, couldn't find a 'checked' version of the icons

If it's ok, then this closes #155 

### Before
elk_before

### After
elk_after

### DIFF
--- a/components/account/AccountHeader.vue
+++ b/components/account/AccountHeader.vue
@@ -225,11 +225,19 @@ const personalNoteMaxLength = 2000
         </div>
       </div>
       <div v-if="iconFields.length" flex="~ wrap gap-2">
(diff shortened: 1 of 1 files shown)


EddiesTech:
There is already a pull request for a new design that includes a verified icon like this, but it has not been merged. I would pledge for that one to be merged as it has other improvements.
https://github.com/elk-zone/elk/pull/1694

PR #2386, opened 2023-09-08
fix(unocss): resolve match warning for uno

zyyv:
Because of Uno extract the untransformed code to parse, made some warnings.
image

As a workaround, we can prepare the unitilities of generated in comment node.

### DIFF
--- a/components/status/StatusActions.vue
+++ b/components/status/StatusActions.vue
@@ -81,7 +81,7 @@ function reply() {
         :text="!getPreferences(userSettings, 'hideFavoriteCount') && status.favouritesCount ? status.favouritesCount : ''"
         :color="useStarFavoriteIcon ? 'text-yellow' : 'text-rose'"
         :hover="useStarFavoriteIcon ? 'text-yellow' : 'text-rose'"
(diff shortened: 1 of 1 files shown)


edimitchel (approved):

components/status/StatusActions.vue
```diff
@@ -112,5 +112,8 @@ function reply() {
         @click="toggleBookmark()"
       />
     </div>
+
+    <!-- ⚠️ Placeholder for UnoCSS parse. ⚠️ -->
+    <!-- <template elk-group-hover="bg-yellow/10" elk-group-hover="bg-rose/10"></template> -->
```
userquin, line 117: cannot be moved to the script setup?

zyyv: Anywhere, but need to meet the support for `presetAttributify`.

I mean, it's just a comment node, seems to have no other effect.

userquin (approved):

antfu:
I don't think this is a long-term solution. This adds extra mental burden as everytime we do so we need to remember to update the comment as well, which I don't it's a not ideal solution. I think we should improve the shortcut instead. I'll see what I can do

edimitchel:
I was wondering some like that, isn't something UnoCSS should handle better?

PR #2393, opened 2023-09-15
feat: manage followed tags

Sma11X:
This PR add a page to manage followed tags for unfollowing tags

- added at three dots
image

- fetch followed tags list
image


Sma11X:
close for now. duplicated with #1251

PR #2397, opened 2023-09-16
feat: show relationship badge and remove hover-only label on follow button

### LINKED ISSUE #1020: Improve the way to follow someone

When someone follows me, I think that finding the way to follow him/her is not so obvious. See the screenshot. 

B3A87BCC-6AAB-4881-94FD-3762E92A0224

The « follow area » tells me that the user follows me. 

### Suggested solution

Replace the text by « follow back ». 

Maybe it’s in fact a translation issue. 

### Alternative

_No response_

### Additional context

_No response_

shuuji3:
resolve #1020

This PR implements two changes:
- Avoid the current hover-only label on the follow button, which means the button label is always the same regardless hovered or not.
- Add a new relationship badge: "following you" or "mutuals".

## Examples

When the account follows you but you don't follow them:
Screenshot 2023-09-17 at 0 32 17

When the account follows you and you follows them:
Screenshot 2023-09-17 at 0 27 49


nicosomb (approved):

edimitchel:
Why not putting the relationship badge aside the username?

shuuji3:
@edimitchel Because I felt a bit crowded when the account had many badges or the name was long, also it can mimic the other platform. But maybe it was not necessary, and better to place the same position as other badges for consistency. Let me adjust the position.

shuuji3:
Yes, it was not as bad as I thought.

Screenshot 2023-09-17 at 16 41 55
Screenshot 2023-09-17 at 16 44 23

maybeanerd:
This PR seems like it would be ready (If not maybe for a rebase)

Ping to revive it? Looking forward to this change 👀

components/account/AccountRelationshipIndicator.vue
```diff
@@ -0,0 +1,18 @@
+<script setup lang="ts">
+import type { mastodon } from 'masto'
+
+defineProps<{
+  relationship?: mastodon.v1.Relationship
```
maybeanerd, line 5: IMO this shouldn't be optional and rather the v-if should be applied outside of this component

maybeanerd (changes requested):

components/account/AccountRelationshipIndicator.vue
shuuji3: Thanks for your review. That makes sense. I changed it to be non-optional.

shuuji3:
To preview the PR, you can visit, for example, https://deploy-preview-2397--elk-zone.netlify.app/universeodon.com/@elk@webtoo.ls to see "Mutuals" badge.

maybeanerd (approved):
LGTM!

shuuji3:
Merging after opening 2.5 years later... 🥲

PR #1964, opened 2023-04-10
fix: wrong profile showing on hover sometimes (#1913)

JohnyTheCarrot:
closes #1913

PR #2407, opened 2023-09-20
fix: mention in content do not render hover card

### LINKED ISSUE #2362: Mentions in content not render hover card
When `Hide account hover card` is disable, it might render a hover card but fail
image

Sma11X:
resolve #2362 
Ensure the `account` value not wrapped by `ref` to get the right route for hover card.

Also fix the account at profile bio do not render the hover card.

### DIFF
--- a/components/account/AccountHoverCard.vue
+++ b/components/account/AccountHoverCard.vue
@@ -1,10 +1,11 @@
 <script setup lang="ts">
 import type { mastodon } from 'masto'
 
-const { account } = defineProps<{
+const props = defineProps<{
   account: mastodon.v1.Account
 }>()
 
+const account = props.account
 const relationship = $(useRelationship(account))
 </script>
 


sxzz (commented):
~~I cannot reproduce the issue. Could you please provide some steps or video?~~
Reproduced

sxzz:
Actually caused by this line. `account` could be `Ref<mastodon.v1.Account>` or `Ref<Ref<mastodon.v1.Account>>`. We should find a way to make sure it's always a ref.
https://github.com/elk-zone/elk/blob/20c30e92a367e980e48d348af71c2f5ed5bf46c8/components/account/AccountHoverWrapper.vue#L14

Sma11X:
I've tried to unref the state return by `useAccountByHandle` but it will block the page loading. And due to #1964 we also cannot remove `computed`.
Have no ideas about it.😥

PR #2417, opened 2023-09-30
feat: display embedded media player

### LINKED ISSUE #356: Embedded videos from Streaming services
If the PreviewCard.type is `video` and PreviewCard.html contains  an iframe, we can embed the video in the StatusCard or StatusDetail and be able to watch on Elk. 

For video preview, There's a case when the width and height property does not match the image
    - this has a square image: https://elk.zone/fosstodon.org/@polarfront@snabelen.no/109474214556574293
    - but we're getting `width: 350` and `height: 467`
    - I think because this is a `type: video` and the `width` and `height` properties might be describing the iframe instead of the image

We can go for two overlay buttons: one to play video on Elk and other links to the source:

Screenshot 2022-12-07 at 2 22 41 PM

YouTube is giving an example.

```
{
    "url": "https://www.youtube.com/watch?v=-R0UYHS8A_A",
    "title": "Afternoon Jazz 🎷 [lofi hip hop/study beats]",
    "description": "",
    "type": "video",
    "author_name": "Lofi Girl",
    "author_url": "https://www.youtube.com/@LofiGirl",
    "provider_name": "YouTube",
    "provider_url": "https://www.youtube.com/",
    "html": "<iframe width=\"200\" height=\"113\" src=\"https://www.youtube.com/embed/-R0UYHS8A_A?feature=oembed\" frameborder=\"0\" allowfullscreen=\"\" title=\"Afternoon Jazz 🎷 [lofi hip hop/study beats]\"></iframe>",
    "width": 200,
    "height": 113,
    "image": "https://cdn.fosstodon.org/cache/preview_cards/images/012/574/228/original/1afec34c903be880.jpg",
    "embed_url": "",
    "blurhash": "UAB.7RnhNHJAxtoeoIR-0~NcxYR+EMENs:xY"
}
```

## Screenshots (current preview w/o embed)

Screenshot 2022-12-06 at 4 55 50 PM

ayoayco:
close #356 

This PR will:
1. create new `StatusEmbeddedMedia.vue` that will show instead of the preview card when there is a value in `'status.card.html'` -- usually when sharing streamable media
2. add `sanitizeEmbeddedIframe` function to clean up the value and only allow `iframe`s with `src` attribute on `https` protocol
3. respect the Mastodon media visibility preference -- will hide the embedded iframe if user has the preference to hide all media
4. add an overlay with play icon that follows the preferred primary color, and some information for users: provider site, author, title, and warning about IP address exposure
5. hide feature behind preference/flag "Embedded Media Player"

**@see** Example post with lots of shared video links: https://deploy-preview-2417--elk-zone.netlify.app/hci.social/@bwaber/111157708714070414

## how it looks

elk-embed-media-3


components/status/StatusEmbeddedMedia.vue
```diff
@@ -0,0 +1,20 @@
+<script setup lang="ts">
+import type { mastodon } from 'masto'
+
+const { status } = defineProps<{
+  status: mastodon.v1.Status
+}>()
+
+const sanitizedHtml = ref(await sanitizeEmbeddedIframe(status.card?.html || ''))
```
patak-dev, line 8: `sanitizeEmbeddedIframe` isn't an async function, no? The `await` isn't needed here.
It should be a `computed` from the status passed as a prop (see other derived variables in the other status cards components.

patak-dev:
Awesome! 🙌🏼

ayoayco:
Weird thing on deployed preview, iframes seem to be blocked... Maybe a netlify setting? @danielroe

components/status/StatusEmbeddedMedia.vue
ayoayco: ahhh I'll try :) always still not sure which one to use. My understanding is `computed()` when combining variables from `ref()`?

ayoayco: Pushed a change, not sure if it is the best... I use `computed` from the `status` prop, but can't make it work without `ref` and `await` when using `sanitizeEmbeddedIframe`, because that function returns a `Promise<string>` 😬 -- any ideas?

Here's what I ended up with:
```ts
const cardHtml = $computed(() => status.card?.html ?? '')
const sanitizedHtml = ref(await sanitizeEmbeddedIframe(cardHtml))
```

ayoayco: Ah, found how we do it for content in StatusBody... will do that too

patak-dev:
I'm worried about having many iframes automatically open on the user timeline. I think there is a reason Twitter only creates one when you click on the video. I think we should replicate how it works there (example: https://twitter.com/RyanCarniato/status/1699911405208801597). Maybe also add an option to disable expansion on an iframe and directly going to the external service instead?

ayoayco:
I see, yeah we can go for similar with what mastodon does too and only insert the iframe when user clicks the play button 👍

ayoayco:
I'm showing the card thumbnail with a play icon overlay, and only after the play icon is clicked we insert the iframe.

components/status/StatusEmbeddedMedia.vue
ayoayco: Done!

ayoayco:
@patak-dev Implemented an overlay with the play icon and some information: provider site, author, title, and warning about IP address exposure. Here's how it looks:

Screenshot 2023-09-30 at 21 53 49

Also the link opens the source page in a new tab :)

embed-peertube

patak-dev:
About the blocked iframes, maybe this feature could be under a flag for the moment defaulting to false so you can test how it works in main.elk.zone

ayoayco:
Oh yeah, I'll create a feature flag setting

ayoayco:
@patak-dev sorry forgot to follow this up... it should be ready, behind a feature flag so we can test on main.

antfu:
It think it would be nice if the play button were a bit faded. Other than that it looks good to me 👍

ayoayco:
> It think it would be nice if the play button were a bit faded. Other than that it looks good to me 👍

@antfu I was playing around with different levels of opacity and couldn’t decide in the end, so I removed 😄 I’ll add a bit of fade that change on hover. :)

ayoayco:
Added some default fade (85%) that change on hover to less fade (95%)

faded-play-button

patak-dev (approved):
Awesome! I think we could try it out in main.elk.zone

patak-dev:
Let's merge this one so we can test it out. Thanks again @ayoayco!

VipinVIP:
IDK if this should be created as a new issue, but for all embedded media's are not playing in the latest canary build. 
Checking the console show's CSP error like this

> 
	Content-Security-Policy: The page’s settings blocked the loading of a resource at https://www.youtube.com/embed/9SfwaDWHTZ8?feature=oembed (“default-src”). 

I so far have checked only with firefox browser and Mullvad browser

ayoayco:
Thanks for checking @VipinVIP! we weren’t sure if this will happen in canary so we put it behind a toggle and set to false. Looks we need to update a setting but I don’t know yet where.

PR #2421, opened 2023-10-06
feat(i18n): Update the fr-FR locale

abeldotam:
Hi folks, how are you doing? Everything is good in your life atm?
I added some french translations. 🙂

Before: `fr-FR 96.4% (529/549)`
After: `fr-FR 99.1% (544/549)`

5 still missing, but there's just one translation in Hungarian, not sure of the meaning so I'll wait for more context.
In the meantime, it's already that!

Cheers.

### DIFF
--- a/locales/fr-FR.json
+++ b/locales/fr-FR.json
@@ -41,7 +41,8 @@
     "unfollow": "Ne plus suivre",
     "unmute": "Réafficher",
     "view_other_followers": "Les comptes abonnés d'autres instances peuvent ne pas être affichés.",
-    "view_other_following": "Les comptes suivis d'autres instances peuvent ne pas être affichés."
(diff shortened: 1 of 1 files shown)


roiLeo (approved):

locales/fr-FR.json
```diff
@@ -86,6 +88,7 @@
     "activate": "Activer",
     "complete": "Compléter",
     "compose_desc": "Écrire un nouveau message",
+    "n-people-in-the-past-n-days": "",
```
patak-dev, line 91: This seems to be a mistake from the Hungarian translation, it should have been with `_` like the one below. Will you remove all the keys you added that have a `-`?

abeldotam: Indeed, that's more like a translation with i18n-ally that seems to failed, my bad for that.
Removed.

patak-dev (approved):

PR #2423, opened 2023-10-08
fix: profile header avatar being pushed lower on smaller devices

T0TProduction:
this fixes an issue where on smaller devices the avatar was moved down, not aligning with the profile header anymore

an example of the issue can be seen here: 
image
Fixed it looks like this:
image

to test the fix, you can take a look at any profile with header image and profile picture on a smaller screen ( when the profile buttons wrap around), the account I used is http://localhost:5314/bumscode.com/@tagesschau@ard.social on screen width 320px

### DIFF
--- a/components/account/AccountHeader.vue
+++ b/components/account/AccountHeader.vue
@@ -115,7 +115,7 @@ const personalNoteMaxLength = 2000
     <div p4 mt--18 flex flex-col gap-4>
       <div relative>
         <div flex justify-between>
-          <button shrink-0 :class="{ 'rounded-full': !isSelf, 'squircle': isSelf }" p1 bg-base border-bg-base z-2 @click="previewAvatar">
+          <button shrink-0 h-full :class="{ 'rounded-full': !isSelf, 'squircle': isSelf }" p1 bg-base border-bg-base z-2 @click="previewAvatar">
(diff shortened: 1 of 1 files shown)


components/account/AccountHeader.vue
```diff
@@ -115,7 +115,7 @@ const personalNoteMaxLength = 2000
     <div p4 mt--18 flex flex-col gap-4>
       <div relative>
         <div flex justify-between>
-          <button shrink-0 :class="{ 'rounded-full': !isSelf, 'squircle': isSelf }" p1 bg-base border-bg-base z-2 @click="previewAvatar">
+          <button shrink-0 class="h-full" :class="{ 'rounded-full': !isSelf, 'squircle': isSelf }" p1 bg-base border-bg-base z-2 @click="previewAvatar">
```
patak-dev, line 118: This can directly be `h-full`, no?
```suggestion
          <button shrink-0 h-full :class="{ 'rounded-full': !isSelf, 'squircle': isSelf }" p1 bg-base border-bg-base z-2 @click="previewAvatar">
```

T0TProduction: Ah I forgot that was possible in this project, you're totally right

T0TProduction: adjusted it 😄

patak-dev (approved):

PR #2424, opened 2023-10-08
feat(ui): Color of the color scheme for AccountMoreButton

birksovskiy:
I was going to sleep, but made this.
image


patak-dev:
Thanks for the PR but other buttons in the same section (like the bell) have their own color. And the three dots for the account switcher doesn't have any color. So if we change something here, we probably need a review of the colors in all these buttons. For now, I think we can keep the current one.

PR #2426, opened 2023-10-10
update: use instance's translation service

HolgerHuo:
Since Mastodon v4.0.0, post translation api is exposed at `/api/v1/statuses/:id/translate` as well as through `masto` package. Thus we can use instance's translation service as Elk's translation provider.

### DIFF
--- a/composables/masto/translate.ts
+++ b/composables/masto/translate.ts
@@ -1,13 +1,5 @@
 import type { mastodon } from 'masto'
 
-export interface TranslationResponse {
-  translatedText: string
-  detectedLanguage: {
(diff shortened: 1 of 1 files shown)


antfu:
I think we shouldn't remove the current implementation, as not all Mastodon server would support that API. Instead, we could use it when available and fallback to the current one.

HolgerHuo:
Hi! I've added a condition to use instance's translation provider only when the instance provides the option. However, the method used to test if the instance supports it seems not so graceful. I haven't come up with a better solution yet. Maybe you'll have any ideas?

PR #2444, opened 2023-10-25
feat(settings): convert metadata to text

### LINKED ISSUE #1216: Editing profile metadata doesn't cope with "cards"
### Pre-Checks

### Describe the bug

I wanted to update the title of one of the pieces of metadata in my profile. The metadata is actually links to some of my other accounts 

When I brought up the edit page those forms were full of html (card markup?) Which was then resaved as plain text.

Elk should remove the formatting in this case. I don't know if the API proceeds am unformatted version or if it would have to manually convert to text?

Screenshot_2023-01-16-10-46-08-719_com microsoft emmx

Profile image after editing metadata in elk. Note all the markup which is how it appeared on the edit page too.

### Additional Context

_No response_

### LINKED ISSUE #2433: @user@server inside profile metadata is read as HTML instead of text
I have a bot i made in the metadata of my profile, and this breaks the metadata when i update my profile with elk.

Steps (all within elk): 
- Add a metadata tag to your profile @'ing another account (for me that's `@billie_lyrics@botsin.space`)
- Save
- That same metadata tag turns into html code inside of the edit field
- When you save again after that, that html code is escaped (with `&lt;` and the such) and shows as raw text on your profile.

Sma11X:
fix #2433 
fix #1216 
This PR adds feature to convert HTML code to text , cause once the metadata has content starts with `@` or `#` , or a link starts with protocols, mastodon will save them as HTML code.

### DIFF
--- a/composables/settings/metadata.ts
+++ b/composables/settings/metadata.ts
@@ -1,1 +1,28 @@
+import type { Node } from 'ultrahtml'
+import { decode } from 'tiny-decode'
+import { TEXT_NODE, parse } from 'ultrahtml'
+
 export const maxAccountFieldCount = computed(() => isGlitchEdition.value ? 16 : 4)
+
+export function convertMetadata(metadata: string) {
+  try {
+    const tree = parse(metadata)
+    return (tree.children as Node[]).map(n => convertToText(n)).join('').trim()
+  }
+  catch (err) {
+    console.error(err)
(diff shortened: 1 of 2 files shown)


patak-dev (approved):

patak-dev:
Thanks! Let's test it out in https://main.elk.zone

PR #2453, opened 2023-11-02
fix(pwa): add ignore vary to web manifests and emoji handlers

userquin:
This is the second attempt to try to fix web manifest in chrome dev tools.


userquin:
There is no way, precaching web manifests also failing...

PR #2456, opened 2023-11-04
feat(i18n): Update the id-ID locale

adydetra:
Done 100%, I've tried it locally, with run `PWA`.

I've checked, maybe if there is time in the future there will be a slight improvement in the wording I will revise. because I am an Indonesian citizen.


patak-dev (approved):
Thanks for the improvements!

PR #2462, opened 2023-11-08
chore(types): include `$pwa` module augmentation types for Nuxt and Vue

userquin:
VSCode (also working when using IntelliJ and should work also with WebStorm):

imagen

imagen


userquin:
upps, reviewing what's happening with `$scrollToTop` and `$i18n`...

userquin:
superseded by #2498

PR #2449, opened 2023-10-30
feat(a11y): improve button color contrast with white text

shuuji3:
This PR attempts to change the button color for better contrast and readability.

See this comment https://github.com/elk-zone/elk/issues/1365#issuecomment-1778542677 for the details.

**Screenshots**
Screenshot 2023-10-30 at 23 22 57
Screenshot 2023-10-30 at 23 22 50

PR #2471, opened 2023-11-13
fix: btn-solid text color for dark mode

patak-dev:
#2449 switched to white for the solid button text color, that looks good on light mode but a bit off in dark mode.

See
image

Against
image

#2449 also didn't change all similar buttons (like the Follow button that had custom logic and was already using `text-inverted` so it is white text for light mode and black text for dark mode). This PR switches to using `text-inverted` everywhere.


antfu (approved):

shuuji3:
Thanks for the following up 🙏🏻

Issue #2472, opened 2023-11-14
Add icons for Friendica, Threads, BlueSky, and Nostr.


expertmanofficial:
If you [visit my profile in Elk](https://elk.zone/@expert@buddyverse.xyz) you'll notice that not all links in my profile has an icon.

That's right. For Friendica I **had to** use the name "Facebook" so it would match with an similar icon, but would be better if there was actually an icon for that.

BlueSky and Threads doesn't have an icon as well which is bad as well. I don't have Nostr but you can find any profile that have it linked, and you'll notice it doesn't have an icon too.

Anyways, please add them all.

expertmanofficial:
OK, Threads icon seems to be added, what about others two?

hawkticehurst:
@expertmanofficial it looks like Elk uses an icon set called RemixIcon (https://remixicon.com/) but they do not have icons for Bluesky, Friendica, or Nostr -- which is likely why only the Threads icon was added.

You'll like want to make an upstream request with RemixIcon to add those icons.

expertmanofficial:
will check one that out

shuuji3:
FYI: Bluesky published a new logo a few days ago (https://blueskyweb.xyz/blog/12-21-2023-butterfly). I commented this link on the above request for the remix icon.

hawkticehurst:
Hey @shuuji3, if you're up for it would you be willing to also add the TikTok logo (https://remixicon.com/icon/tiktok-fill)?

shuuji3:
@hawkticehurst Hi, I just created a PR for that.

hawkticehurst:
Yay, thank you so much! I appreciate it :)

*P.S. no worries at all on choosing the line style of the icon to match with the others.

expertmanofficial:
Hey, so Friendica and BlueSky icons has been added in latest version, could you please add them to Elk?

https://github.com/Remix-Design/RemixIcon/issues/697#issuecomment-1893794951

expertmanofficial:
hell yeah, friendica icon is added!

PR #2474, opened 2023-11-16
feat: support Threads icon in profile links

shuuji3:
This adds Threads icon support in the links of the profile page.

partially implements #2472 (only Threads icon)


shuuji3:
It looks like I need additional changes to include the actual icon data.
Screenshot from 2023-11-17 00-52-35

shuuji3:
Threads icon was added in remix icon v3.5.0 but the current version was v3.4.0, which is installed via `@iconify/json`. Upgrading that version solved the issue.

patak-dev (approved):

PR #2475, opened 2023-11-17
fix: NavBottom fixed, not sticky

### LINKED ISSUE #1558: android firefox touch is offset by size of address bar
Hi,

gorgeous app, thanks! I was wondering how I can't hit the buttons at the bottom of screen, or bottom of toot input box. Then I noticed that touch placement is off by size of auto-hiding address bar of firefox.

So in my one+ 8 I use FF for web. I tried the latest version at main.elk.zone. I first noticed it while I tried hitting the language selection, I could not hit it. It always opened wrong view. After moment I realized I have to scroll page a bit to get address bar visible from FF, and then I manage to hit the button. I can also hit little above the button, and it works.

I believe the app doesn't calculate the size of screen, it is not realizing FF autohides the address bar.

zmc:
I could be missing something, but this fixes the bottom nav bar for me on mobile in Firefox, doesn't regress in Chrome, and doesn't affect the desktop layout.

resolve #1558

### DIFF
--- a/layouts/default.vue
+++ b/layouts/default.vue
@@ -54,7 +54,7 @@ const isGrayscale = usePreferences('grayscaleMode')
         <div min-h="[calc(100vh-3.5rem)]" sm:min-h-screen>
           <slot />
         </div>
-        <div sticky left-0 right-0 bottom-0 z-10 bg-base pb="[env(safe-area-inset-bottom)]" transition="padding 20">
+        <div fixed left-0 right-0 bottom-0 z-10 bg-base pb="[env(safe-area-inset-bottom)]" transition="padding 20">
           <CommonOfflineChecker v-if="isHydrated" />
           <NavBottom v-if="isHydrated" sm:hidden />
         </div>


zmc:
> We tried this already and we needed to revert it, see:
> 
>     * [fix: mobile header and footer prevent flicker on ios safari #660](https://github.com/elk-zone/elk/pull/660)
> 
>     * [fix: Position NavBottom with "fixed" instead of "sticky" #2252](https://github.com/elk-zone/elk/pull/2252)
> 
> 
> It seems we need something else to avoid the regression

I see. Could you elaborate a bit on why it had to be reverted? I can only find the phrase "some issues with Android".

patak-dev:
I dont recall, I think there was flickering

zmc:
Is it possible the flickering was caused by something else? I have seen none with this branch on my Pixel 6.

edimitchel:
We should try this on safari, and ensure we don't have problem with the content which became hidden by the header.
See on the closed pr mentioned the behavior to avoid

PR #2483, opened 2023-11-27
feat: unify the style of `under construction` page

yrming:


pages/settings/profile/featured-tags.vue
```diff
@@ -20,10 +20,11 @@ useHydratedHead({
     </template>
     <div text-center mt-10>
       <h1 text-4xl>
+        <span sr-only>{{ $t('settings.profile.featured_tags.under_construction') }}</span>
         🚧
       </h1>
       <h3 text-xl>
-        {{ $t('settings.profile.featured_tags.label') }}
+        {{ $t('settings.profile.featured_tags.under_construction') }}
```
patak-dev, line 27: I think it is good to show "Under Construction" but the reference to the sub section is lost, so it would be better for the `label` to still be displayed.

yrming: I think what you said makes sense. Now that I have modified it, please review it again.

patak-dev:
Thanks!

PR #2485, opened 2023-11-28
feat: show emoji tooltip

Shinigami92:
Show tooltip on hover over emoji

### DIFF
--- a/composables/content-render.ts
+++ b/composables/content-render.ts
@@ -6,6 +6,7 @@ import { RouterLink } from 'vue-router'
 import { decode } from 'tiny-decode'
 import type { ContentParseOptions } from './content-parse'
 import { parseMastodonHTML } from './content-parse'
+import Emoji from '~/components/emoji/Emoji.vue'
 import ContentCode from '~/components/content/ContentCode.vue'
 import ContentMentionGroup from '~/components/content/ContentMentionGroup.vue'
 import AccountHoverWrapper from '~/components/account/AccountHoverWrapper.vue'
@@ -19,8 +20,11 @@ function getTextualAstComponents(astChildren: Node[]): string {
 }
 
 /**
-* Raw HTML to VNodes
-*/
+ * Raw HTML to VNodes.
+ *
+ * @param content HTML content.
+ * @param options Options.
+ */
 export function contentToVNode(
   content: string,
   options?: ContentParseOptions,
@@ -42,6 +46,17 @@ export function nodeToVNode(node: Node): VNode | string | null {
 
   if (node.name === 'mention-group')
     return h(ContentMentionGroup, node.attributes, () => node.children.map(treeToVNode))
+
+  // add tooltip to emojis
+  if (node.name === 'picture' || (node.name === 'img' && node.attributes?.alt)) {
+    const props = node.attributes ?? {}
+    props.as = node.name
+    return h(
+      Emoji,
+      props,
+      () => node.children.map(treeToVNode),
+    )
+  }
 
   if ('children' in node) {
     if (node.name === 'a' && (node.attributes.href?.startsWith('/') || node.attributes.href?.startsWith('.'))) {
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@ node_modules
 *.log
 dist
 .output
+.pnpm-store
 .nuxt
 .env
 .DS_Store
--- a/components/emoji/Emoji.vue
+++ b/components/emoji/Emoji.vue
@@ -1,0 +1,30 @@
+<script setup lang="ts">
+const { as, alt, dataEmojiId } = $defineProps<{
+  as: string
+  alt?: string
+  dataEmojiId?: string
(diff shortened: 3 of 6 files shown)


composables/content-render.ts
```diff
@@ -4,6 +4,7 @@ import { Fragment, h, isVNode } from 'vue'
 import type { VNode } from 'vue'
 import { RouterLink } from 'vue-router'
 import { decode } from 'tiny-decode'
+import { find as findEmoji } from 'node-emoji'
```
antfu, line 7: The only thing I am not sure is the introduction of this dep. This module supports to work on client side, and this package seems to be designed for node. I guess we already have some logic to extract emoji from a string that we could reuse.

Shinigami92: I also thought that, but in the end `node-emoji` is "just" a wrapper around `emojilib`, that is also used by `emoji-mart`
However, `emoji-mart` has a `getEmojiDataFromNative`, but this is async and also need to call `init` once somewhere before.
But `emoji-mart` is just the emoji-picker component and also I'm not sure if it is emoji-"complete"

all in all, we need somehow a lookup set to "translate" emojis into text, and we could do this manually by hand (feels overkill and very cumbersome) or we just use a lib like `node-emoji` (wrapper around `emojilib`)

I feel safe about `node-emoji` and I assume it just has node in its name, because it is used for nodejs/js 🤷

Shinigami92:
@antfu is there something else I should do before this can be merged?
Or should we ask someone else for another review?

composables/content-render.ts
antfu: If we could ensure only one `emojilib` gets bundled, then I am okay with having it.

antfu: Just checked Netlify, against the main branch, the full bundle size is

- this PR: `8.78 MB (1.73 MB gzip)`
- main: `8.45 MB (1.68 MB gzip)`

And the `entry.js`

- this PR: `763.66 kB │ gzip: 218.47 kB`
- main: `537.79 kB │ gzip: 168.74 kB`

I think I am ok with the full bundle adding 330KB, but it shouldn't go to the main entry. `(763 - 537) / 537 = 42%` increase of the initial entry bundle.

I guess we need to somehow make it lazy-loaded.

Shinigami92:
there is a type error I need to fix

Shinigami92:
@antfu fixed the ts-error and also added .pnpm-store to gitignore, because from my home-pc this folder is very big :wink:

composables/content-render.ts
```diff
@@ -6,6 +6,7 @@ import { RouterLink } from 'vue-router'
 import { decode } from 'tiny-decode'
 import type { ContentParseOptions } from './content-parse'
 import { parseMastodonHTML } from './content-parse'
+import Emoji from '~/components/emoji/Emoji.vue'
```
antfu, line 9: Maybe a more explicit name, like `EmojiTitleWrapper` or something?

Shinigami92: That feels wrong to me 😕 
Either is it a real wrapper, because it renders directly the component, nor is it potentially hard-related to the title
I tried to align the name with what already is inside Elk's components folder

So I could imagine that someone could add more features to emojis in the future, e.g. not using the `title` attribute, but translate it directly to plain text for e.g. a11y reasons and this is togglable via option/flag 🤷 

Also please note that the `title` can also be undefined if nothing matching was found (it's potentially a rar case), but so it is not a `TitleWrapper`

If you just want to satisfy the vue lint rule [multi-word-component-names](https://eslint.vuejs.org/rules/multi-word-component-names.html), why is it not active in the first place and also why do other components also don't follow that 🤔
Like e.g. `list/Account` or `Dropdown`

So what I could suggest and only comes into my mind, would be `EmojiContent`, but TBH I'm not sure if it is worth these 7 additional letters
What counts more for me is the feature itself
When I go to MS-Teams, GitHub, or nearly every other tool, I have this feature, and so I miss it in Elk

Shinigami92:
some types broke (again) due to the rebase of main... fixing 🧑‍🏭

composables/content-render.ts
Shinigami92: @antfu how should we go forward?

Shinigami92:
After 2 weeks of vacation, I updated the branch again
@danielroe would you like to have a look?

Shinigami92:
@elk-zone/team please let me know what I should do to get this PR merged
I don't want to update this branch over and over again

the dependency is only 60kb and lazy loaded, and these 60kb are extremely minimal in comparison to normal images or even videos
and if everything is configured normal, this chunk should even be cached on the browsers side
when I scroll through my timeline, I get very quickly around 30mb loaded

so what is stopping this from getting merged? please tell me and don't leave this PR just alone without any answers

danielroe (approved):

patak-dev (approved):

Shinigami92:
@patak-dev thank you :heart:

PR #2490, opened 2023-11-29
chore: update nuxt to v3.8.2

danielroe:


userquin:
uhmm, we should change nuxt hook for pwa generation

PR #2494, opened 2023-12-01
chore(pwa): change sw generation logic

userquin:
This PR includes:
- use Nuxt `nitro:build:public-assets` hook to generate the service worker: Nitro moving assets (`_nuxt/**`) and html pages after running pwa api to generate the service worker: all assets and html pages missing from sw precache manifest
- include logic to handle Nuxt Payload Extraction and App Manifest
- remove manifest transform from configuration: overriden by the configuration module
- Vite 5 support, Nuxt almost ready

Without this PR (new Nitro version with Nuxt 3.8.2 using `rollup:before` hook, `Generate public .output/public` must run before sw generation):

imagen

With this PR:

imagen

/cc @patak-dev


userquin:
rn main is broken, go to any settings page, activate offline and refresh page (remember to update Elk)

imagen

imagen

patak-dev:
Ok, let's merge and iterate if needed then

patak-dev (approved):

PR #2498, opened 2023-12-02
chore(pwa): improve DX for `$pwa`

userquin:
This PR includes:
- add types dts for pwa injections
- update plugin signature to use types dts
- register types dts in `prepare:types` Nuxt hook
- register `vite-plugin-pwa/vue` instead `vite-plugin-pwa/client`

Daniel, it seems we've a new `vue-tsc 1.8.24`, still using `1.8.8` and TypeScript `5.1.6`, maybe we should update both to latest.

Volar in JetBrains IDE should now work in ts modules and also in script setup and template in Vue SFC:

imagen


modules/pwa/index.ts
```diff
@@ -56,8 +56,10 @@ export default defineNuxtModule<VitePWANuxtOptions>({
     else {
       // Register PWA types
       nuxt.hook('prepare:types', ({ references }) => {
+        const types = resolver.resolve('runtime/types')
+        references.push({ path: resolver.resolve(nuxt.options.buildDir, types) })
```
danielroe, line 60: why the double resolve here? shouldn't this just be `resolver.resolve('runtime/types')`?

modules/pwa/runtime/pwa-plugin.client.ts
```diff
@@ -105,7 +110,7 @@ export default defineNuxtPlugin(() => {
 
   return {
     provide: {
-      pwa: reactive({
+      pwa: reactive<PwaInjection>({
```
danielroe, line 113: How about instead using `satisfies PwaInjection`?

modules/pwa/runtime/pwa-plugin.client.ts
```diff
@@ -1,7 +1,12 @@
 import { useRegisterSW } from 'virtual:pwa-register/vue'
+import type { UnwrapNestedRefs } from 'vue'
+import type { PwaInjection } from './types'
 import { STORAGE_KEY_PWA_HIDE_INSTALL } from '~/constants'
+import type { Plugin } from '#app'
 
-export default defineNuxtPlugin(() => {
+const plugin: Plugin<{
```
danielroe, line 7: At least we should add a todo comment here that we need to investigate why the injection isn't working in JetBrains. 🤔

In general I'd rather see us fix the underlying issue (in Nuxt, if there is one), or report it to JetBrains so they can fix it rather than requiring users to manually annotate types.

modules/pwa/index.ts
userquin: upps

userquin: this may be the problem, testing on local we don't need the type in the plugin (ofc, the module is in local, I'll review Nuxt PWA module)

modules/pwa/runtime/pwa-plugin.client.ts
userquin: removed

PR #2506, opened 2023-12-10
feat(i18n): update vi-VN.json

mastoduy:
Add new translations
- Copy account address
- Embbed media player
- Correct "Thread"


mastoduy (commented):
Update "Home timline"

patak-dev (approved):

PR #2524, opened 2023-12-21
fix: correct linking of theme colors

maybeanerd:
This makes the theme color mapping work as it should IMO. The current implementation just worked because the `key` and `--theme-color-name` happen to be the same, which I think is an assumption we can't make forever.

In our skin of elk, we changed the name of a theme and it was necessary to do these changes (https://github.com/maybeanerd/crab/pull/3).

This should enable future named themes for elk in general without breaking anything.


patak-dev (approved):

patak-dev:
Thanks! 🧡

PR #2525, opened 2023-12-21
chore: run unit tests in watch mode by default

maybeanerd:
The current test script waits for changes, which is not what we want to be happening in CI/CD

This makes it run only once, but adds a `test:watch` for development


maybeanerd:
I just noticed that according to the vitest docs this should be the automatic behavior already (https://vitest.dev/guide/cli#vitest), but with the CI/CD on GitHub actions it seems to not work?
Does this also happen on this repo at times, or only in our fork?

maybeanerd:
Looks like it's also an issue here: https://github.com/elk-zone/elk/actions/runs/7294167547/job/19878550089

package.json
```diff
@@ -19,9 +19,11 @@
     "typecheck": "stale-dep && nuxi typecheck",
     "prepare": "ignore-dependency-scripts \"tsx scripts/prepare.ts\"",
     "generate": "nuxi generate",
-    "test:unit": "stale-dep && vitest",
+    "test:unit:watch": "stale-dep && vitest",
```
patak-dev, line 22: I think it is better to avoid the `watch` here, and follow the convention from Vitest that `test` / `vitest` watchs by default.

package.json
```diff
@@ -19,9 +19,11 @@
...
+    "test:unit:ci": "stale-dep && vitest run",
     "test:typecheck": "stale-dep && vue-tsc --noEmit && vue-tsc --noEmit --project service-worker/tsconfig.json",
-    "test": "nr test:unit",
+    "test": "nr test:unit:ci",
```
patak-dev, line 25: We should leave `test` as it was, but call `test:ci` from the workflow for CI

danielroe (approved):

PR #2530, opened 2023-12-26
feat: upgrade to masto.js v6

patak-dev:
Check out [the migration guide to see what's new in masto.js v6](https://github.com/neet/masto.js/releases/tag/v6.0.0)

The new rest API is quite nice (most of the changes in the PR are related to it). And the way streaming works also changed significantly. As far as I can tell, everything is working as expected (except for search, see https://github.com/neet/masto.js/issues/1018. Pushing the PR as a draft until the fixes for are available as a new release).

### DIFF
--- a/components/account/AccountHeader.vue
+++ b/components/account/AccountHeader.vue
@@ -53,7 +53,7 @@ function previewAvatar() {
 async function toggleNotifications() {
   relationship!.notifying = !relationship?.notifying
   try {
-    const newRel = await client.v1.accounts.follow(account.id, { notify: relationship?.notifying })
+    const newRel = await client.v1.accounts.$select(account.id).follow({ notify: relationship?.notifying })
     Object.assign(relationship!, newRel)
   }
   catch {
@@ -97,7 +97,7 @@ async function editNote(event: Event) {
   if (relationship.note?.trim() === newNote.trim())
     return
 
-  const newNoteApiResult = await client.v1.accounts.createNote(account.id, { comment: newNote })
+  const newNoteApiResult = await client.v1.accounts.$select(account.id).note.create({ comment: newNote })
   relationship.note = newNoteApiResult.note
   personalNoteDraft.value = relationship.note ?? ''
 }
--- a/components/notification/NotificationCard.vue
+++ b/components/notification/NotificationCard.vue
@@ -15,7 +15,6 @@ const { notification } = defineProps<{
           ps-3 pe-4 inset-is-0
           rounded-ie-be-3
           py-3 bg-base top-0
-          :lang="notification.status?.language ?? undefined"
         >
           <div i-ri-user-3-line text-xl me-3 color-blue />
           <AccountDisplayName :account="notification.account" text-primary me-1 font-bold line-clamp-1 ws-pre-wrap break-all />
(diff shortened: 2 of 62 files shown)


components/account/AccountHeader.vue
```diff
@@ -53,7 +53,7 @@ function previewAvatar() {
 async function toggleNotifications() {
   relationship!.notifying = !relationship?.notifying
   try {
-    const newRel = await client.v1.accounts.follow(account.id, { notify: relationship?.notifying })
+    const newRel = await client.v1.accounts.$select(account.id).follow({ notify: relationship?.notifying })
```
Shinigami92, line 56: Why is there a `?` behind `relationship`, while on line 54 it is specifically given?
If `relationship` would be nullish on line 54, we would already run into a runtime error and never reach line 56

components/notification/NotificationCard.vue
```diff
@@ -15,7 +15,6 @@ const { notification } = defineProps<{
           ps-3 pe-4 inset-is-0
           rounded-ie-be-3
           py-3 bg-base top-0
-          :lang="notification.status?.language ?? undefined"
```
Shinigami92, line 18: Could you explain why this got removed? also below

Shinigami92 (commented):
generally LGTM

userquin:
We need to review types and logic in service-worker (@patak-dev I've sent you private DM in discord)

components/notification/NotificationCard.vue
patak-dev: A `follow` notification never has a `status`, same for the others were it was removed. The types in masto v6 may better reflect that and it was erroring out after the upgrade.

components/account/AccountHeader.vue
patak-dev: Ya, line 54 also doesn't need it on the right side. We can do a PR to remove this later as it is unrelated to the upgrade.

patak-dev:
> We need to review types and logic in service-worker (@patak-dev I've sent you private DM in discord)

The line you shared in Discord seems ok after the upgrade as it is a direct call. It doesn't use masto. If you could check anyways that things looks good on the service-worker/pwa side after this PR, appreciated!

userquin:
@patak-dev it seems to be fine: 

imagen

userquin (approved):
sw logic seems to be fine

patak-dev:
Search is fixed now. Tests are failing because of a type error on `AbortSignal.timeout`, but I think it may be an issue with Vitest? Node 18 should have that function

Shinigami92 (approved):

patak-dev:
> Search is fixed now. Tests are failing because of a type error on `AbortSignal.timeout`, but I think it may be an issue with Vitest? Node 18 should have that function

Fixed this one by [polyfilling `AbortSignal.timeout` during test setup](https://github.com/elk-zone/elk/pull/2530/commits/e0e4ce1bd6aa0fbc7763e6233a2bc06041e19339).

I think we should be good to go now.

patak-dev:
Let's merge this one so we try it out in https://main.elk.zone

shuuji3:
Thanks for this upgrade. I can proceed #2530 with this new version. 👍🏻

PR #2541, opened 2024-01-02
feat: add LTR/RTL in hashtags and mentions support

### LINKED ISSUE #2540: LTR/RTL support for tags
From this toot: https://elk.zone/m.webtoo.ls/@bebatjof@strangeobject.space/111670808929246472

We should:
- add RTL text direction detection
- change content-render module to change tags anchors containing RTL text: add `dir="rtl"` attribute
- change tags header title page

userquin:
Right now only hashtags, mentions (no idea if an account can have arabic text), tags and explore/tags pages, we can check later body support.

This PR using [bdi](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/bdi) tag, RTL text detection has been removed, there is no need.

closes #2540


utils/rtl-text-detection.ts
```diff
@@ -0,0 +1,18 @@
...
+  '\uFB50-\uFDFF',
+  '\uFE70-\uFEFF',
+  /* hebrew range */
+  '\u05D0-\u05FF',
+].join('')
+
+const reRTL = new RegExp(`[${rtlChars}]`, 'g')
+
+export function isRtlText(text: string) {
+  const textCount = text.replace(/[0-9\s\\\/.,\-+="']/g, '').length
```
danielroe, line 14: Can we make this more performant by first checking if there are any RTL characters and early returning if not?

userquin: done

userquin:
We should change also `component/tag/TargCard.vue` SFC

composables/content-render.ts
```diff
@@ -1,7 +1,6 @@
-import { TEXT_NODE } from 'ultrahtml'
-import type { Node } from 'ultrahtml'
-import { Fragment, h, isVNode } from 'vue'
+import { ELEMENT_NODE, type ElementNode, type Node, TEXT_NODE } from 'ultrahtml'
 import type { VNode } from 'vue'
+import { Fragment, h, isVNode } from 'vue'
```
danielroe, line 3:
```suggestion
import { TEXT_NODE } from 'ultrahtml'
import type { Node } from 'ultrahtml'
import { Fragment, h, isVNode } from 'vue'
import type { VNode } from 'vue'
```

danielroe (commented):
this looks perfect 👌 

If you wanted, a test in https://github.com/elk-zone/elk/blob/74138a9a586c7af88e1b839dbae863b9333ca07e/tests/nuxt/content-rich.test.ts#L239 would be the icing on the cake

composables/content-render.ts
```diff
@@ -1,5 +1,5 @@
-import { TEXT_NODE } from 'ultrahtml'
-import type { Node } from 'ultrahtml'
+import { ELEMENT_NODE , TEXT_NODE } from 'ultrahtml'
```
danielroe, line 1:
```suggestion
import { ELEMENT_NODE, TEXT_NODE } from 'ultrahtml'
```

danielroe (approved):

userquin:
Adding the test found a testing bug, I'll check it again later.

We also need to add a test containing `<a><bdi>...`.

userquin:
The `RouterLink` mock component wasn't rendering the children, that's why the bdi and its content not being rendered (in fact any anchor content in any test).

I added the default slot and moved all inlined snapshots to the snapshot file, now all anchor rendered properly.

There is still an issue with the bdi parsing test, ultrahtml is able to parse the tree properly but but rendering it doesn't work, the children not being rendered:
- the test: https://github.com/elk-zone/elk/pull/2541/files#diff-af4f073a117db5361f7d5fa6aa2a5b6d490fc124506185bf8d49379801198f94R158
- test output: https://github.com/elk-zone/elk/pull/2541/files#diff-05d510870be534ed8ce54b3d5dd03b681aff1f6a0fa3c3f5e47b4afed39c77daR138-R150

If I included this code [here](https://github.com/elk-zone/elk/blob/main/composables/content-parse.ts#L138): 
```js
  function traverse(node: Node, level = 0) {
    if (node.type === ELEMENT_NODE || node.type === DOCUMENT_NODE) {
      console.log(' '.repeat(level * 2), node.name)
      if (node.children)
        node.children.forEach(n => traverse(n, level + 1))
    }
    else if (node.type === TEXT_NODE && node.value) {
      console.log(' '.repeat(level * 2), node.value)
    }
  }
  traverse(parse(html))

```

the `bdi` content is there.

userquin:
Mentions:

imagen

Hashtags:

imagen

danielroe (approved):

PR #2543, opened 2024-01-03
chore: update nuxt to v3.9.3 and vue 3.4.15

userquin:
This PR also updates some nuxt modules to latest version, changed also configuration for new i18n module.

Right now it is not working, I guess some problem with vue-macros, the paginator watcher not triggering `loadNext`, adding immediate works but the browser hangs after a few seconds or when scrolling.

We also need to review nuxt security module, added a new patch (when this PR ready we can remove old patch).


page-lifecycle.d.ts
```diff
@@ -2,15 +2,15 @@ declare module 'page-lifecycle/dist/lifecycle.mjs' {
...
     }
     interface PageLifecycle extends EventTarget {
-        get state(): PageLifecycleState
-        get pageWasDiscarded(): boolean
-        addUnsavedChanges: (id: Symbol | any) => void
-        removeUnsavedChanges: (id: Symbol | any) => void
-        addEventListener: (type: string, listener: (evt: PageLifecycleEvent) => void) => void
+      get state(): PageLifecycleState
+      get pageWasDiscarded(): boolean
+      addUnsavedChanges: (id: symbol | any) => void
```
edimitchel, line 11: Why not Symbol?

nuxt.config.ts
```diff
@@ -84,6 +89,17 @@ export default defineNuxtConfig({
     build: {
       target: 'esnext',
     },
+    optimizeDeps: {
```
edimitchel, line 92: This change looks like a problem on the compile-time schedule, and probably a regression from Nuxt with the ability to create dependency between modules..

danielroe: Any change here is more likely a change in Vite (as we upgraded from Vite 4 -> Vite 5 in v3.9).

userquin: I included the list for testing (testing with current main and this PR on my local), I need to remove all entries, we only need focus trap for client

patak-dev: I think it is a good idea to include most of these in the config. We're lazy loading a lot of dependencies and they aren't discovered by the scanner. Some will end up triggering a full-reload on cold start. We may merge a new strategy in Vite 5.1 that we can adopt and this will be even more important then (we will start getting warnings to add the deps to the config). See:
- https://github.com/vitejs/vite/pull/15244

userquin: has anyone in the team tried running dev in a clean clone (main branch with Nuxt 3.8)? there are almost 10 page refresh even when scrolling on timeline

@patak-dev I'm going to activate Vite 5 `{ server: { fs: cachedChecks: true } }` ;)

patak-dev: @userquin see https://github.com/nuxt/nuxt/issues/24196#issuecomment-1853505843, there were issues in Nuxt 3.8 preventing the optimized deps cache to be used (so every startup was a cold start). This should be fixed in 3.9 (I imagine also in newer patches for 3.8?)

userquin: I'm going to include also the new floating-vue v5.0.0 :fingers_crossed:

userquin: @patak-dev I don't see any improvement activating `cachedChecks: true` on my laptop (with or without optimized deps, just a few ms, in both cases removing the `.cache` folder):

```shell
# without cachedChecks: true
ℹ Vite server warmed up in 8133ms                                                                                                                     15:02:18
ℹ Vite client warmed up in 12612ms                                                                                                                    15:02:19
✔ Nitro built in 5417 ms

# with cachedChecks: true
ℹ Vite server warmed up in 8190ms                                                                                                                     15:03:23
ℹ Vite client warmed up in 12484ms                                                                                                                    15:03:24
✔ Nitro built in 5196 ms

userquin:
Closing, as #2610 has now been merged.

PR #2547, opened 2024-01-06
fix: can't refresh statuses after delete a status in account/index pa…

### LINKED ISSUE #2546: Can't refresh statuses after deleting a status
Look these gif
Recording 2024-01-06 at 14 50 16
There are two bugs, after we switch from home page into account page, we delete a status, the timeline have not refreshed, and when we switch back into home page, the deleted status still rendered after expanding the show 1 item text

HawtinZeng:
fixed #2546


HawtinZeng:
@danielroe help me to review this PR, thank u!

EddiesTech:
Awesome work and a good spot!

HawtinZeng:
> Awesome work and a good spot!

Can u review this PR and merge it into main?

composables/paginator.ts
```diff
@@ -12,6 +12,7 @@ export function usePaginator<T, P, U = T>(
   // called `next` method will mutate the internal state of the variable,
   // and we need its initial state after HMR
   // so clone it
+
```
patak-dev, line 15:
```suggestion
```

patak-dev (approved):
Thanks for the fix!

patak-dev:
Why was this closed @HawtinZeng? Reopening it as I think we should get this fix :)

pages/[[server]]/@[account]/index/index.vue
```diff
@@ -21,11 +21,12 @@ if (account) {
     title: () => `${t('account.posts')} | ${getDisplayName(account)} (@${account.acct})`,
   })
 }
+const stream = $(useStreaming(client => client.v1.stream.streamUser()))
```
Comment (patak-dev) [inline pages/[[server]]/@[account]/index/index.vue:24]: This should use the new API now, as we merged masto.js v6. See
- https://github.com/elk-zone/elk/pull/2530
```suggestion
const stream = useStreaming(client => client.user.subscribe()))
```

pages/[[server]]/@[account]/index/index.vue
```diff
@@ -21,11 +21,12 @@ if (account) {
...
   })
 }
+const stream = useStreaming(client => client.user.subscribe())
 </script>
 
 <template>
   <div>
     <AccountTabs />
-    <TimelinePaginator :paginator="paginator" :preprocess="reorderAndFilter" context="account" :account="account" />
+    <TimelinePaginator :paginator="paginator" :preprocess="reorderAndFilter" context="account" :account="account" :stream="stream" />
```
Comment (patak-dev) [inline pages/[[server]]/@[account]/index/index.vue:30]: Re-checking the PR. Is this change needed? Isn't the user stream unrelated to the `[[server]]/@[account]` here?

HawtinZeng:
> Why was this closed @HawtinZeng? Reopening it as I think we should get this fix :)

Because I saw u have approved this PR.

pages/[[server]]/@[account]/index/index.vue
Comment (HawtinZeng) [reply pages/[[server]]/@[account]/index/index.vue:30]: Sorry, I will check the API

Comment (HawtinZeng) [reply pages/[[server]]/@[account]/index/index.vue:30]: I have checked this, I don't think user stream is unrelated to [[server]]/@[account], when the user delete a status in his/her account, we should delete the same status in user stream(which will) be shown in home timeline page.Maybe it looks wired...

Comment (HawtinZeng) [reply pages/[[server]]/@[account]/index/index.vue:30]: The perfact way is that we can use access the same stream instance in the home timeline, but I have no idea to get access to home steam...

Comment (HawtinZeng) [reply pages/[[server]]/@[account]/index/index.vue:30]: Maybe we can use pinia or attach the stream into Vue prototype,

Issue #2464, opened 2023-11-10
handle of other person removed from post


sloanlance:
See the post at: https://elk.zone/mastodon.social/@sloanlance/111386832778347376
Compare with the post at: https://mastodon.social/@sloanlance/111386832778347376

Notice that the post at mastodon.social includes the handle of the person to whom the reply is addressing.

The original text of the post…

> `@RhetTbull@fosstodon.org, a mask requirement is a reasonable caution. COVID-19 still exists & is spreading for various reasons. @pycon@fosstodon.org wants to avoid liability & their event becoming a super-spreader.`

However, on elk.zone, that handle is removed.  See the screenshot below…

<img width="792" alt="Screenshot 2023-11-10 at 10 03 05" src="https://github.com/elk-zone/elk/assets/17595351/3bfff700-0c62-40c0-b68d-46be5732c3cb">

Elk still has the handle of the user.  If I use Elk to edit my post, the handle is shown in the editor.  However, the handle isn't shown after the post is saved.

HawtinZeng:
I am trying to fix this bug, if you are interested in this bug, let me know please.

PR #2548, opened 2024-01-07
fix: disable auto collapse of the first mentioned account in reply content issue: #2464

HawtinZeng:


HawtinZeng:
@patak-dev hi, patak, I have saw that your wrote the code related this issue, can u help me to review my code?

HawtinZeng:
#2464 2464

patak-dev:
Not showing the mentions cleans up quite a bit of noise when reading a thread. Let's try to keep the feature. I think we could make it a bit more strict and take into account the text. We could define for example that if the text starts with `,` then we should show the mention.

HawtinZeng:
> Not showing the mentions cleans up quite a bit of noise when reading a thread. Let's try to keep the feature. I think we could make it a bit more strict and take into account the text. We could define for example that if the text starts with `,` then we should show the mention.

That sounds good, i will commit my changes soon.

PR #2561, opened 2024-01-16
feat: add new setting to disable blur for low-performance device

### LINKED ISSUE #2529: Add a setting to disable background blur filter for modals
On browsers without hardware acceleration, the blur effect that applies on the backdrop of modals (eg. opening a post) causes the tab with Elk to run with very poor performance. The effect is especially worsened with multiple instances of the blur filter (eg. opening your toot, then also opening the image alt. text editor on top). I've tested that disabling the blur filter attributes on the modal backdrops solves the slowdown. 

There should be a setting that replaces this effect with a less graphically intensive option, such as a transparent black background, for example.

shuuji3:
resolve #2529

This PR adds a new setting to allow disabling the background blur (default is enabled) for the low-performance device.

Screenshot from 2024-01-16 17-26-20

## How to test
1. Uncheck "Enable background blur" on the settings page: https://deploy-preview-2561--elk-zone.netlify.app/settings/preferences
2. Try, for example, to mute some users to see the confirmation dialog without the background blur.


patak-dev (changes requested):
I think we better call this one `Optimize for low-performance device", defaulting to false, and initially have this background blur removal, but we could add other things like the glass effect in the top bar (if we still have that) later on, etc.

shuuji3:
Yeah, that's a better expression of this option. The current name was a bit too specific and difficult for normal users to understand.

> the glass effect in the top bar

This was also implemented by blur CSS so already covered by this line: https://github.com/elk-zone/elk/pull/2561/files#diff-8047d450d250724e6d8a943121218d82f64e307a9055744a041d159325df2d68L29 but yes, let's add other switches in this group if we found another performance bottleneck later.

patak-dev (approved):
Thanks for the PR and the changes! Looks great!

Issue #2554, opened 2024-01-10
Dev error messages on Notification screen


pjohanneson:
I just updated to https://github.com/elk-zone/elk/commit/0451ac98c9860b0ccb2a582b9ab17cea4624ea10 and now I'm seeing this message on the Notifications screen when I click on the "Show {n} New Notifications".

![Screenshot of the error message, which reads "DEV: MISSING notification.type '' "](https://github.com/elk-zone/elk/assets/21602/4f117aa9-c4f1-4302-a15b-b0347a354537)

Shinigami92:
I scrolled through my notifications and didn't see any of such errors 🤔 
Can you provide more info e.g. by looking into the browser dev tools network tab and search for the notification that potentially should be rendered instead?

patak-dev:
We upgraded to masto.js v6, it may be an issue with the Mastodon version or fork. What version is your server running?

pjohanneson:
@Shinigami92 Looking at the Network tab, when I click the "Show {n} New Notifications", all the requests I make seem to be cached. I'm not sure I'm looking in the right place, though.

@patak-dev I'm on https://mstdn.ca/about, which is running "v4.3.0-alpha.0+glitch".

Updating to https://github.com/elk-zone/elk/commit/77175416a66ecca62d5a7c03004fd408e1d471fc hasn't fixed the problem. However, the issue *isn't* happening on http://elk.zone/, just on https://main.elk.zone/ .

Shinigami92:
I get this when I'm on the notification tab and while I'm there a new message is loaded but from the normal timeline.

patak-dev:
I can't reproduce this one, so we'll need others that can reproduce it to investigate. We will need to fix it before we can release main.elk.zone to elk.zone

haayman:
This indeed happens everytime there's a new toot for the main timeline.
I can't find anything in the network tab where this happens. I see one websocket but no activity there.
There are some errors on startup that may be the cause of this?

![image](https://github.com/elk-zone/elk/assets/683823/842c10aa-d571-4c4c-b1f1-4c89f5f1d8b9)
![image](https://github.com/elk-zone/elk/assets/683823/1f594e8e-724c-4437-b4c0-5f02a79acb50)

haayman:
I'm running in dev mode now and see extra error messages when I click 'Show {n} new items' under Notifications.
I don't know whether this is related. The problem is more that the new items shouldn't be on the Notifications tabs in the first place I suppose 
[...]

PR #2564, opened 2024-01-21
refactor: fix lint warnings

shuuji3:
This refactoring includes two types of fixes:
- simply fix lint warnings inside code comments [`e936b59` (#2564)](https://github.com/elk-zone/elk/pull/2564/commits/e936b595adf6a39c26bfbe4c9de8f388426912c2)
- remove the `eventType` param from `usePaginator` [`0ef77a1` (#2564)](https://github.com/elk-zone/elk/pull/2564/commits/0ef77a12a8da8bf7162d13e433acfb383532bfbe)

Here are all warnings raised from `pnpm lint`:
https://github.com/elk-zone/elk/actions/runs/7568568639/job/20610027245#step:8:1

It looks like `eventType` was no longer used since masto.js v6 update (https://github.com/elk-zone/elk/commit/6c5bb83ac32aecb5868a0e47ed9bcd2195a9d706#diff-0bb6139d018feba7b2e65911186183e5a06a771c4baf602f0d985dd066374973R33). As I don't fully understand the difference yet, I'd like to ensure if it's ok to remove this param completely.

### DIFF
--- a/components/common/CommonPaginator.vue
+++ b/components/common/CommonPaginator.vue
@@ -10,7 +10,6 @@ const {
   stream,
   keyProp = 'id',
   virtualScroller = false,
-  eventType = 'update',
   preprocess,
   endMessage = true,
 } = defineProps<{
@@ -18,7 +17,6 @@ const {
   keyProp?: keyof T
   virtualScroller?: boolean
   stream?: mastodon.streaming.Subscription
-  eventType?: 'notification' | 'update'
   preprocess?: (items: (U | T)[]) => U[]
   endMessage?: boolean | string
 }>()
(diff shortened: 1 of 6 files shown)


patak-dev (changes requested):
Nice! It looks like you found what is the issue with:
- https://github.com/elk-zone/elk/issues/2554

When I did the upgrade, I introduced a regression here:
https://github.com/elk-zone/elk/commit/6c5bb83ac32aecb5868a0e47ed9bcd2195a9d706#diff-0bb6139d018feba7b2e65911186183e5a06a771c4baf602f0d985dd066374973L35

The `eventType` is needed, and we have to use it to filter the event types that the paginator is going to deal with. Before, we only processes `update` or `notification` depending on that parameter, and after the masto.js v6 PR, we started always processing both types in all paginators.

Would you revert the `eventType` changes from this PR so we only merge the other fixes?

And if you'd like, you could send another PR to fix #2554. If not, I'll send that later.

shuuji3:
Ah, I see! I've reverted the second commit. Could you take a look #2554 as I'm not familiar with this part?

patak-dev:
Yes, thanks again for poking into this and finding the issue. I'll send a PR to fix it later

patak-dev (approved):

PR #2565, opened 2024-01-21
fix(suggestion): allow case-insensitive emoji suggestion

### LINKED ISSUE #2552: Emoji Picker is case sensitive
The current emoji picker that is brought up when typing `:` followed by any word is case sensitive. This makes it quite confusing to use as you'll think any emote where the case changes has been removed from your homeserver, when infact it's just that you've spelled it all lowercase expecting the search to find it.

## Expectation
Typing an emoji's name no matter the case should return appropriate results

## Result
Typing an emoji's name will only work if you type it case-sensitive

## Related info
Elk Version: [v0.10.3 (7ec76ff@release)](https://github.com/elk-zone/elk/commit/7ec76ffed95090c36d7de5038f9f6e301fd7ec75)
Browser: LibreWolf 121.0-1 Linux-Flathub

shuuji3:
fix #2552

Currently, we're simply comparing strings with `startsWith()` so users have to type the exact same case as its emoji id or shortcode.

This change supports a case-insensitive search despite either id/shortcode or user query containing any upper/lower cases.

Screenshot from 2024-01-21 17-19-08
Screenshot from 2024-01-21 17-14-32

Also, it looks like Mastodon's custom emoji search doesn't distinguish cases of shortcode: https://github.com/mastodon/mastodon/pull/15738


composables/tiptap/suggestion.ts
```diff
@@ -64,10 +64,10 @@ export const TiptapEmojiSuggestion: Partial<SuggestionOptions> = {
 
     const emojis = await import('@emoji-mart/data')
       .then(r => r.default as EmojiMartData)
-      .then(data => Object.values(data.emojis).filter(({ id }) => id.startsWith(query)))
+      .then(data => Object.values(data.emojis).filter(({ id }) => id.toLowerCase().startsWith(query.toLowerCase())))
 
     const customEmojis: CustomEmoji[] = currentCustomEmojis.value.emojis
-      .filter(emoji => emoji.shortcode.startsWith(query))
+      .filter(emoji => emoji.shortcode.toLowerCase().startsWith(query.toLowerCase()))
```
patak-dev, line 70: We should cache `query.toLowerCase()` to avoid recomputing it during the loops.
But I think we should do the same with `shortcode` and `id`, and cache the result of `import('@emoji-mart/data')` with the data as we needed in lowercase. This will also speed up things

shuuji3: OK, I removed multiple calls of `toLowerCase()` for `query`.

Also, I used `useAsyncData` to cache emojimart data import. Is this the right approach for caching?

Regarding making `id` and `shortcode` lowercase, there is an issue. If we lower the shortcode beforehand, we cannot show the original shortcode for users like the below.

Screenshot from 2024-01-21 18-29-35

This is problematic since some custom emoji use camelcase naming like `:thisIsCamelCaseEmojiName:`. The lowercased name `:thisiscamelcaseemojiname:` is a bit difficult to read.

Maybe we can keep as is since I assume it's not so heavy computation to call `toLowerCase()` as many times as emoji counts.

patak-dev (approved):

PR #2574, opened 2024-01-26
feat(i18n): Update vi-VN.json

mastoduy:
Added "Lock" and revamped some words


mastoduy (commented):
Add missing word "hidden" when typing

PR #2580, opened 2024-01-28
fix: remove the comma after collaped or grouped mentions #2464

HawtinZeng:


userquin:
can you send another PR for the `no-cors` in the request?

HawtinZeng:
> can you send another PR for the `no-cors` in the request?

Hi, after check the docs about `no-cors`, I think it's wrong to use it send cors request, `no-cors` just filter out all cors request. So I deletet it.

Issue #2601, opened 2024-02-21
User profile panel is not shown when hovering


shuuji3:
Tested version: Canary (https://github.com/elk-zone/elk/commit/b9394c2fa54d3899dd79bcc808cfc2af4245c144)

When hovering on the `@username` link, it doesn't show the profile panel but console log the following error:

![Screenshot from 2024-02-22 00-12-27](https://github.com/elk-zone/elk/assets/1425259/c962961a-c68a-44b7-b7fc-52590c1fa7e3)

```js
entry.1f2092ce.js:1 TypeError: Cannot read properties of undefined (reading 'includes')
    at Fa (entry.1f2092ce.js:8:13373)
    at sl (entry.1f2092ce.js:8:13539)
    at qs (entry.1f2092ce.js:8:13749)
    at Proxy.<anonymous> (entry.1f2092ce.js:52:16681)
    at vl (entry.1f2092ce.js:1:17142)
    at wu.z [as fn] (entry.1f2092ce.js:1:52929)
    at wu.run (entry.1f2092ce.js:1:4872)
    at U.b.update (entry.1f2092ce.js:1:53208)
    at U (entry.1f2092ce.js:1:53234)
    at ee (entry.1f2092ce.js:1:52047)
```

shuuji3:
`<AccountHoverCard>`'s `account` prop expects a value of `mastodon.v1.Account`.

However, the current type is `ComputedRef<mastodon.v1.Account | Ref<mastodon.v1.Account | null> | undefined >` here: https://github.com/elk-zone/elk/blob/main/components/account/AccountHoverWrapper.vue#L14

This wrong typed value is being passed into `getFullHandle()` and `account.acct` become `undefined` at the end here: https://github.com/elk-zone/elk/blob/8d792d003ddad681e7b2f680d30a0128561ebd46/composables/masto/account.ts#L29

shuuji3:
This issue had already fixed in the Canary version.

PR #2607, opened 2024-02-23
feat: bump to latest vue 3.4.19

userquin:
This PR includes:
- bump to vue 3.4.19
- refactor `AccountHoverWrapper.vue`
- extract types exported from `CommonRouteTabs.vue` to `~/types` and update all components using those types
- custom `useI18n` to provide `isHydrated`  variants of `t`, `d` and `n` 
- custom  `$t`, `$d` and `$n` to provide `isHydrated`  variants
- ~~include `isHydrated` logic in all pages and components~~: reverted some `isHydrated` since we have included the custom versions

We need to review and update all `isHydrated`, `t`, `d`, `n`, `$t`, `$d` and `$n` usages in vue sfc and ts modules: will be included in another PR.


nuxt.config.ts
```diff
@@ -86,6 +86,20 @@ export default defineNuxtConfig({
...
+        '@tiptap/extension-text', '@tiptap/extension-mention', '@tiptap/extension-hard-break',
+        '@tiptap/extension-bold', '@tiptap/extension-italic', '@tiptap/extension-code',
+        '@tiptap/extension-history', 'prosemirror-state', 'browser-fs-access', 'blurhash',
+        '@vueuse/integrations/useFocusTrap', '@tiptap/extension-code-block', 'prosemirror-highlight',
+        '@tiptap/core', 'tippy.js', 'prosemirror-highlight/shiki', '@fnando/sparkline',
+        '@vueuse/gesture', 'github-reserved-names', 'file-saver', 'slimeform', 'vue-advanced-cropper',
+        'workbox-window', 'workbox-precaching', 'workbox-routing', 'workbox-cacheable-response',
+        'workbox-strategies', 'workbox-expiration',
+      ],
+    },
```
patak-dev, line 102: We could probably remove the ones that would be detected by the scanner, but I think it is ok for now to include all of these 👍🏼

patak-dev:
If `isHydrated` is going to be needed for every `$t`, should we have that condition directly inside `$t`? (wrapping the original `$t`). If only is needed in some `$t`, then I think it is worth to have a a new `$ht` that does this so the code is more maintainable.

nuxt.config.ts
userquin: Some dependencies only detected when matching the case: for example, github-reserved-names and file-saver when visiting the settings > users and about pages

When visiting  a new page and found a new entry in the console I just add them to the list.

imagen

patak-dev: perfect 👍🏼

components/account/AccountHoverWrapper.vue
```diff
@@ -15,10 +14,10 @@ const props = defineProps<{
   disabled?: boolean
 }>()
 
+const resolving = ref(false)
```
patak-dev:
```suggestion
```

userquin: already removed in last commit

userquin:
> If `isHydrated` is going to be needed for every `$t`, should we have that condition directly inside `$t`? (wrapping the original `$t`). If only is needed in some `$t`, then I think it is worth to have a a new `$ht` that does this so the code is more maintainable.

To make that change we need to remove all imports, including `const { t } = useI18n()` in all pages/components and then change all `$t()` usages in sfc templates with `t()`.

imagen

patak-dev (approved):
Looks great! Thanks for the hard work to make this happen @userquin 👏🏼 

@danielroe I think we should merge it. I'll leave it open for a bit in case you would like to check it out. We can remove other `isHydrated` conditionals in a future PR (I see there are some already removed here, and it is ok). The Nuxt update should be a manageable PR now.

composables/paginator.ts
```diff
@@ -13,7 +13,7 @@ export function usePaginator<T, P, U = T>(
   // so clone it
   const paginator = _paginator.clone()
 
-  const state = ref<PaginatorState>(isHydrated.value ? 'idle' : 'loading')
+  const state = shallowRef<PaginatorState>(isHydrated.value ? 'idle' : 'loading')
```
patak-dev, line 16: Ah, I forgot about this one. I think we can revert this change, no?

pages/[[server]]/list/[list]/index.vue
```diff
@@ -17,15 +17,15 @@ const tabs = computed<CommonRouteTabOption[]>(() => [
       name: 'list',
       params: { server: server.value, list: list.value },
     },
-    display: t('tab.posts'),
+    display: isHydrated.value ? t('tab.posts') : '',
```
Comment (patak-dev) [inline pages/[[server]]/list/[list]/index.vue:20]: Could this one be removed too? And the one below?

pages/[[server]]/public/index.vue
```diff
@@ -1,6 +1,5 @@
 <script setup lang="ts">
 
-
 const { t } = useI18n()
```
Comment (patak-dev) [inline pages/[[server]]/public/index.vue:3]:
```suggestion

const { t } = useI18n()
```

composables/paginator.ts
userquin: done

pages/[[server]]/list/[list]/index.vue
Comment (userquin) [reply pages/[[server]]/list/[list]/index.vue]: done

pages/notifications.vue
```diff
@@ -50,8 +47,7 @@ const filterIconMap: Record<mastodon.v1.NotificationType, string> = {
   'admin.report': 'i-ri:flag-line',
 }
 
-const filterText = computed(() => (`${t('tab.notifications_more_tooltip')}${filter ? `: ${t(`tab.notifications_${filter}`)}` : ''}`))
-
+const filterText = computed(() => (isHydrated.value ? `${t('tab.notifications_more_tooltip')}${filter.value ? `: ${t(`tab.notifications_${filter.value}`)}` : ''}` : ''))
```
patak-dev, line 50: I think we don't need this `isHydrated` conditional anymore

patak-dev (approved):
Given that the `isHydrated` conditionals changes are now contained, I think we can move forward and merge this PR. If later on we find out a way to avoid these, there isn't the need for a big revert like before.

We discussed with @userquin that it would be ideal if the server will know somehow the language (like we do with the theme color?) so it can directly render the proper strings and avoid the need for the `t` wrappers.

There is also a half fix for https://github.com/elk-zone/elk/issues/2601 included.

PR #2603, opened 2024-02-22
chore: update to Nuxt 3.10.2

userquin:
This PR requires a review and includes:
- update to nuxt 3.10.2
- update to i18n 8.1.1 (some types moved to the module: `LocaleObject` and `Directions` from `@nuxtjs/i18n`): update all required types and moved the configuration to the new i18n options
- update a lot of pages that now requires `isHydrated` to avoid hydration missmatch (also some setting pages updated to remove `isHydrated`): talking with @patak-dev we need to review the `isHydrated`, but I guess we cannot remove it since we've `app.keepAlive`
- pwa not yet tested: updated pwa plugin to latest `0.19.0`, I will do a full review tmr
- ~~module resolution set to `node`: we need a fix for `@vueuse/gesture`, cannot be used with `Bundler`: @danielroe sent a PR to the repo https://github.com/vueuse/gesture/pull/25 (@antfu maybe you can do something on the repo: I have sent also a PR to fix types in `@vueuse/motion`, we don't use the nuxt module but it can be a problem if we enable it here https://github.com/vueuse/motion/pull/175)~~
- vitest version fixed, the same using `@nuxt/test-utils`  (`1.2.2`)
- changed all `process.client`, `process.server` and `process.dev` with `import.meta.client`, `import.meta.server` and `import.meta.dev` respectivelly, we also need to review `process.test` (it is configured in `vite.define`) 
- rn `typescheck` script should be fine, on my local works

The problem with the browser was the `AccountHoverWrapper.vue`, we need to defer the account fetch until visible via interceptor observer when providing only the account handle.

Also fixed the paginator module, changing the state to shallowRef to avoid multiples calls from the watch effect.

We also need to review page lifecycle again, switching between tabs reloading the page when back to elk tab.

supersedes #2543

PR #2610, opened 2024-02-24
chore: update nuxt to 3.10.3

userquin:
I'm going to check what's happening with latest `@nuxt/test-utils`, rn tests working without updating `vitest` and  `@nuxt/test-utils`.

supersedes #2603


patak-dev (approved):
This looks great to me, it is now a lot more clear what is related to the Nuxt update.

For future reference, check out the prev PR from @userquin to update to Vue 3.4 here:
- https://github.com/elk-zone/elk/pull/2607

Let's merge this so we keep updating the rest of the dependencies. Now the history is easy to navigate if @danielroe or others want to check what was involved in updating Nuxt.

PR #2611, opened 2024-02-24
fix: fix `[object Object]` on the mentions tab

shuuji3:
There was one missing unwrapping for the computedRef 🙂 

## Before
Screenshot from 2024-02-24 23-00-19

## After
Screenshot from 2024-02-24 23-00-39


patak-dev:
Good catch!

PR #2612, opened 2024-02-24
feat: . shortcut to show new items

patak-dev:
This should work, but I couldn't test it properly. We have an issue with the subscription to the federated stream it seems, maybe to other streams too.

### DIFF
--- a/components/magickeys/MagickeysKeyboardShortcuts.vue
+++ b/components/magickeys/MagickeysKeyboardShortcuts.vue
@@ -66,6 +66,10 @@ const shortcutItemGroups = computed<ShortcutItemGroup[]>(() => [
         shortcut: { keys: ['c'], isSequence: false },
       },
       {
+        description: t('magic_keys.groups.actions.show_new_items'),
+        shortcut: { keys: ['.'], isSequence: false },
+      },
+      {
         description: t('magic_keys.groups.actions.favourite'),
         shortcut: { keys: ['f'], isSequence: false },
       },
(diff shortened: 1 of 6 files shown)


shuuji3:
Hmm, it looks like something is invalidating the magic keys feature.

When I tried the following, it didn't work: 

1. - Open Elk
2. - Type `?` to see if the magic keys is working -> OK
3. - Click the "Federated" icon on the nav bar
4. - Type `?` to see if the magic keys is working -> stop working
5. - Type `.` -> nothing happens

But the following works as expected:

1. Open Elk
2. Type `?` to see if the magic keys is working -> OK
3. Type `Ctrl+/` to avoid clicking the icon
4. Type `federated` and Enter
5. Type `?` to see if the magic keys is working -> OK
6. Type `.` -> reloaded new posts👌🏻 

Also, directly opening the federated tab (https://deploy-preview-2612--elk-zone.netlify.app/universeodon.com/public) works.

I think we can merge this new shortcut and fix something that breaks the magic keys later separately.

patak-dev:
We saw some issues with MagicKeys with @userquin after updating Vue to 3.4, it could be something in VueUse or on our side. Let's merge this one as it should be unrelated as you explained

meyerweb:
I’ve noticed that this (maybe all magic keys?) seems to fail after I switch to a new tab and then come back to the Elk tab.  I usually switch tabs using keyboard shortcuts (e.g., ⌥⌘→)  so I don’t know if that’s triggering a failure state or not.  If I ever manage to get it fully reproducible, I’ll report back in a new issue.

Issue #1833, opened 2023-02-27
Highlight tags in post where you follow the tag


maxcalvin:
Currently there is nothing really obvious about how a post got into your feed (aside from boosts ofc). This is part two of two about how we might make it a bit clearer.

When a post is from a tag that you follow, the tag(s) could highlight in some obvious way to show that you follow those tags.

Maybe a different colour or a little tick or something.

Shinigami92:
I would like to have this feature as well

PR #2621, opened 2024-02-25
feat: show tag hover card when hovering cursor on hashtag links

### LINKED ISSUE #149: Show details on hover for mentions and tags in statuses
In bird-site there are shown some details for a mention when hovering over it.

Would be cool to have this in Elk, but not only for mention, also for tags. Tags could show some details like history and if you follow it already or not. (example: [GET /api/v1/tags/rust](https://mas.to/api/v1/tags/rust))

Tags are much more important on mastodon as on bird-site as they can be used to follow more content.

https://user-images.githubusercontent.com/7195563/204109692-c34c3ddf-2301-4038-b6b7-eb42d7358dc1.mp4

shuuji3:
This implements the hovercard feature for hashtags.

- resolve #149
- partially help the request for #1833

I think there are several benefits of tag hovercard:
- can follow/unfollow interested tags quickly
- can check whether the tag has been followed in the past or not, without moving to each tag page. This is related to #1833 and useful especially when the post has many hashtags.
- can check how popular each tag is. Currently, clicking a hashtag link navigates users to the tag timeline (example: https://main.elk.zone/mastodon.social/tags/Fediverse) where they cannot see the history graph.

## Screenshots
Screenshot from 2024-02-25 17-35-58

Screenshot from 2024-02-25 17-53-58


patak-dev:
This looks amazing! I agree the UX after this change for tags is greatly improved.

One thing before we merge, could you check the layout of that card?
image

I think we should have more gap between the text and the graph. And maybe push the star to the right of the tag name so the tag looks aligned with the text below.

components/account/TagHoverWrapper.vue
```diff
@@ -0,0 +1,55 @@
...
+  hoverCard,
+  ([{ intersectionRatio }]) => {
+    targetIsVisible.value = intersectionRatio <= 0.75
+  },
+)
+watch(
+  () => [props.tagName, targetIsVisible.value] satisfies WatcherType,
+  async ([newTagName, newVisible]) => {
+    if (newTagName) {
+      tag.value = await fetchTag(newTagName)
```
userquin, line 31: Ensure you dont fetch it if not in screen

userquin: the browser will hang

shuuji3: Good catch! The original implementation was fetching API too aggressively.😅

I moved the `fetchTag()` to `TagCard` component to ensure that it's only called when showing the card.

userquin: I'm going to take a look at this...

shuuji3:
@patak-dev I adjusted the layout a bit. What do you think?

Screenshot from 2024-02-25 20-14-29

Screenshot from 2024-02-25 20-14-34

components/tag/TagCard.vue
```diff
@@ -32,13 +40,13 @@ function go(evt: MouseEvent | KeyboardEvent) {
 
 <template>
   <div
-    block p4 hover:bg-active flex justify-between cursor-pointer
+    block p4 hover:bg-active flex justify-between cursor-pointer flex-gap-2
```
shuuji3, line 43: This gap also improves the layout of the hashtag page in Explore tab for very long hashtag names.

Screenshot from 2024-02-25 20-21-16

userquin: I think for long tags the start should be top aligned

shuuji3:
Hmm, I'm not sure why snapshot tests start revealing `<VMenu>` tag here: https://github.com/elk-zone/elk/actions/runs/8037557151/job/21952513803#step:7:88

I wonder if it's safe to update the snapshot test. 🤔

userquin:
> Hmm, I'm not sure why snapshot tests start revealing `<VMenu>` tag here: https://github.com/elk-zone/elk/actions/runs/8037557151/job/21952513803#step:7:88
> 
> I wonder if it's safe to update the snapshot test. 🤔

We'll need to mock it, I'm ging to fix it

userquin:
@shuuji3 your local deps are stale, run `nr test:ci` in your local

components/tag/TagCard.vue
patak-dev: I'd like the tag to always be aligned with the text below. I proposed having the star to the right, but I see now that it isn't a good idea. Maybe we should have the star always at the top left, and then the tag and the text below to the right of it. Or the star could be at the right of the graph, kind of like the Follow button in the notification cards for people following you

userquin:
@shuuji3 the logic seems to be wrong, we should avoid fetching the tag until visible in the viewport, the browser will hang if lot of tags, the result is not deterministic since it depends on the current timeline content

shuuji3:
@userquin Did your local branch have this commit (https://github.com/elk-zone/elk/pull/2621/commits/381d166c871495944c2f7a2bd7168d4bae99ceb2)? I observed that issue in the network tab on devtool, but after this commit. that issue should be fixed.

userquin:
yes, but this is wrong, the tagName will not prevent to fetch the tag once set, you have the fetch with toplevel await in the TagCard script setup

components/tag/TagCard.vue
shuuji3: OK, I tried another layout by making the follow button an independent block and placing it on the left. This ensures both texts are aligned and more robust with a long hashtag.

For Hovercard, `items-center` looks lined up better than the top. For Hashtag, both look fine. 

## Hovercard
Screenshot from 2024-02-25 21-19-41

### align center (`items-center`)
Screenshot from 2024-02-25 21-19-06
Screenshot from 2024-02-25 21-10-25

### align top (`items-start`)
Screenshot from 2024-02-25 21-18-56
Screenshot from 2024-02-25 21-11-57

## Hashtag page

Screenshot from 2024-02-25 21-44-05
Screenshot from 2024-02-25 21-43-37

patak-dev:
Thanks for playing with the design of the card, I think it looks a lot better now! 🧡

components/account/TagHoverWrapper.vue
```diff
@@ -1,39 +1,55 @@
...
   },
 )
 
 watch(
-  () => [props.tagName, targetIsVisible.value] satisfies WatcherType,
-  async ([newTagName, newVisible]) => {
-    if (newTagName) {
-      tagName.value = newTagName
+  () => [props.tag, props.tagName, targetIsVisible.value] satisfies WatcherType,
+  async ([newTag, newTagName, newVisible], oldProps) => {
```
userquin, line 32: remove `async`

shuuji3: done 👍🏻

shuuji3:
Updated to the new layout and fixed the remaining test error now.

shuuji3:
CI is still stacking while my local test passed within 3 seconds 😅 

```
...
 Test Files  6 passed (6)
      Tests  43 passed (43)
   Start at  23:12:43
   Duration  2.85s (transform 2.42s, setup 3.76s, collect 1.90s, tests 129ms, environment 368ms, prepare 1.28s)


Process finished with exit code 0
```

userquin:
I'm going to extract fetch promises to another PR, and remove the logic here for account logic

userquin:
@shuuji3 you should update this PR when my last 2 PR merged: change logic in fetchTag and should be ready:
- https://github.com/elk-zone/elk/pull/2623
- https://github.com/elk-zone/elk/pull/2624

userquin (approved):

composables/settings/definition.ts
```diff
@@ -15,7 +15,7 @@ export interface PreferencesSettings {
   hideFollowerCount: boolean
   hideTranslation: boolean
   hideUsernameEmojis: boolean
-  hideAccountHoverCard: boolean
+  hideAccountAndTagHoverCard: boolean
```
patak-dev, line 18: Let's split this in two, we keep `hideAccountHoverCard` and we add a new `hideTagHoverCard`. I think it is better so we respect current preferences and also I imagine there are folks that will care about the accounts but would prefer to hide the tags.

patak-dev (changes requested):

composables/settings/definition.ts
shuuji3: Yes, that makes sense as they had a bit different roles. I added `hideTagHoverCard` as a separate one.

patak-dev (approved):
Awesome!

I see it is still a draft, mark it as ready when you think we should merge it

shuuji3:
Now, `fetchTag()` should be called only once when hovering it. Also, the component code can be simplified. I hope this works fine.

https://github.com/elk-zone/elk/assets/1425259/78bc97df-2bd7-42cc-abae-bc890815fbf3

components/account/TagHoverWrapper.vue
```diff
@@ -1,63 +1,38 @@
...
+  }
+})
 
 const userSettings = useUserSettings()
 </script>
 
 <template>
-  <span ref="hoverCard">
-    <VMenu v-if="!disabled && tag && !getPreferences(userSettings, 'hideTagHoverCard')" placement="bottom-start" :delay="{ show: 500, hide: 100 }" v-bind="$attrs" :close-on-content-click="false">
+  <span @mouseenter="hovered = true">
```
userquin, line 28: add @mouseleave and check if hovered is true in the watcher

shuuji3:
I had to update the same test snapshot again for strange `<VMenu>`. I guess this original comment is correct in a way:
https://github.com/shuuji3/elk/blob/db070eed8111bfc71b0686f9003f8cd736aff55a/tests/nuxt/content-rich.test.ts#L163-L163

userquin:
> I had to update the same test snapshot again for strange `<VMenu>`. I guess this original comment is correct in a way

The bdi content should be there I'm going to check what's wrong in the content-render module

components/account/TagHoverWrapper.vue
shuuji3: Updated here to check mouseleave too so that each hovering will fetch tag cache.

shuuji3:
TODO: show spinner before receiving the actual tag data.

userquin:
Check https://github.com/elk-zone/elk/pull/2632

shuuji3:
There was a good skeleton component for the tag: https://github.com/elk-zone/elk/blob/main/components/tag/TagCardSkeleton.vue

https://github.com/elk-zone/elk/assets/1425259/f272a388-3ac3-4471-b806-f92e20e190bb

shuuji3:
[`f67be19` (#2621)](https://github.com/elk-zone/elk/pull/2621/commits/f67be19ea9cb86c6d6688a1cd4ce99614cec6978)

Also, there was a useful `useElementHover` utility instead of the manual mouse event handling. This can be used in #2632 as well.

patak-dev (approved):
Amazing work! 👏🏼

PR #2622, opened 2024-02-25
chore(test): add `hanging-process` reporter on CI

userquin:


patak-dev (approved):

patak-dev:
The new info doesn't seems very useful, no? https://github.com/elk-zone/elk/actions/runs/8038042822/job/21953547942

userquin:
> The new info doesn't seems very useful, no? https://github.com/elk-zone/elk/actions/runs/8038042822/job/21953547942

yes and no, it seems we have something trying to resolve some async call: I'll add `!import.meta.test` in the account wrapper. The link in the CI is about using fetch in node: https://vitest.dev/guide/common-errors.html#failed-to-terminate-worker

PR #2623, opened 2024-02-25
fix(cache): return cached account as promise

### LINKED ISSUE #2601: User profile panel is not shown when hovering

userquin:


composables/cache.ts
```diff
@@ -26,23 +25,12 @@ export function fetchStatus(id: string, force = false): Promise<mastodon.v1.Stat
...
-
-  const promise = new Promise<any>((resolve) => {
-    fetchPromise!.then(resolve)
-  })
-  if (!cache.has(key) || force)
-    cache.set(key, promise)
-
+  const promise = useMastoClient().v1.statuses.$select(id).fetch()
+    .then((status) => {
+      cacheStatus(status)
```
patak-dev, line 30: This is a bit confusing. Is it really needed for the `key` to be recalculated? Couldn't we directly call `cache.set(key, status)` here?

userquin: We register 2 things there... the code is the original,  we can chabge it but the code will be duplicated

composables/cache.ts
```diff
@@ -57,27 +45,16 @@ export function fetchAccountById(id?: string | null): Promise<mastodon.v1.Accoun
...
-  if (!cache.has(key))
-    cache.set(key, promise)
-
+  const domain = getInstanceDomainFromServer(server)
+  const promise = useMastoClient().v1.accounts.$select(id).fetch()
+    .then((r) => {
+      if (r.acct && !r.acct.includes('@') && domain)
+        r.acct = `${r.acct}@${domain}`
+
+      cacheAccount(r, server, true)
```
patak-dev, line 54: Same here, but this one seems to be doing more. It may be good to call `cache.set(key, r)` here and then do whatever else `cacheAccount` does in a separate function. We could do that in another PR though.

composables/cache.ts
```diff
@@ -106,23 +83,13 @@ export async function fetchAccountByHandle(acct: string): Promise<mastodon.v1.Ac
...
-
-  const promise = new Promise<any>((resolve) => {
-    fetchPromise!.then(resolve)
-  })
-  if (!cache.has(key))
-    cache.set(key, promise)
 
+  const promise = lookupAccount()
+    .then((r) => {
+      cacheAccount(r, server, true)
```
patak-dev, line 89: Same here. I think it is important that we make sure the `key` is the same, so we don't leave any promises in the cache.

composables/cache.ts
userquin: This pr is about to fix the return on these 3 methods,  should return a promise to avoid errors like `.then not found`

userquin: I can send a pr tmr to includeyour suggestions

patak-dev: Ah, it seems I was looking at a previous commit

patak-dev (approved):

PR #2631, opened 2024-02-26
fix: allow to edit alt description of attached image again

### LINKED ISSUE #2629: Stopped being able to add alt text
I would suggest this is a severe bug - alt text it is crucial to making Elk and Mastodon accessible

on main.elk.zone

I can no longer add alt text to images. When I do, then post, the image is posted with no alt text

On chrome android and safari macos

shuuji3:
fix #2629

I wrote the cause of the issue in #2629.


patak-dev (approved):

patak-dev:
Good catch!

PR #2632, opened 2024-02-27
feat(ui): fetch account data on demand

userquin:
This PR includes:
- fetch account data when hovering on the account link (on demand)
- ~~add status card skeleton when resolving the account data~~

This will also prevent fetching account data of all accounts in the viewport while scrolling.

We can change the skeleton or switch to use an spinner loader.

Check this video https://streamable.com/4kk87e : added few seconds timeout to see the skeleton on first fetch (afterwards it is immediate)

/cc @patak-dev @shuuji3

### DIFF
--- a/components/account/AccountHoverCard.vue
+++ b/components/account/AccountHoverCard.vue
@@ -10,8 +10,8 @@ const relationship = computed(() => account ? useRelationship(account).value : u
 
 <template>
   <div>
-    <StatusCardSkeleton v-if="!account" />
-    <div v-else v-show="relationship" flex="~ col gap2" rounded min-w-90 max-w-120 z-100 overflow-hidden p-4>
+    <StatusCardSkeleton v-if="!relationship" />
+    <div v-else-if="account" flex="~ col gap2" rounded min-w-90 max-w-120 z-100 overflow-hidden p-4>
       <div flex="~ gap2" items-center>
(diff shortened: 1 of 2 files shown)


patak-dev:
Is the skeleton still working for you? I can't see it when I test the preview: https://deploy-preview-2632--elk-zone.netlify.app/

userquin:
skeleton logic removed, only fetch on hover

components/account/AccountHoverCard.vue
```diff
@@ -5,7 +5,7 @@ const { account } = defineProps<{
   account: mastodon.v1.Account
 }>()
 
-const relationship = useRelationship(account)
+const relationship = computed(() => account ? useRelationship(account).value : undefined)
```
patak-dev, line 8: I don't think we should call `useRelationship` inside a `computed`. I think you should extract `useRelationship` to an outside `ref`. If you don't want that ref (that changes once `fetchRelationships()` finished), then you could abstract a helper around `requestedRelationships`

userquin: done

userquin: too many `useRelationship` usages to refactor it here: we should debonce the call

patak-dev (approved):

PR #2635, opened 2024-02-28
feat: allow to scroll-to-top by clicking header title in account, list, and tag pages

shuuji3:
In Elk, we can click the title to scroll to the top pages. This is useful after reading many posts. But some pages don't have this feature and the title text style was inconsistent.

This RP ensures that three pages (user account, list, tag) have this feature. All of them usually have many posts so the scroll-to-top feature should be useful too.

### Before
Screenshot from 2024-02-28 23-13-52

### After
Screenshot from 2024-02-28 23-13-28

### Before
Screenshot from 2024-02-28 23-14-58
### After
Screenshot from 2024-02-28 23-14-46

### Before
Screenshot from 2024-02-28 23-14-19
### After

Screenshot from 2024-02-28 23-14-30

### DIFF
--- a/pages/[[server]]/@[account]/index.vue
+++ b/pages/[[server]]/@[account]/index.vue
@@ -24,10 +24,11 @@ onReactivated(() => {
   <MainContent back>
     <template #title>
       <ContentRich
-        timeline-title-style
+        timeline-title-style cursor-pointer
         :content="account ? getDisplayName(account) : t('nav.profile')"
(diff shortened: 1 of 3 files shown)


patak-dev:
I think this is a good change, but we should first fix an issue with the back arrow as it will happen after this PR in more cases. The back arrow should only be there if there is history inside the page to navigate, if not we shouldn't show it. For example, I think the arrow shouldn't appear if I visit this link: https://deploy-preview-2635--elk-zone.netlify.app/universeodon.com/@elkdev

PR #2640, opened 2024-02-29
fix: show correct reply target user account in reply post header

shuuji3:
The reply status account is expected to be the target account but actually, it showed the reply author.

ref. https://discord.com/channels/1044887051155292200/1046769315405365288/1212694499160096858


userquin:
This is also fixed in #2638, can be merged to fix th eproblem in main.elk.zone

userquin (approved):

PR #2642, opened 2024-02-29
feat: introduce new "Followed tags" page

### LINKED ISSUE #1194: provide a list of followed hashtags

### LINKED ISSUE #2081: Add a tab to Favorites to manage favorited hashtags
I made the mistake of favoriting #art and its now flooding my feed and pushing out everything else. I went to my favorites menu and was surprised that this option wasn't already there (it was the instinctual choice).

Having a tab in favorites to look at all your favorited hashtags, with the option to unfollow them would be a big improvement. Thanks!

### LINKED ISSUE #2604: Followed #tags block
Hey Elk,

Just came up my favourite feature of FediLab: followed tags menu item. 

I like it very much! Pretty usable, handy and probably not too much time to implement. Just an option to add/remove tags, and a tag list page. Clever feature? 

Screenshot_2024-02-22-18-29-05-383_app fedilab android

What do you think? Thank you for your consideration.

shuuji3:
resolve #2604

This probably also
resolve #2081 and
resolve #1194

I also wrote about a use-case of this feature in #2604.

From the "Followed tags" page, users can check or unfollow the current following tags, or navigate to the specific hashtag timeline to read posts at any time they like. They can use this timeline similar to the list timeline but for specific interests.

Also, the hashtag icon is conflicted with the "Explore" icon but Mastodon updated the icon to the compass icon just last week. So I adopted that icon at the same time.

ref. Change explore icon from hashtag to compass in web UI by Gargron · Pull Request #29294 · mastodon/mastodon - https://github.com/mastodon/mastodon/pull/29294

https://mastodon.social

(#1251 was another attempt but I prefer the new icon because of the reason above)

## Desktop
Screenshot from 2024-02-29 22-22-09
Screenshot from 2024-03-01 12-56-49
Screenshot from 2024-02-29 22-18-56


shuuji3:
ci is failed due to the stuck of `pnpm test:ci` process exit waiting for 6 hours.

patak-dev:
This looks awesome @shuuji3!

I merged main that has some fixes from userquin to the CI tests, let's check if that helps.

I'm thinking we could use "Hashtags" instead of "Followed Tags". I think users would understand that these are their tags. And may we could add at the end of the list later on something like "Explore hashtags gaining traction" so it makes even more sense. Maybe it could be there when there aren't any followed tags so it doesn't bother later or we could make it dismissible.

About "Hastags" vs "Tags", it seems this is the word we are using currently in other parts of the app. We could review switching to "Tags" everywhere later on.

patak-dev (approved):
lovely ❤️

shuuji3:
@patak-dev Great, CI was fixed and the execution time was reduced drastically from 6 hours to 8 seconds 😄 

> I'm thinking we could use "Hashtags" instead of "Followed Tags".

Right, I just reused the existing wording but that would be more clear.

I haven't implemented the trending tags when the list is empty yet, but I'll try adjusting in a separate PR.

PR #2645, opened 2024-03-02
fix: fix emoji placement (#2626)

### LINKED ISSUE #2626: Emoji gets inserted one character earlier than it should (?)
It seems that the emoji gets inserted one character earlier than it should.

**Tested on**: iPhone 11 with iOS 17.3.1, Elk v0.11.0 installed as a PWA / bookmark.
**Steps to reproduce**:
1. Compose a new post, write for example "12345".
2. Move cursor between 3 and 4 and insert an emoji from the emoji keyboard.
3. The emoji appears between 2 and 3.

Here's a screen recording:
https://github.com/elk-zone/elk/assets/20538005/1258c4c4-3aee-4f3f-90b8-c25c17a24acf

nonnullish:
Fixes #2626.

The issue was in tiptap's [`nodeInputRule` ](https://github.com/ueberdosis/tiptap/blob/e00700110af771c6f7f826a5d54802ebfed6aeb4/packages/core/src/inputRules/nodeInputRule.ts#L60), so I rewrote that function.

This is my first PR in this repo so please let me know if anything's not right.

### DIFF
--- a/composables/tiptap/emoji.ts
+++ b/composables/tiptap/emoji.ts
@@ -1,10 +1,12 @@
-import type { ExtendedRegExpMatchArray } from '@tiptap/core'
+import type { ExtendedRegExpMatchArray, InputRuleFinder, nodeInputRule } from '@tiptap/core'
 import {
+  InputRule,
   Node,
+  callOrReturn,
   mergeAttributes,
-  nodeInputRule,
   nodePasteRule,
 } from '@tiptap/core'
+import type { NodeType } from '@tiptap/pm/model'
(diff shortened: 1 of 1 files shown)


userquin:
Add `InputRule` to `mocks/tiptap.ts` module

shuuji3:
Thank you, I confirmed the fix now 👍🏻

https://github.com/elk-zone/elk/assets/1425259/d914ff2e-8eda-4918-978a-4334bc5f87f7

patak-dev (approved):
This is such a big quality of life improvement! Thanks a lot @nonnullish 🧡

PR #2659, opened 2024-03-07
fix(pwa): wrong web manifest colors

userquin:
The colors should be in sync with `color-mode` plugin


userquin:
maybe we should extract those colors to constants module

patak-dev (approved):

PR #2661, opened 2024-03-07
fix: consistent hover highlight styling in mobile bottom navigation menus

### LINKED ISSUE #2656: Bottom navigation items have narrow background highlight in mobile view
## Isssue
This is a small styling issue. In the mobile view, the bottom two menus have wide background highlights but the others have only narrow highlights.

 

Also, it seems that both have slightly different colors:
- Top items: `var(--c-bg-active)` (`#f2f2f2`)
- Bottom items: `rgb(243 244 246 / var(--un-bg-opacity))`

## Solution

All items have a wide click area so it is more natural to have wider highlighted areas like the bottom ones.

shuuji3:
fix #2656 

Adopted the same hover highlight style as the bottom two items like the Zen mode menu.

### Before

In dark mode, the highlight was really hard to see:

Screenshot from 2024-03-08 00-25-37

### DIFF
--- a/components/nav/NavSideItem.vue
+++ b/components/nav/NavSideItem.vue
@@ -57,11 +57,21 @@ const noUserVisual = computed(() => isHydrated.value && props.userOnly && !curre
       <div
         class="item"
         flex items-center gap4
-        w-fit rounded-3
-        px2 mx3 sm:mxa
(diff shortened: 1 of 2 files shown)


patak-dev (approved):
Awesome!

PR #2665, opened 2024-03-09
feat: allow to set mute duration and notifications mute option

### LINKED ISSUE #2301: Add mute duration
This is kind of a companion proposal to https://github.com/elk-zone/elk/issues/1078.

I often find myself wanting to mute an account for a day or two when they are "heavy-posting" about a topic.

The official Mastodon client lets you specify a duration with a drop down menu when muting an account (see screenshot in the issue above).

Could this capability be added to elk?

shuuji3:
resolve #2301

This PR adds the two mute options in the confirm dialog for muting.

I set the default duration setting as 1 hour (this can be changed if the other value is better), and the mute notifications enabled, which is the current default.

### Default

Screenshot from 2024-03-09 15-20-15

### Setting example

Screenshot from 2024-03-09 15-03-25

### Invalid duration

Screenshot from 2024-03-09 16-39-41

disables the "Mute" button

## Implementation notes

To implement this feature, I adjusted the existing `ConfirmDialog` component to be able to show mute-related options and return selected information. (4276b50d49caf59252c5db0fd9c8bb0e6c49028e cca244ad12b56ac58173113cec38d458a5109a95)

Also, I implemented the `DurationPicker` component (ee0742e7701cf0d703b8e50aa4297fa8fd241581), where we can select minutes with 5 minutes step. I chose a 5-minute step for the minutes input to make it easier to select, but users can input any minute value type by directly typing.

Another adjustment is `CommonCheckbox` (bbb15dc6992dd4b8ed96e88641d8f854e4819186). Previously, it could only show its checkbox at the end of the right side like on the settings page. This change allows us to put the checkbox before the label and accept the checkbox color change too.

### DIFF
--- a/components/account/AccountMoreButton.vue
+++ b/components/account/AccountMoreButton.vue
@@ -25,13 +25,16 @@ function shareAccount() {
 }
 
 async function toggleReblogs() {
-  if (!relationship.value!.showingReblogs && await openConfirmDialog({
-    title: t('confirm.show_reblogs.title'),
-    description: t('confirm.show_reblogs.description', [account.acct]),
-    confirm: t('confirm.show_reblogs.confirm'),
-    cancel: t('confirm.show_reblogs.cancel'),
-  }) !== 'confirm')
-    return
+  if (!relationship.value!.showingReblogs) {
+    const dialogChoice = await openConfirmDialog({
+      title: t('confirm.show_reblogs.title'),
+      description: t('confirm.show_reblogs.description', [account.acct]),
+      confirm: t('confirm.show_reblogs.confirm'),
+      cancel: t('confirm.show_reblogs.cancel'),
+    })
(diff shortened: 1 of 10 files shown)


patak-dev (approved):
This is great! A step closer to feature parity 🙌🏼

I think we should merge this one now, and later on we could iterate and refactor it if we end up with more extra options. It is probably better that the confirm dialog will have a slot where you could inject extra options, and have each of these dialogs as a separate component.

shuuji3:
@patak-dev Thanks for a good suggestion! Right, it was a better choice to use slot inject extra options, instead of adding up pieces only needed for the mute functionality in this case.

I think I can refactor it later. Since I actually haven't had so much experience composing a component using the slot, that'll be a good exercise for me too 🙂

PR #2667, opened 2024-03-09
fix: show loading spinner while fetching account relationship

### LINKED ISSUE #2647: Avoid misleading default "Follow" button label
## Issue 

Currently, the follow button in the profile shows the default label "Follow" for an initial subsecond. This causes a temporary misunderstanding to users and can cause them to think that, for example, they might unfollowed the profile user by mistake.

https://github.com/elk-zone/elk/assets/1425259/fd389576-1077-4585-9ccb-280d40989774

(Reloading the profile page showing the "Follow" button.)

## Solution

To prevent showing a button label different from the actual following state by default, it would be nice to have some kind of loading effect or a different label indicating the following state is loading at the moment.

shuuji3:
fix #2647

This PR makes the loading spinner shown while loading a follow relationship with an account so that we can avoid showing a different button label from the actual relationship. The difference is noticeable, especially when the loading time is long.

### Before
https://github.com/elk-zone/elk/assets/1425259/3c8a90ec-eb23-403f-9612-a4dd55e93713
### After
https://github.com/elk-zone/elk/assets/1425259/a0f2f38c-8f6a-4fe1-9e5f-a64129a73f56

### DIFF
--- a/components/account/AccountFollowButton.vue
+++ b/components/account/AccountFollowButton.vue
@@ -13,6 +13,11 @@ const { t } = useI18n()
 const isSelf = useSelfAccount(() => account)
 const enable = computed(() => !isSelf.value && currentUser.value)
 const relationship = computed(() => props.relationship || useRelationship(account).value)
+const isLoading = ref(false)
+
+watchEffect(() => {
+  isLoading.value = relationship.value === undefined
+})
 
 const { client } = useMasto()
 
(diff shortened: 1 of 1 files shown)


components/account/AccountFollowButton.vue
```diff
@@ -13,6 +13,11 @@ const { t } = useI18n()
 const isSelf = useSelfAccount(() => account)
 const enable = computed(() => !isSelf.value && currentUser.value)
 const relationship = computed(() => props.relationship || useRelationship(account).value)
+const isLoading = ref(false)
+
+watchEffect(() => {
+  isLoading.value = relationship.value === undefined
+})
```
patak-dev, line 20: Could this be rewritten as:
```suggestion
const isLoading = computed(() => relationship.value === undefined)
```
?

patak-dev:
Looks awesome 💯
This is a big improvement

components/account/AccountFollowButton.vue
shuuji3: That's correct 👍🏻

patak-dev (approved):

PR #2669, opened 2024-03-09
fix: fix incorrect follow status on followers and following pages

### LINKED ISSUE #2668: Follow status on follower/following page is incorrect
The follow status seems to be shuffled among accounts on follower/following page.

I found a potential solution. I'll send a PR to fix this.

shuuji3:
fix #2668

There were two cases missed by the current implementation:
1. Mastodon relationships API accepts batch requests for multiple IDs (ref. https://docs.joinmastodon.org/methods/accounts/#relationships) but it does not always ensure the order of returned relationships. For example, when we requests for `[id1, id2]`, the response can be `[relationship2, relationship1]`. 
1. when there's an issue with fetching the relationship for some accounts, the response array omits those accounts in response. For example, the request `[id1, id2, id3]` can result in the response `[relationship1, relationship3]`.

This fix ensures using only returned relationships and updating the correct corresponding relationship data with the same ID.
 
## Debugging notes

I inserted the following logs here to see how the relationship value is assigned:

https://github.com/elk-zone/elk/blob/344833535641f6ce394d4cbf7cb1f423a2ea33dd/composables/masto/relationship.ts#L32

```shell
    console.log('assigning:', requested[i][0], '<-', relationships[i].id)
```

I got the results like the following:

```
...
relationship.ts:36 assigning: 1147 <- 109491336352118388
relationship.ts:36 assigning: 111186761817498794 <- 109515037453915631
relationship.ts:36 assigning: 13179 <- 110645833556279910
relationship.ts:36 assigning: 1 <- 111186761817498794
...
```

which means that for example, the fetched relationship for the user ID 111186761817498794 is stored as the relationship for the user ID 1. This is clearly wrong.

This is the example logs after the fix:
```shell
...
relationship.ts:41 assigning: 108390124684993425 <- 108390124684993425
relationship.ts:41 assigning: 111086202750549423 <- 111086202750549423
relationship.ts:41 assigning: 111694022200023471 <- 111694022200023471
...
```

### DIFF
--- a/composables/masto/relationship.ts
+++ b/composables/masto/relationship.ts
@@ -11,9 +11,12 @@ let timeoutHandle: NodeJS.Timeout | undefined
 export function useRelationship(account: mastodon.v1.Account): Ref<mastodon.v1.Relationship | undefined> {
   if (!currentUser.value)
     return ref()
+
   let relationship = requestedRelationships.get(account.id)
   if (relationship)
     return relationship
+
+  // allow batch relationship requests
   relationship = ref<mastodon.v1.Relationship | undefined>()
   requestedRelationships.set(account.id, relationship)
   if (timeoutHandle)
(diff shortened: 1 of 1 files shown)


patak-dev (approved):
Good catch!

PR #2674, opened 2024-03-10
chore: upgrade `@vueuse/core` from `10.8.0` to `10.9.0`

### LINKED ISSUE #2628: Keyboard shortcuts stop working after the first one
on main.elk.zone EDIT - and elk.zone

press c for Compose, the compose screen appears
press esc to dismiss
press c again, or any other keyboard command, nothing happens

safari on macos

EDIT

In chrome on macos, I can press c and esc and continue to press c and it works. However if I press ? then esc to dismiss, again c and esc stop working

shuuji3:
probably fix #2628

There was an GitHub Issue for `useActiveElement` in the `@vueuse/core` repository, which was fixed by the latest version `10.9.0`: https://github.com/vueuse/vueuse/issues/3799.

We are using almost the same logic for our keyboard shortcuts here: https://github.com/elk-zone/elk/blob/edcc8741bfe63302ae1ba58886ccd7279115e345/plugins/magic-keys.client.ts#L10-L18

Also, confirmed this issue was fixed: https://github.com/elk-zone/elk/pull/2612#issuecomment-1962390478

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,7 +60,7 @@
     "@upstash/redis": "^1.27.1",
     "@vercel/kv": "^1.0.1",
     "@vue-macros/nuxt": "^1.6.0",
-    "@vueuse/core": "^10.8.0",
+    "@vueuse/core": "^10.9.0",
     "@vueuse/gesture": "^2.0.0",
     "@vueuse/integrations": "^10.8.0",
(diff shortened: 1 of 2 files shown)


patak-dev (approved):
🤞

PR #2676, opened 2024-03-10
fix: prevent empty search keyword to send invalid request

### LINKED ISSUE #2675: Empty search keyword cause console error log: `param is missing or the value is empty: q`
This issue doesn't affect users, but Elk is sending an invalid request to `https://mastodon.social/api/v2/search?q=&resolve=true` causing a 400 Bad Request error.

## How to reproduce
1. Open https://elk.zone/search
2. Type one character in search input
3. Delete the character
4. The following error is shown in the console

```
GET https://mastodon.social/api/v2/search?q=&resolve=true 400 (Bad Request)
param is missing or the value is empty: q
```

shuuji3:
fix #2675

This was caused by another missing `.value` 😅

### DIFF
--- a/composables/masto/search.ts
+++ b/composables/masto/search.ts
@@ -63,7 +63,7 @@ export function useSearch(query: MaybeRefOrGetter<string>, options: UseSearchOpt
   })
 
   debouncedWatch(() => resolveUnref(query), async () => {
-    if (!q || !isHydrated.value)
+    if (!q.value || !isHydrated.value)
       return
 
     loading.value = true


patak-dev (approved):
These are really hard to spot. Maybe I should have used an automated tool 🫠

Thanks for fixing it!

userquin:
@shuuji3 L62 needs also .value

shuuji3:
@userquin @patak-dev Yes, that's another one👍🏻  I also found a way to enable this rule and found other ones too. I will send a PR to fix them together today.

shuuji3:
I wrote the workaround here: #2678

PR #2686, opened 2024-03-12
fix(ui): hashtags not working when composing

userquin:
Updating tiptap seems to break the hashtag "plugin" (added data-type and data-id to the span), similar problem wth mentions plugin in #2655


patak-dev (approved):
Good catch!

PR #2699, opened 2024-03-18
fix: pre tag overwritten by default style

### LINKED ISSUE #2590: Code wrapping mismatch between edit and view
I created a post which I used the triple back tick notation for code/preformatted.

In the editor this was word wrapped however when viewing the post it isn't.

Sma11X:
fix #2590 
Ensure the `pre` tag style is not overwritten by the user agent stylesheet. The default value was set to `pre`.

### DIFF
--- a/styles/global.css
+++ b/styles/global.css
@@ -85,6 +85,9 @@ em-emoji-picker {
   p:last-child {
     --at-apply: mb-1;
   }
+  pre {
+    --at-apply: whitespace-pre-wrap;
+  }
   code {
     --at-apply: bg-code text-code px1 py0.5 rounded text-0.875em leading-0.8em;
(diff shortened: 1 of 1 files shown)


shuuji3 (approved):
Thanks for the fix! I can confirm that the `pre-wrap` is now correctly applied in the same way as the editor dialog on both the timeline and the status pages.

## Before
Screenshot from 2024-03-18 18-04-58

## After
Screenshot from 2024-03-18 18-05-43

shuuji3:
(By the way, I noticed that the syntax color is not applied on the initial load. But this is a separate issue so I will file it as another Issue.)

patak-dev (approved):

PR #2707, opened 2024-03-21
fix: correct local timeline stream

Sma11X:
Now the `load new items` button is currently not visible due to being associated with the wrong stream.


patak-dev (approved):
Thanks for the fix! 🧡

PR #2709, opened 2024-03-21
feat(ui): add max. file size check before upload attachment

### LINKED ISSUE #1674: Video upload error
I think #595 came back?

image

One difference is that there is a warning on `MastoClient` this time I think
```#create is deprecated. Use MastoClient.v2.media.create instead undefined```

image

### PR DESCRIPTION
**Pull request by userquin**
This PR prevents the file upload using the `mediaAttachments` instance configuration:
- `videoSizeLimit` for videos
- `imageSizeLimit` for images

There are 2 more entries `imageMatrixLimit` and `videoMatrixLimit`, no idea if we should use them.

There is a problem with audio files, for example, webtoo.ls allowing audio files but there is no `audioSizeLimit` in `mediaAttachments` options: should we use video or image? Right now using video size limit (we also need to include the audio message, this PR using the video message).

This PR also includes 2 new entries in `state`, using Intl.NumberFormat inside `useUploadMediaAttachment` composable (composables/masto/publish.ts module):
- `attachments_limit_image_error`
- `attachments_limit_video_error`
- `attachments_limit_audio_error`
- `attachments_limit_unknown_error`

This PR includes translations for English (en.json), Spanish (es.json and es-419.json): also fixes an entry in es.json (video => vídeo). Messages are not reactive: if you go to the settings and change the locale the messages in the compose page will not be updated.

```ts
const number = 123456.789;

console.log(
  new Intl.NumberFormat('ar-EG', { style: 'unit', unit: 'megabyte', unitDisplay: "narrow", maximumFractionDigits: 0 }).format(
    number/ (1024*1204),
  ),
);
console.log(
  new Intl.NumberFormat('en-US', { style: 'unit', unit: 'megabyte', unitDisplay: "narrow", maximumFractionDigits: 0 }).format(
    number/ (1024*1204),
  ),
);
```

/cc @mrcego 

closes #1674

en.json

imagen

es.json

imagen

es-419.json

imagen


userquin:
It seems `webtoo.ls` getting wrong instance values, previous configuration entries allows uploading videos up to 40MB but that's not true, check linked issue (16MB).

shuuji3:
I found this report on Mastodon repository: https://github.com/mastodon/mastodon/issues/22757

I suspect 413 responses are raised due to the HTTP server application setting, instead of Mastodon setting. It looks like if Mastodon checks the uploaded video, it will return 422 when the size is exceeded.

If this assumption is correct, we will be able to upload 16MB video by adjusting the server configuration. 

Either way, this new warning is a good addition.

userquin:
@shuuji3 the problem is that masto api not throwing proper error and so we cannot access to the status code, 413 should be used since it is the proper http header for it: entity too large iirc.

This PR should be ready, just added audio and unknown logic and messages.

userquin:
should we use `v2` api? this PR still using v1

userquin:
Intl.NumberFormat working also in FireFox, we need to test it in Safari:

imagen

shuuji3:
checked quickly on Safari console (macOS)
Screenshot 2024-03-22 at 12 24 27

This method is well supported now so it should be ok: https://caniuse.com/?search=Intl.NumberFormat

> should we use `v2` api? this PR still using v1 in the create call

Does the masto.js interface has changed a lot? If we need to use the API differently, maybe it would be better to a separate PR to isolate the change.

userquin:
The media v2 no, only the create method (now sent with a multipart form), maybe also the implementation the api is the same.

userquin:
We alos need to handle kb,  I'm going to move the logic to i18n composable.

shuuji3:
> There are 2 more entries `imageMatrixLimit` and `videoMatrixLimit`, no idea if we should use them.

The `imageMatrixLimit` is the number of pixels in an image and m.webtoo.ls has the limit of `16777216`. I thought the image with more than this pixel could not be loaded, but the larger image was resized on the server side. So I think we don't have to use this number to check the image resolution on the client side.

I tried uploading 10,000x10,000px png image (exceeding `16777216`) but it was successfully uploaded and resized to 1,440x1,440px by Mastodon server.
Screenshot from 2024-03-25 13-09-01

shuuji3 (approved):
I tested large audio files (90MB and 100MB) on mastodon.social, which configured 100MB as the video size limitation. I couldn't find the audio specific size limitation but the limitation is virtually the same as the video size limitation (<100MB). So it's reasonable to assume the limit of the audio size is the same as video.

composables/i18n.ts
```diff
@@ -77,3 +77,30 @@ export function useTimeAgoOptions(short = false): UseTimeAgoOptions<false> {
...
+      style: 'unit',
+      unit: 'kilobyte',
+      unitDisplay: 'narrow',
+      maximumFractionDigits: 0,
+    }),
+  ]))
+
+  function formatFileSize(size: number) {
+    return size >= 1048576
+      ? formatters.value[0].format(size / 1048576)
```
shuuji3, line 101: nit: maybe this is a bit more readable?

```suggestion
      ? formatters.value[0].format(size / (1024 * 1024))
```

composables/masto/publish.ts
```diff
@@ -224,8 +225,32 @@ export function useUploadMediaAttachment(draft: Ref<Draft>) {
...
+              ...failedAttachments.value,
+              [file.name, t(
+                file.type.startsWith('audio/')
+                  ? 'state.attachments_limit_audio_error'
+                  : file.type.startsWith('video/')
+                    ? 'state.attachments_limit_video_error'
+                    : 'state.attachments_limit_unknown_error',
+                [formatFileSize(maxVideoSize)],
+              )],
+            ]
```
shuuji3, line 250: nit: small refactoring

```suggestion
            const key
              = file.type.startsWith('audio/')
                ? 'state.attachments_limit_audio_error'
                : file.type.startsWith('video/')
                  ? 'state.attachments_limit_video_error'
                  : 'state.attachments_limit_unknown_error'
            const errorMessage = t(key, [formatFileSize(maxVideoSize)])
            failedAttachments.value = [
              ...failedAttachments.value,
              [file.name, errorMessage],
            ]
```

shuuji3 (approved):
Thanks

Issue #2148, opened 2023-06-06
feat: Threaded Writing


TGRush:
Threaded Writing is a feature which can be used to write out long messages without needing your instance to support a given character limit by splitting up your messages into multiple at once.

This is something that e.g Mammoth on iOS allows to do, which can make it much easier to write threads.

---

## Implementation suggestions

To implement this, I propose a couple of suggestions on things to look out for:

1. Allow the user to customize the way in which your message is split
  The User should be able to customize what string should be put at the beginning/end of each post/reply after being split to customize the way other people on the network see it in the end.
  Said string should have variables (e.g `%number%`) to signify how far you're currently in to reading a given thread.
  **This custom string would be optional.**

2. Editing a thread should be visual.
  The user should be able to see their text being split up into multiple posts/replies as part of the thread in one way or another. This is (in my opinion) best achieved by inserting the string into the normal compose window, and when the character limit is hit, offer a `+` button to put the overlapping text into the next post of the thread. Words at the beginning / end of the new / last post should have `...` at the end to signify the content continuing in the next reply.
  
3. There should be a "posting status" next to / in the *"Publish"* button to signify how many of the n amount of replies have been succesfully posted / replied already, until said action is finished.

---

I appreciate reading through this and all the work you're doing on making Elk as awesome as can be, thank you all very much, fellow developers and contributors! :heart:

hrw:
Or open another input box per each post (Twitter or Tweetdeck had it). This way each of them can be edited before posting and user has complete control without a need to learn extra things.

maybeanerd:
Is anybody working on this? I might just give it a try and propose an implementation.

maybeanerd:
Should be solved by https://github.com/elk-zone/elk/pull/2715 now

PR #2715, opened 2024-03-29
WIP: feat: add threaded drafts & posts

maybeanerd:
Based on #2148

# This is WIP!

I'm currently experimenting around and trying out different things, but I wanted to be transparent about it and open up the possibility for early feedback, which is why I'm openin this Draft PR.

Please feel free to provide feedback on everything, especially the part on implementation details/architecture

Here's my current plan:

## UX
- Add a "start thread" button next to the publish button. It will show the thread index of the post, as well as the total number of posts in the thread
- automatically prefix threaded posts with a `🧵 x/n` (maybe customizable? I would start with a fixed behavior)
- when adding a new post to a thread using the button, essentially another editor pops up below the existing one.
  - The existing ones publish button is removed.
  - The existing post can be removed from the thread again
- since each post in a thread has their own editor, they can each include media, polls, different language and different visibility settings
  - One could think about e.g. enforcing some of these to be the same across a thread, but IMO the freedom to do whatever is quite nice
- allow saving the entire thread as a single draft 

## technical architecture
After playing around with the state that this PR has as of now, I noticed that mixing drafts and threads is a real mess.
I think what I will try doing is make _all_ posting scenarios be a thread under the hood, but handle threads with length == 1 a bit differently (theyre just normal posts)

This should also allow saving an entire thread, or single posts, as distinct drafts that dont conflict by accident.


composables/masto/publish.ts
```diff
@@ -138,6 +138,8 @@ export function usePublish(options: {
     }
   }
 
+  // TODO return threadId and threadIndex to be used in the thread composer
```
maybeanerd, line 141: Outdated, remove

composables/thread.ts
```diff
@@ -0,0 +1,91 @@
+import type { Draft } from '~/types'
+
+/**
+ * The list of draft keys that are part of the thread.
+ * There is a single global thread for the whole app, to keep complexity low.
```
maybeanerd, line 5: Let's see if that's still the case when a draft can include an entire thread. Might be able to move this to the composables state

composables/thread.ts
```diff
@@ -0,0 +1,91 @@
...
+  if (baseDraftKey)
+    threadBaseDraftKey.value = baseDraftKey
+
+  /**
+   * Current length of the thread
+   */
+  const threadLength = computed<number>(() => Object.keys(threadDrafts.value).length)
+
+  /**
+   * Whether the thread is active (has at least one draft)
```
maybeanerd, line 26: Rn it's more than one, this explanation is outdated

maybeanerd (commented):
Just a few notes for myself

maybeanerd:
just a quick visual note of how this is shaping up to look
image

Still need to change the publish button, and I think I will also change the left side a bit (maybe make it look like threads in the timeline?)

maybeanerd:
So, I think visually this is something that kind of is okay right now, and the drafts include the entire thread now

I'll focus on actually being able to post the thread, and adding the thread count to the posts themselves, next

maybeanerd:
Here's a snapshot of the current state, but you should also be able to get this far in the preview deployment:
image

maybeanerd:
Looks like adding the emoji to the start of the messages somehow messes with tiptap? Without including an emoji, it works perfectly, but right now  you cant type anything into the message until you remove the existing text first. super weird, will look into it
image

userquin:
Maybe we need to patch some plugin, check these pr (tiptap updated):
- https://github.com/elk-zone/elk/pull/2686
- https://github.com/elk-zone/elk/pull/2655

maybeanerd:
> Maybe we need to patch some plugin, check these pr (tiptap updated):
> 
>     * [fix(ui): hashtags not working when composing #2686](https://github.com/elk-zone/elk/pull/2686)
> 
>     * [fix(ui): mentions not working when composing #2655](https://github.com/elk-zone/elk/pull/2655)

potentially 🤔 
I found a workaround for now though at least

maybeanerd:
I added another workaround, but I think you're right. To fix this for good, we need to look into the emoji handling configured in the tiptap composable.
TBH its a bit much for me now, if you already have an idea what might be the issue, feel free to suggest something. IMO this workaround is good enough (at least for now), but it's really not nice.

maybeanerd:
It seems to work quite well now.

Still missing is error handling. I think for now i'll go the simple route: whenever a status post would fail, the thread posting fails and stops. Already posted statuses are kept.

Still need to build even that though😅

Once that's in there, i'll mark this as ready for review - and I touched so much code that there's definitely gonna be more to change then 👀

shuuji3:
Thank you for working on this! The current UX looks pretty good. 😀

I think going the simple route is a good idea. Once the basic functionality is finished, the feature could be merged into Canary so that we can try and get feedback too, then we can incrementally improve the functionalities.

This is just a personal preference but I usually don't use 🧵-prefix so it would be nice if it's customizable. But it's totally fine not to include the customization in this PR. 🙂

maybeanerd:
basic error handling is now here: 
image

maybeanerd:
There's a few thing we can definitely do better, but I didn't want to fill this PR with too much. It already touches many things

I think in the future one could attempt to refactor the publishWidget to get entirely rid of usePublish in there and move that handling to the thread composable. The current solution kind of uses both, letting usePublish handle most of the per-status state, even if its part of a thread.

maybeanerd:
Ready for review ✨

maybeanerd:
With every change on main there are new conflicts and It tends to get really tedious to resolve at times - I re-applied all changes that touched the conflicting files now (please squash merge in the end 😁 )

Could we get this PR reviewed rather sooner than later, to then be able to iterate and not worry about losing changes from other PRs accidentally?

shuuji3 (changes requested):
Sorry for letting you resolve code conflicts 🙏🏻 

I made a few comments.

There are two things I noticed when composing in the home timeline:
- When publishing the normal (non-threaded) post, the text in the editor should be cleared but now the text remains there.
- After publishing the threaded posts, the page is navigated to the last thread page. But Elk usually stays home timeline page without navigation. So it may be better to keep the same behavior. (but this one can be fixed by the following PR)

composables/masto/statusDrafts.ts
```diff
@@ -112,9 +112,14 @@ export function getReplyDraft(status: mastodon.v1.Status) {
   }
 }
 
-export function isEmptyDraft(draft: Draft | null | undefined) {
+export function isEmptyDraft(drafts: Array<DraftItem> | null | undefined) {
+  if (!drafts)
+    return true
+
+  const draft = drafts.at(0)
```
shuuji3, line 119: This line raises an exception if we have some existing draft value in the local storage written before this change.

I think this is due to the change in the data structure of the draft.

Here's the example `elk-drafts` value:

image

As you can see, the old structure (Draft object) and the new one (array of Draft object) are mixed. Is it possible to handle both old and new values to ensure backward compatibility?

maybeanerd: Oh yes good find! I wanted to, but totally forgot testing it with existing drafts. Will try to make it compatible

maybeanerd: I think I fixed it, but please feel free to double check with your real examples to be sure 👍🏻 

I also figured out some reactivity issues with drafts, which should now be fixed as well

maybeanerd:
whoops i forgot to check the two thing you mentioned outside of the comments
-> 
    When publishing the normal (non-threaded) post, the text in the editor should be cleared but now the text remains there.
    After publishing the threaded posts, the page is navigated to the last thread page. But Elk usually stays home timeline page without navigation. So it may be better to keep the same behavior. (but this one can be fixed by the following PR since only affects to the new draft feature)

those are not yet fixed. will do soonTM

composables/masto/statusDrafts.ts
shuuji3: Thanks! I confirmed the fix now 👍🏻

shuuji3:
That makes sense. Thank you for the additional change!

shuuji3 (approved):
I think this new thread feature is good enough now to test in the Canary Elk. 😃

patak-dev:
This is great! Thanks a lot for this feature 🙌🏼

image

Before we merge:
- I think we should have a tooltip in the Start thread button (like the one other buttons have next to it)
- We should remove the "🧵 x/N" from each message. The thread emoji is equivalent to low quality posts in a lot of circles in social. Grifters that abused it to share the exact same "The best 15 CSS resources! Links in threads!!!1!1!". Let's keep it clean. Each user can decide to add a marker. I prefer threads without any x/N for example, as Elk already properly order posts. And the messages look a lot better once they are boosted deep in the thread.

maybeanerd:
Fully agree with the tooltip, will do!

Concerning the prefix, I thought we'd later add a setting and make it optional - but we can also entirely remove it, I included it mostly because it was requested in the original issue

patak-dev:
I don't think the prefix will have enough demand to justify a setting (that it would also involve configuring the emoji, format, etc). Let's keep it simple and remove it 👍🏼

maybeanerd:
Done!

patak-dev:
Nice! Let's merge it 🚀

One detail, I think the current remove message button is a bit confusing. You can't remove the last message in the thread for example. And the delete button being in the bottom-right makes it closer to the next message than the one that is going to be removed. I think it would be better to have a new red cross circular button in the top-right of each message, and then the last one can also be removed.

PR #2720, opened 2024-03-30
fix(ui): wrong polls are removed

lazzzis:
The removed options are assigned to `options` by mistake.

before:

https://github.com/elk-zone/elk/assets/10359255/840e2c85-b345-494a-a796-bd83754ddcbd

after:
 
https://github.com/elk-zone/elk/assets/10359255/b13ecbdd-d19e-479d-b408-8b4daa15b6d3


patak-dev (approved):
Ha, nice bug. Thanks!

PR #2728, opened 2024-04-01
feat: show background color for direct post to reduce mistake

### LINKED ISSUE #2723: Show background color when sending "Direct" message
When sending a post with "Direct" visibility, I'm sometimes a bit worried that I'm sending a public post by mistake because the only indicator for "Direct" is the `@` icon next to the publish button.

image

It would be nice to have the same background color as the posted message in the editor too, so I can be more confident that the message won't be a public one. Also, it matches the style of the post result in the same way as the bold text is **bold** in the editor (WYSIWYG).

image

shuuji3:
resolves #2723

## Screenshots
Screenshot from 2024-03-31 18-18-35

Also, it properly expands the background when it has many lines:
Screenshot from 2024-03-31 18-22-17


shuuji3:
Sorry, I was using the wrong branches 😅

PR #2730, opened 2024-04-01
fix: improve keyboard operability especially on search page and editor

shuuji3:
This PR contains two improvements for keyboard navigations:
- [fix: allow to escape from compose editor and search input by esc key](https://github.com/elk-zone/elk/commit/4c2e38adb1229e3fad8fac673fa070893b55f5e0)
- [fix: allow focus to search input within search page](https://github.com/elk-zone/elk/commit/08a8ab62c6adad3cab0d9bfae21f663d18991c8a)

The first one is to allow using the ESC key to unfocus the editor and search inputs. 

This is the default behavior for the native HTML input/textarea elements so users expect that the ESC key works as usual, but Elk uses the custom elements that prevent navigating with the ESC key and it needs to use the Tab key to go to outside input.

The second one is to fix the behavior of `/` key in the search page.

When typing `/` in other pages than the search page, Elk will navigate to the search page and then autofocus the search input element. But if we are already inside the search page, `/` won't let us focus on the input.

### DIFF
--- a/components/search/SearchWidget.vue
+++ b/components/search/SearchWidget.vue
@@ -82,6 +82,7 @@ function activate() {
         placeholder-text-secondary
         @keydown.down.prevent="shift(1)"
         @keydown.up.prevent="shift(-1)"
+        @keydown.esc.prevent=" input?.blur()"
         @keypress.enter="activate"
       >
       <button v-if="query.length" btn-action-icon text-secondary @click="query = ''; input?.focus()">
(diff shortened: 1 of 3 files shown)


components/search/SearchWidget.vue
```diff
@@ -82,6 +82,7 @@ function activate() {
         placeholder-text-secondary
         @keydown.down.prevent="shift(1)"
         @keydown.up.prevent="shift(-1)"
+        @keydown.esc.prevent=" input?.blur()"
```
patak-dev, line 85:
```suggestion
        @keydown.esc.prevent="input?.blur()"
```

patak-dev (approved):

PR #2732, opened 2024-04-01
feat: show background color for direct post to reduce mistake

### LINKED ISSUE #2723: Show background color when sending "Direct" message

shuuji3:
resolves #2723

## Screenshots
Screenshot from 2024-03-31 18-18-35

Also, it properly expands the background when it has many lines:
Screenshot from 2024-03-31 18-22-17

### DIFF
--- a/components/publish/PublishWidget.vue
+++ b/components/publish/PublishWidget.vue
@@ -158,6 +158,8 @@ const isExceedingCharacterLimit = computed(() => {
 
 const postLanguageDisplay = computed(() => languagesNameList.find(i => i.code === (draft.value.params.language || preferredLanguage))?.nativeName)
 
+const isDM = computed(() => draft.value.params.visibility === 'direct')
+
 async function handlePaste(evt: ClipboardEvent) {
   const files = evt.clipboardData?.files
   if (!files || files.length === 0)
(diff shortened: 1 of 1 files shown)


patak-dev (approved):
Ah, this was such an important missing piece 🙌🏼

patak-dev:
Trying to see why the `elkdev` user doesn't work locally for me 🤔 
In the meantime, after using this for a bit, the styles in this PR are good for the main editor. I was thinking if we should have a bit of gradient to avoid the feeling of going to send a bunch of empty lines.

image

There is also some missing bottom padding for the reply editor in threads 

image

And the same missing bottom padding in both the main and thread editors when it reached the expansion point

image

shuuji3:
Good point.👍🏻  If the message is a single line, I think the background should be the same height as the posted message. And, the bottom margin/padding seems to be less than the top one. I'll adjust this unbalanced padding/margin.

PR #2733, opened 2024-04-01
feat: add "How to contribute?" link next to language status

shuuji3:
Currently, the language translation status only shows the percentage number of the translation status and it's a bit unclear which text is not translated and where to find the Elk documentation.

This will also help potential new translators find the entry point for contribution. 

## Screenshot

Japanese example:

Screenshot from 2024-04-01 12-36-11

### DIFF
--- a/locales/en.json
+++ b/locales/en.json
@@ -459,6 +459,7 @@
     },
     "language": {
       "display_language": "Display Language",
+      "how_to_contribute": "How to contribute?",
       "label": "Language",
       "post_language": "Posting Language",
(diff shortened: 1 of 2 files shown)


shuuji3:
I also wonder if it's a good idea to add the link to Elk documentation below the "Show intro"?
 
image

patak-dev (approved):
Awesome!

userquin:
json file being three shaked? (should be moved to a virtual module (Vite or Nuxt virtual module) ? )

Issue #1858, opened 2023-03-05
feat: keyboard shortcuts - v5 - timeline/thread navigation + actions


hartmut-co-uk:
note: follow up issue of https://github.com/elk-zone/elk/issues/39

### Planned Timeline / Thread navigation + actions keyboard shortcuts

- navigation (both in _timeline_ AND _status detail + replies thread_)
  - `j` Next status
  - `k` Previous status
- actions for _active_ (selected) status
  - `r` Reply
  - `b` Boost
  - `f` Favourite
  - ??

botolo:
Just to say that this will be the key feature that will make me move away from Twitter once and for all.

shuuji3:
I'm experimenting with `j`/`k` navigation in #2739. It may still take a while until it stabilizes.

PR #2739, opened 2024-04-01
feat: keyboard status navigation with `j`/`k`

shuuji3:
resolves `j`/`k` navigation part of #1858

### DIFF
--- a/components/magickeys/MagickeysKeyboardShortcuts.vue
+++ b/components/magickeys/MagickeysKeyboardShortcuts.vue
@@ -32,14 +32,14 @@ const shortcutItemGroups = computed<ShortcutItemGroup[]>(() => [
         description: t('magic_keys.groups.navigation.shortcut_help'),
         shortcut: { keys: ['?'], isSequence: false },
       },
-      // {
-      //   description: t('magic_keys.groups.navigation.next_status'),
-      //   shortcut: { keys: ['j'], isSequence: false },
-      // },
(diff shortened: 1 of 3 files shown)


shuuji3:
This implementation is working well if we disable the experimental virtual scrolling. When enabled and a post has a large height, the next post is not loaded until it is on-screen. As a result, this keyboard navigation code fails to find the next post to focus on and stack to the last post or is forced to go to the top post.

We are using `DynamicScroller` (https://github.com/Akryum/vue-virtual-scroller/blob/master/packages/vue-virtual-scroller/README.md#dynamicscroller) but it has little option to customize so I'm not sure how to fix this issue.

I'll disable this keyboard shortcut when the virtual scroller is enabled. But that means most users cannot use it as it is enabled by default.

shuuji3:
Disabled this feature when "Virtual Scrolling" is enabled and add instructions to do so in the shortcut help dialog.

image

patak-dev:
I'm not sure about adding j/k only when virtual scrolling is disabled. It may push users to disable it and we could get performance reports later on. Maybe we could wait to see if a way to have it always enabled comes up?

shuuji3:
Right, that's not an ideal situation. OK, let's revert the message to encourage disabling virtual scrolling and seek a solution.

patak-dev (approved):

shuuji3:
Thanks!

PR #2743, opened 2024-04-02
fix(ui): clicking on custom emoji does not navigate to status detail

lazzzis:
Clicking on emoji does not navigate to status detail page

Before:

https://github.com/elk-zone/elk/assets/10359255/c5255d95-3e69-479e-99b6-366764d8500b

After:

https://github.com/elk-zone/elk/assets/10359255/1066c285-e45a-48b9-bed2-98e248a07ac9


shuuji3 (approved):
Good catch! Thanks for the fix :+1:

PR #2751, opened 2024-04-04
fix(ui): media preview card is misaligned

lazzzis:
Not sure why both `w-full`(width: 100%) and `w-100vh`(width: 100vh) are applied to media preview card. Sometimes, `w-100vh` are overriding `w-full` on my devices, making the picture misaligned.

Plus, I guess `w-100vh` is a typo? Is it supposed to be `w-100vw`?

Screenshot 2024-04-03 at 6 01 47 PM

### DIFF
--- a/components/modal/ModalMediaPreview.vue
+++ b/components/modal/ModalMediaPreview.vue
@@ -37,7 +37,7 @@ onUnmounted(() => locked.value = false)
 </script>
 
 <template>
-  <div relative h-full w-full flex pt-12 w-100vh @click="onClick">
+  <div relative h-full w-full flex pt-12 @click="onClick">
     <button
       v-if="hasNext" pointer-events-auto btn-action-icon bg="black/20" :aria-label="$t('action.previous')"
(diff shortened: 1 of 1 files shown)


shuuji3:
Thank you, I also encountered the same issue in the past. 

To validate this fix, I'm trying several image posts but couldn't reproduce the issue yet. Could you share the example post that we can test or upload the sample image to `elkdev` account? 

Maybe this was an accidental change by this commit: https://github.com/elk-zone/elk/pull/749/files#diff-91eeec335ef839760a2f6d9a82acc438b6fc0e796f1d603b65c38c89f7945672R40

I tested the image swipe on Google Chrome Android but the swipe works as expected. So I think `w-100vh` was not necessary and this change is fine.

userquin (approved):

lazzzis:
Unfortunately, I am also unable to constantly repro this issue. The picture in the screenshot was randomly picked from the elkdev's timeline.

userquin:
I can reproduce the problem using the first link in this issue https://github.com/elk-zone/elk/pull/2752#issue-2224675483

shuuji3 (approved):
Thanks, understand. :+1:

PR #2752, opened 2024-04-04
feat(ui): improve gif support

lazzzis:
1. add `gif` indicator on gif media
2. clicking on `gif` can also enable the media preview dialog

example posts:
https://deploy-preview-2752--elk-zone.netlify.app/universeodon.com/@elkdev/112211964228804763

https://elk.zone/o3o.ca/@rikan@m.otter.homes/112061227739128715
https://elk.zone/mastodon.ktachibana.party/@furretya/112210841006362590

Before

https://github.com/elk-zone/elk/assets/10359255/d0424fff-6331-4185-bc16-3077b520d44e

After:

https://github.com/elk-zone/elk/assets/10359255/6742877e-4f1a-40b8-9a7e-1e4de4eca4e4

### DIFF
--- a/components/status/StatusAttachment.vue
+++ b/components/status/StatusAttachment.vue
@@ -68,6 +68,7 @@ const video = ref<HTMLVideoElement | undefined>()
 const prefersReducedMotion = usePreferredReducedMotion()
 const isAudio = computed(() => attachment.type === 'audio')
 const isVideo = computed(() => attachment.type === 'video')
+const isGif = computed(() => attachment.type === 'gifv')
 
 const enableAutoplay = usePreferences('enableAutoplay')
 
@@ -164,7 +165,7 @@ watch(shouldLoadAttachment, () => {
       <button
         type="button"
         relative
-        @click="!shouldLoadAttachment ? loadAttachment() : null"
+        @click="!shouldLoadAttachment ? loadAttachment() : openMediaPreview(attachments ? attachments : [attachment], attachments?.indexOf(attachment) || 0)"
       >
         <video
           ref="video"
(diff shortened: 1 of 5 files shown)


components/status/StatusAttachment.vue
```diff
@@ -281,6 +283,14 @@ watch(shouldLoadAttachment, () => {
           </div>
         </template>
       </VDropdown>
+      <div v-if="isGif && !getPreferences(userSettings, 'hideGifIndicatorOnPosts')">
+        <button
+          aria-hidden font-bold text-sm
+          rounded-1 bg-black:65 text-white px1.2 py0.2 pointer-events-none
+        >
+          {{ $t('status.gif') }}
```
shuuji3, line 291: Maybe we don't have to allow "GIF" to be translated? Because this is the proper noun of file format so I think all languages will use "GIF" as is.

shuuji3 (changes requested):
Awesome, this is a great improvement for GIF support! 🙂 

I think this is a good style that matches the official Mastodon UI and other clients. 
 
Screenshot from 2024-04-04 17-10-02

Also, the small badge can avoid hiding the GIF image unnecessarily so users can view the meaning of the image even when disabling the animation.

I tested two hiding indicator settings as well as disabling the autoplay setting and all settings are working correctly.

components/status/StatusAttachment.vue
patak-dev: yes, I think we don't need to translate GIF

locales/en.json
```diff
@@ -614,6 +615,7 @@
     "favourited_by": "Favorited By",
     "filter_hidden_phrase": "Filtered by",
     "filter_show_anyway": "Show anyway",
+    "gif": "GIF",
```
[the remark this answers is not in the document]
shuuji3: Thanks, I didn't know that. In that case, I think it's better to keep this change for Arabic users.

shuuji3 (approved):

PR #2754, opened 2024-04-04
chore(ui): use full width in settings toggle items

userquin:
Check dev server, w-full is before w-fit, sometimes the styles are reversed and the toogle items don't fill the full width:

imagen


shuuji3 (approved):
I confirmed the fix with dev server on the Stackblitz.👍🏻

This issue doesn't happen on the production build by `pnpm build && pnpm start`.

PR #2761, opened 2024-04-05
feat: allow choosing favorite buttons in bottom navigation bar

### LINKED ISSUE #2757: Customizable bottom nav bar buttons on mobile
This is the task tracking issue to resolve #641.

The goal of this new feature is to allow users to select their preferred buttons in the bottom nav bar on mobile view. The original issue proposed to move the three-dot button somewhere but that's out of scope and this will focus on the button selection/ordering.

## ToDo

- [x]  Extract the bottom nav buttons into independent components 
- [x]  Implement a settings page to arrange the order of the buttons 
- [ ]  ~~Add other types of buttons for the selection~~ postpone to a future PR

### LINKED ISSUE #2758: Extract the bottom nav buttons into independent components
See #2757 for the details.

shuuji3:
resolves #2757

### DIFF
--- a/components/nav/NavBottom.vue
+++ b/components/nav/NavBottom.vue
@@ -1,12 +1,12 @@
 <script setup lang="ts">
+import { NavButtonExplore, NavButtonFederated, NavButtonHome, NavButtonLocal, NavButtonMention, NavButtonNotification, NavButtonSearch } from '#components'
+
+const navButtons = currentUser.value
+  ? [NavButtonHome, NavButtonSearch, NavButtonNotification, NavButtonMention]
+  : [NavButtonExplore, NavButtonLocal, NavButtonFederated]
+
 // only one icon can be lit up at the same time
-import { STORAGE_KEY_LAST_ACCESSED_EXPLORE_ROUTE, STORAGE_KEY_LAST_ACCESSED_NOTIFICATION_ROUTE } from '~/constants'
-
 const moreMenuVisible = ref(false)
-
-const { notifications } = useNotifications()
-const lastAccessedNotificationRoute = useLocalStorage(STORAGE_KEY_LAST_ACCESSED_NOTIFICATION_ROUTE, '')
-const lastAccessedExploreRoute = useLocalStorage(STORAGE_KEY_LAST_ACCESSED_EXPLORE_ROUTE, '')
 </script>
 
 <template>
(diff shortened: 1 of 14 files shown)


components/nav/NavBottom.vue
```diff
@@ -1,12 +1,12 @@
...
+import { NavButtonExplore, NavButtonFederated, NavButtonHome, NavButtonLocal, NavButtonMention, NavButtonNotification, NavButtonSearch } from '#components'
 
 const moreMenuVisible = ref(false)
 
-const { notifications } = useNotifications()
-const lastAccessedNotificationRoute = useLocalStorage(STORAGE_KEY_LAST_ACCESSED_NOTIFICATION_ROUTE, '')
-const lastAccessedExploreRoute = useLocalStorage(STORAGE_KEY_LAST_ACCESSED_EXPLORE_ROUTE, '')
+const navButtons = currentUser.value
+  ? [NavButtonHome, NavButtonSearch, NavButtonNotification, NavButtonMention]
+  : [NavButtonExplore, NavButtonLocal, NavButtonFederated]
```
shuuji3, line 9: These buttons will be customizable by the following PRs.

components/settings/SettingsBottomNav.vue
```diff
@@ -0,0 +1,81 @@
...
+  >
+    <button btn-base i-ri:close-fill text-primary mx-4 @click="remove" />
+  </nav>
+
+  <!-- button selection -->
+  <div flex="~ gap4 wrap" py4>
+    <button
+      v-for="{ name, label, icon } in availableNavButtons"
+      :key="name"
+      btn-text flex="~ gap-2" p2 border="~ base rounded" bg-base ws-nowrap
```
userquin, line 69: maybe you can add `selected` option to `NavButton` to simply logic...

can you add these attributes?
```vue
      type="button"
      role="switch"
      :aria-checked="selectedNavButtonNames.includes(name)"
```

and remove `disable` attribute?

You also need to add remove logic...

userquin: I'll check it next weekend on my local, don't apply changes yet

shuuji3: Thanks for suggestion for button styling! 


I'm actually still explorering what kind of UI would be better overall. A bit more adjustment would be included until finalizing. Maybe I'll add some ordering button or better remove button etc. later for the easy arrangement.

userquin: Maybe an editable forn? Check push notification in notifications settings (you will needd to enable them, dont forget to remove once you've check it)

shuuji3: OK, I was able to implement the toggle button logic now. 🙂

shuuji3:
Now it's ready to review.

- Each button can add/remove the button to/from the nav bar.
- Added "Clear" button - remove all buttons from nav bar
- Added "Reset" button ()

 

Screenshot from 2024-04-12 17-07-30

patak-dev (approved):
This is incredible! A lot of people will be very happy when this feature is released.

The current way to show what buttons were available wasn't very clear to me, and I didn't discover easily that I should first clear then add the buttons to get the order I want. I don't have a better proposal that doesn't involve complex drag and drop so I think we should merge this for now. Users aren't going to use this all the time. If we have a better idea later, we can improve it.

shuuji3:
@patak-dev I just realized that in the normal use case, users want to create a different button arrangement than the current one so there is no need to show the current ones. #2803 will set the default selection empty.

Yes, the drag and drop pattern was exactly one of the candidates I was thinking about, but avoid it for now due to complexity for now. If I come up with a better UI, I'll make another PR.

PR #2762, opened 2024-04-05
chore: update github avatar images

shuuji3:
The command `pnpm update:team:avatars` actually did not actually update the existing avatar images.

I assume we can use the latest avatar images instead of keeping the previous images.

🖼 Preview: https://deploy-preview-2762--elk-zone.netlify.app/settings/about


userquin (approved):
:sweat_smile:

PR #2764, opened 2024-04-05
fix: adjust background spacing in direct message

### LINKED ISSUE #2749: Direct message background in editor has unbalanced styles
ref. https://github.com/elk-zone/elk/pull/2732#issuecomment-2031079772

shuuji3:
fix #2749

## Screenshots
image

Screenshot from 2024-04-06 01-18-58

### DIFF
--- a/components/publish/PublishWidget.vue
+++ b/components/publish/PublishWidget.vue
@@ -279,13 +279,13 @@ onDeactivated(() => {
           </ol>
         </CommonErrorMessage>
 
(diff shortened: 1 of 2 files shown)


patak-dev (approved):
This is great! Thanks a lot ❤️ 
Merging and let's do a release!

One detail for the far future, it may be interesting to see if we could style the DMs in a way that when we swap public/direct there is no layout shift for the message you are composing.

PR #2766, opened 2024-04-05
fix(ui,a11y): focus lost when navigating using the keyboard (tab)

userquin:
This PR removes all `no-auto-focus` attributes from `CommonTooltip` component usages, adding also `no-auto-focus` in `VMenu` and `CommonTooltip` to disable it to fix losing focus when using the keyboard.

Without this PR, check the video or open elk.zone and traverse the app using the tab key: https://streamable.com/5lngi3

With this PR: https://streamable.com/c0jpsj

### DIFF
--- a/components/account/AccountBotIndicator.vue
+++ b/components/account/AccountBotIndicator.vue
@@ -11,7 +11,7 @@ defineProps<{
     text-secondary-light
   >
     <slot name="prepend" />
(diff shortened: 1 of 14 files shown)


shuuji3 (approved):
Thanks, now auto-focus is working correctly. The cursor blinking issue was also resolved. 💯

PR #2767, opened 2024-04-05
fix(a11y): settings items without href

userquin:
Any anchor without href cannot be traversed with the keyboard.

This PR uses custom nuxt link and vue component for button when `button` attr configured: the `Show intro` "link" cannot be activated using the keyboard (maybe there is a shorcut)

### DIFF
--- a/components/settings/SettingsItem.vue
+++ b/components/settings/SettingsItem.vue
@@ -11,10 +11,27 @@ const props = defineProps<{
   large?: true
   match?: boolean
   target?: string
+  button?: boolean
 }>()
 
+defineEmits<{
+  (event: 'click'): void
+}>()
+
+const vm = getCurrentInstance()
 const router = useRouter()
 const scrollOnClick = computed(() => props.to && !(props.target === '_blank' || props.external))
+
+function focus() {
+  setTimeout(() => {
+    if (props.button)
+      vm?.vnode.el?.querySelector('button')?.focus()
+    else
+      vm?.vnode.el?.focus()
+  }, 100)
+}
+
+defineExpose({ focus })
 
 useCommand({
   scope: 'Settings',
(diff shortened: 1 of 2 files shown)


shuuji3:
> Any anchor without href cannot be traversed with the keyboard.

This is actually the correct behavior in browsers: 

> Some focusable HTML elements have a default tabindex value of 0 set under the hood by the [user agent](https://developer.mozilla.org/en-US/docs/Glossary/User_agent). These elements are an [<a>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a) or [<area>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/area) with href attribute, (...)

https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex#:~:text=a%20logical%20sequence.-,Some%20focusable%20HTML%20elements,-have%20a%20default

So I think we can fix the issue by adding `tabindex="0"` and the keydown handlers (in addition to `@click`) instead.

shuuji3:
> Any anchor without href cannot be traversed with the keyboard.

This is actually the correct behavior in browsers: 

> Some focusable HTML elements have a default tabindex value of 0 set under the hood by the [user agent](https://developer.mozilla.org/en-US/docs/Glossary/User_agent). These elements are an [<a>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a) or [<area>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/area) with href attribute, (...)

https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex#:~:text=a%20logical%20sequence.-,Some%20focusable%20HTML%20elements,-have%20a%20default

So maybe setting `tabindex="0"` and changing `<a>` to `<button>` can be a simpler way instead of handling focus event by JavaScript.

I also noticed that there is the attribute `disabled="false"` on `<SettingsItem>`. This is an invalid value and has the same effect as `disabled="true"` (ref. Explicitly set disabled="false" in the HTML does not work - Stack Overflow - https://stackoverflow.com/questions/32745276/explicitly-set-disabled-false-in-the-html-does-not-work/32745315#32745315). This can affect the focusable property. I'm going to make a PR to fix this attribute.

userquin:
The html markup is wrong, we're using an anchor when it is a button, check other pages, we dont use anchor for render theme color buttons

PR #2770, opened 2024-04-06
fix: set correct `disabled` property in `<a>` tag in `SettingsItem`

shuuji3:
ref. #2767


shuuji3:
Maybe this is a different issue. I couldn't reproduce `disabled="false"` rendering with the clean installation of nuxt@3.11.2:
https://stackblitz.com/edit/nuxt-starter-9e9zdh?file=app.vue

PR #2771, opened 2024-04-06
fix(a11y): allow focusing color mode buttons on interface setting

shuuji3:
Currently, the only active color mode button is focusable so it's impossible to select other color modes by the keyboard.

image

This PR removes `tabindex="-1"`, which prevents any keyboard focus, from the buttons.

## How to test
1. Open Preview: https://deploy-preview-2771--elk-zone.netlify.app/settings/interface
2. Select color mode buttons with <kbd>Tab</kbd> key.
3. Compare to the previous behavior: https://elk.zone/settings/interface


userquin:
This should be fixed in https://github.com/elk-zone/elk/pull/2809

PR #2772, opened 2024-04-06
fix(ui): move account lock and bot icons next to name in timeline

### LINKED ISSUE #2492: Own public toots are shown as "only follower"
I published a toot as public but in elk.zone it's show with a lock symbol, as if i tooted with only for my followers. In the normal mastodon client it's shown correctly, so elk also published it correctly. 

Also it seems as elk is not correctly remembering my settings for public/follower only viewability settings.

shuuji3:
fix #2492

As explained in the original issue, showing the account lock 🔒 icon on the status card is quite confusing:

Screenshot from 2024-04-06 01-40-26

I think this decision in the official Mastodon UI is reasonable because the meaning of 🔒 icon on Mastodon differs from that on Twitter.

On Twitter, 🔒 icon for the account has two meanings at the same time: the account requires to follow requests, and the visibility of all the posts is limited to "Only follower". So showing 🔒 icon on the status makes sense.

On the other hand, on Mastodon, the visibility of posts and account following limitations are separated. And the locked account still can publish publicly visible posts so 🔒 icon only indicates that the following approval is required when someone is following.

Mastodon already has the icon for visibility on the post (such as 🌎 and 🔒) so it should be enough to show 🔒 icon only on the account page (or other account-specific component) where users need to know whether the account requires the following request or not.

Although Mastodon UI doesn't show the bot icon too, I think we can keep it because the bot icon can show both meanings: the account is automated and the post itself is programmatically published by the bot.

## After
image

The status detail page still shows the lock indicator as the visibility indicator is clearly separated in different places without confusion.

Screenshot from 2024-04-06 17-24-40

Another example with the lock icon:

image

### DIFF
--- a/components/status/StatusAccountDetails.vue
+++ b/components/status/StatusAccountDetails.vue
@@ -16,6 +16,10 @@ const userSettings = useUserSettings()
     text-link-rounded
   >
     <AccountDisplayName :account="account" :hide-emojis="getPreferences(userSettings, 'hideUsernameEmojis')" font-bold line-clamp-1 ws-pre-wrap break-all />
-    <AccountHandle :account="account" class="zen-none" />
+    <div flex="~ gap-1">
+      <AccountHandle :account="account" class="zen-none" />
+      <AccountBotIndicator v-if="account.bot" text-xs />
+      <AccountLockIndicator v-if="account.locked" text-xs />
+    </div>
   </NuxtLink>
 </template>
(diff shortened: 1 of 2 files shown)


ayoayco:
To me it's more about the position of the icon. I think if the lock is for the account it should be visible & closer to the name/profile picture. Then maybe we can also show consistently in different profile views?

I wonder if we could have two rows in the status card similar as in the profile page...

the name + icons (at top) and username (bottom)

shuuji3:
Sorry for the late response! Right, it's better to have adjusted the icon position instead of reducing information by deleting this.


I think changing to two rows is a good option worth exploring. But that will change all post height and layout in a timeline, so I'm not sure how that UI change affects the current design. I'd like to create a separate PR for that change.

Here's new position for the lock icon. I also moved the bot icon as they should be in the same group (attributes of the author's account).

screenshot of status timeline showing bot and lock icons next to user name, not next to post time
(note: lock and bot icons are mocked for screenshot)

Cl0v1s:
Thanks for this !

PR #2776, opened 2024-04-07
fix(ui): make display name clickable in grouped follow notification

lazzzis:
The name is highlighted and so I assume this is supposed to be clickable.

Screenshot 2024-04-06 at 11 55 53 PM


shuuji3 (approved):
Good catch👍🏻 Assuming the highlighted text to be clickable makes sense.

PR #2778, opened 2024-04-07
fix(ui): fix z-index of horizontal line between statuses

shuuji3:
When focusing on the status, the horizontal line between statuses was above the focus indicator.

The keyboard navigation (currently implementing) always shows the border so this is especially noticeable.

## Before
Screenshot from 2024-04-07 23-44-48

## After
Screenshot from 2024-04-08 00-01-10


ayoayco (approved):
Change looks good 👍

PR #2779, opened 2024-04-07
feat: show pinned posts on individual account page

### LINKED ISSUE #305: missing info boxes in profile and pinned posts

shuuji3:
resolves #305

another attempt inspired by #1921 (by @lazzzis)

## Screenshot

Screenshot from 2024-04-08 02-06-43 (copy)


lazzzis:
Thanks for adding this! This looks awesome! A few concerns:

1. The timeline paginator will prepend "Show {v} new items" and append "Show origin sites" in certain scenarios. Will these two buttons show in pinned timeline as well?
2. the user can pin a replied comment. Does this screenshot look good to you? [link](https://deploy-preview-2779--elk-zone.netlify.app/universeodon.com/@detreville@discuss.systems)

Screenshot 2024-04-07 at 11 00 59 AM

shuuji3:
That's good points.

> 1. "Show {v} new items" and "Show origin sites" 

Probably the first "Show {v} new items" is only shown if a paginator has a bound `stream`. Both paginators (`pinnedPaginator` and `accountPaginator`) on this page are not using `stream`, so I think we don't have to worry about that case.

But I haven't tested the second case "Show origin sites", I'm going to check if it affects the pinned posts (maybe I need to hide that footer in the pinned timeline).

> 2. Does this screenshot look good to you?

This doesn't look good 😅 Let me try adjusting the position.

shuuji3:
I adjusted the position and font style to be consistent with other badges.

fix-a
thread-a

I also tested the different patterns (remove the margin when there is no thread line) but I think the initial version is consistent and better.

fix-b
thread-b

shuuji3:
> But I haven't tested the second case "Show origin sites", I'm going to check if it affects the pinned posts (maybe I need to hide that footer in the pinned timeline).

I checked this case but found out I already set `:end-massage="false"` in this line so this message won't be shown.

https://github.com/elk-zone/elk/pull/2779/files#diff-ea5c53d8a780a38a4c56028e9023299518d6e00488d1d03d1b8215f97dc3d36eR39

The "Show origin sites" text is defined in the `#done` slot in `TimelinePaginator`: https://github.com/elk-zone/elk/blob/3769176eaa45dfb9e995e7757d8c2c1235ba7f54/components/timeline/TimelinePaginator.vue#L42 but `CommonPaginator` does not show the slot if `engMessage` is `false`: https://github.com/elk-zone/elk/blob/59dda09cd479b14d8ccff843296b048b651784f3/components/common/CommonPaginator.vue#L113

ayoayco:
I think we need this too because I don't see the "see pinned posts" we used to have in the menu when viewing a profile if I remember it correctly.👀 Don't know when we lost that.

pages/[[server]]/@[account]/index/index.vue
```diff
@@ -10,11 +10,21 @@ const { t } = useI18n()
...
+  })
+}
+
 function reorderAndFilter(items: mastodon.v1.Status[]) {
   return reorderedTimeline(items, 'account')
 }
 
-const paginator = useMastoClient().v1.accounts.$select(account.id).statuses.list({ limit: 30, excludeReplies: true })
+const pinnedPaginator = useMastoClient().v1.accounts.$select(account.id).statuses.list({ pinned: true })
+const accountPaginator = useMastoClient().v1.accounts.$select(account.id).statuses.list({ limit: 30, excludeReplies: true })
```
Comment (sloanlance) [inline pages/[[server]]/@[account]/index/index.vue:27]: IMHO, I think better names for `accountPaginator` would be `timelinePaginator`, `postPaginator`, or `normalPaginator`.  Because it's not **_accounts_** that are being paginated.

sloanlance (approved):
:shipit: 

Overall, this looks like a good PR.  I'm no expert when it comes to Vue, however.

I think the paginator should be renamed, but it's a minor quibble.  I'm happy to see other people not only agree that pinned posts need to be shown, but that you have the knowledge to implement it and open a PR.  I hope @elk-zone approves this or even makes their own implementation.

@elk-zone:  Why not approve and merge this PR?  It's been waiting for about 7 months.

pages/[[server]]/@[account]/index/index.vue
Comment (shuuji3) [reply pages/[[server]]/@[account]/index/index.vue]: Thank you so much for the code review. I agree and updated to the `postPaginator`.

shuuji3:
Sorry for taking too long to update🙏🏻 That's partly because of the limited bandwidth of members who are holding many projects other than Elk.

I think it's fine to self-merge this PR at this moment as it's already well-reviewed, and there's no objection to adding support for this feature.

ayoayco:
Great job @shuuji3 🎉

PR #2798, opened 2024-04-12
feat(ui): add font size outline when focused

userquin:
When using the keyboard we don't know if the font size is focused:

imagen


shuuji3:
As another design, what do you think about the ring shape like in the color palette?

image

Or GitHub emoji button example:

image

But I think either shape looks good and the rectangular shape has no issue.

shuuji3 (approved):

userquin:
changing layout...

shuuji3 (approved):
Thanks for the update :pray: Looks great!

PR #2799, opened 2024-04-12
fix(a11y): add aria-* entries to interface settings

userquin:


shuuji3 (approved):
LGTM!

shuuji3:
This fix includes #2799

pages/settings/interface/index.vue
```diff
@@ -14,29 +14,29 @@ useHydratedHead({
       </div>
     </template>
     <div p6 flex="~ col gap6">
-      <div space-y-2>
-        <p font-medium>
+      <label space-y-2>
```
userquin, line 17: this should be a div, html markup is wrong, cannot add p, div, or button inside a label

shuuji3: Sorry, maybe I made a mistake during merge

shuuji3: Let me revert to `<div>`

userquin: iirc this should be fixed in another PR, maybe merge problem?

shuuji3: OK, fixed it now.

userquin:
I'll review the index.vue later, we should use some semantic tags like section moving the `p` inside the `SettingXXX` using some heading like h2 + some aria-* attr

userquin:
@shuuji3 if you are using mac, you can activate/deactivate VoiceOver with command+f5, check [Introducing VoiceOver](https://www.apple.com/voiceover/info/guide/_1121.html#:~:text=To%20turn%20on%20VoiceOver,%20press%20Command-F5.)

shuuji3:
@userquin Thanks for the info. I'll give it a try as well as Android's similar feature 👍🏻

PR #2801, opened 2024-04-12
feat(i18n): add spanish bottom navigation translations

userquin:
/cc @patak-dev @mrcego

### DIFF
--- a/locales/es.json
+++ b/locales/es.json
@@ -58,6 +58,7 @@
     "boost": "Retootear",
     "boost_count": "{0}",
     "boosted": "Retooteado",
+    "clear": "Limpiar",
     "clear_publish_failed": "Limpiar errores de publicación",
     "clear_save_failed": "Limpiar errores al guardar",
     "clear_upload_failed": "Limpiar errores de subida de archivos",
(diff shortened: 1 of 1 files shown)


locales/es.json
```diff
@@ -316,6 +317,7 @@
     "list": "Lista",
     "lists": "Listas",
     "local": "Local",
+    "more_menu": "Otras opciones",
```
mrcego, line 320: IMHO, we could translate it like "Más opciones", but also we could handle it inside `es-419.json`. Wdyt?

locales/es.json
```diff
@@ -450,6 +452,8 @@
       "label": "Ajustes de cuenta"
     },
     "interface": {
+      "bottom_nav": "Navegación inferior",
+      "bottom_nav_instructions": "Elija sus botones de navegación favoritos, hasta cinco para la navegación inferior. Debe incluir el botón \"Otras opciones\".",
```
mrcego, line 456: We are translating in informal way according the other keys, so we can say here "Elige tus botones..." and "Debes incluir...". Wdyt?

emanuelpina (approved):

PR #2807, opened 2024-04-13
fix(i18n): correct the spelling of Favorite in the US

at-wr:


shuuji3 (approved):
Thanks!

asmaloney:
Shouldn't this have been done in *[en-US.json](https://github.com/elk-zone/elk/blob/main/locales/en-US.json)*?

(I'm not sure how the localization works here, but there's no _en-CA.json_ for Canada, so with this change they get left with the USian spelling.)

shuuji3:
@asmaloney This is probably due to how Elk sets up the localization. The current localization setup needs at least one main file as the default fallback (in case other languages/variants have not yet been translated), and Elk simply uses `en.json` as the default. When other languages/variants need to use different languages or spelling, each locale can override the default string as needed. 

So the approach for `en-CA.json` in #2820 is totally correct. Thanks for your contribution!

PR #2810, opened 2024-04-13
chore: fix invalid html markup in `SettingsToggleItem`

userquin:


shuuji3 (approved):
Right, `<p>` and `<div>` are not allowed inside a `<button>` element in HTML.

PR #2811, opened 2024-04-13
feat(a11y): add semantic markup to preference settings

userquin:
This PR also fixes invalid html markup inside the heading (h2), cannot use div inside.

@shuuji3 can you check it with VoiceOver? The section headers should be displayed (I haven't installed NVDA on my laptop and my Mac is still hijacked by my nephew)

### DIFF
--- a/pages/settings/preferences/index.vue
+++ b/pages/settings/preferences/index.vue
@@ -15,162 +15,172 @@ const userSettings = useUserSettings()
         {{ $t('settings.preferences.label') }}
       </h1>
     </template>
-    <SettingsToggleItem
-      :checked="getPreferences(userSettings, 'hideAltIndicatorOnPosts')"
-      @click="togglePreferences('hideAltIndicatorOnPosts')"
-    >
-      {{ $t('settings.preferences.hide_alt_indi_on_posts') }}
-    </SettingsToggleItem>
-    <SettingsToggleItem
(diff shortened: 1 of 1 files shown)


pages/settings/preferences/index.vue
```diff
@@ -15,162 +15,172 @@ const userSettings = useUserSettings()
...
-      :checked="getPreferences(userSettings, 'experimentalUserPicker')"
-      @click="togglePreferences('experimentalUserPicker')"
-    >
-      {{ $t('settings.preferences.user_picker') }}
-      <template #description>
-        {{ $t('settings.preferences.user_picker_description') }}
-      </template>
-    </SettingsToggleItem>
+    <section>
+      <h2 px6 py4 mt2 font-bold text-xl flex="~ gap-1" items-center sr-only>
```
userquin, line 19: rn using sr-only here, maybe we should add a new i18n entry `General Preferences` : if sr-only is ok then we should remove the icon

shuuji3: Maybe adding `General Preferences` is more consistent so that every section can have a header. VoiceOver behavior sounds good so either is fine for now.

shuuji3 (approved):
I finally have time to learn how to use VoiceOver and I could navigate the preference page without any issue :+1:

PR #2817, opened 2024-04-14
feat(i18n): use same languages names in display and translations sections (settings/language)

### LINKED ISSUE #2814: Language naming issue at Hide translation block
In Settings > Language > At the 'Hide specific translations' block Hungarian language starts with a lowercase letter, need to replace Uppercase letter. I did not found it in the hu-HU.json file. Screenshot attachment also available. 

Screenshot 2024-04-14 020925

Thank you!

userquin:
This PR includes a new virtual module to use it instead `iso-639-1` dependency: since translations showing only configured languages in `conf/i18n.ts` module, we just use them via the virtual module.

We cannot remove `iso-639-1` dependency, it is being used in `composables/language.ts` (also renamed) and publish widgets (languages and fuse).

We can decide if we change the languages names to use `Intl.DisplayNames/iso-639-1` ones: for example including only country names when added to the country variant entry in `conf/i18n.ts` module.

<details>
<summary><strong>Language Names with Country Variants in Firefox<strong></summary>

imagen

```ts
console.table([ 'ar-EG',
  'ca-ES',
  'ca-valencia',
  'ckb',
  'cs-CZ',
  'de-DE',
  'en-GB',
  'en-US',
  'es-419',
  'es-ES',
  'eu-ES',
  'fa-IR',
  'fi-FI',
  'fr-FR',
  'gl-ES',
  'hu-HU',
  'id-ID',
  'it-IT',
  'ja-JP',
  'ko-KR',
  'nl-NL',
  'pl-PL',
  'pt-BR',
  'pt-PT',
  'ru-RU',
  'th-TH',
  'tl-PH',
  'tr-TR',
  'uk-UA',
  'vi-VN',
  'zh-CN',
  'zh-TW'
].reduce((acc, l) => {
  acc[l] = new Intl.DisplayNames(
    [l],
    { type: 'language', languageDisplay: 'standard', localeMatcher: 'lookup' },
  ).of(l)
  return acc
}, {}))
```
</details>

closes #2814


userquin:
Maybe we can use use `locales` from `useI18n()` directly...

shuuji3 (commented):
I just noticed that we can use **lang+country** code (e.g. `en-US`) for the "Display Language" section but we should use only **lang** code (e.g. `en`) without country variants for the "Hide Translation" setting.

This is because Mastodon's status object only has two characters of lang code such as `en` and `es` so we can only filter by lang code.

ref. Status - Mastodon documentation - https://docs.joinmastodon.org/entities/Status/#language

example:

```js
{
  ...
  id: "112225445529039980",
  content: "<p>Hello</p>",
  language: "en",
  media_attachments: [],
  ...
}
```

## Before
see: https://main.elk.zone/settings/language

Screenshot from 2024-04-15 22-27-19
Screenshot from 2024-04-15 22-34-55

## After
see: https://deploy-preview-2817--elk-zone.netlify.app/settings/language

Screenshot from 2024-04-15 22-32-29

shuuji3:
> Maybe we can use use `locales` from `useI18n()` directly...

If this can simplify code, maybe it's a good idea?

shuuji3:
Also, do we want to use the new function `getDisplayName()` in the "Display Language" section too?

https://github.com/elk-zone/elk/blob/55037f04cda5c9b5fbbc491464dedece73d65386/components/settings/SettingsLanguage.vue#L10-L16

```diff
+import { getDisplayName } from 'virtual:iso-639-1'

...

-       {{ item.name }} 
+       {{ getDisplayName(item.code) }}

userquin:
> Also, do we want to use the new function `getDisplayName()` in the "Display Language" section too?

No, those locales comes from `conf/i18n.ts` module and so the name is the name configured there.

shuuji3:
OK, will we adjust `countryLocaleVariants` in `config/i18n.ts` after this change? I thought using the same method could provide consistent naming.

shuuji3:
We could also consider other languages in the "Hide translation" list (please refer to #2843). Although there is no need to include them in this PR, it may affect how we organize the language names.

PR #2824, opened 2024-04-16
feat(ui): add manage list at the end of the lists

### LINKED ISSUE #2819: [Feature request] show something when there is no list
When the current user hasn't created any list yet, clicking on list button will show an empty box.

https://github.com/elk-zone/elk/assets/10359255/d3451686-7295-4280-b77b-0678507c9861

It would be perfect if we can show something here.

How about adding a 'manage list' button of navigating to `[[server]]/lists` ?

|w/o list|w/ list|
|-|-|
|Screenshot 2024-04-14 at 2 14 50 PM|Screenshot 2024-04-14 at 2 15 39 PM|

lazzzis:
Fix #2819 

The implementation is to replace the end message with a styled button, which navigates to `[[server]]/lists`

### DIFF
--- a/components/list/Lists.vue
+++ b/components/list/Lists.vue
@@ -30,7 +30,7 @@ async function edit(listId: string) {
 </script>
 
 <template>
-  <CommonPaginator :end-message="false" :paginator="paginator">
+  <CommonPaginator end-message :paginator="paginator">
     <template #default="{ item }">
       <div p4 hover:bg-active block w="100%" flex justify-between items-center gap-4>
         <p>{{ item.title }}</p>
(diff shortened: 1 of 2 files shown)


components/list/Lists.vue
```diff
@@ -30,7 +30,7 @@ async function edit(listId: string) {
 </script>
 
 <template>
-  <CommonPaginator :end-message="false" :paginator="paginator">
+  <CommonPaginator end-message :paginator="paginator">
```
shuuji3, line 33:
```suggestion
  <CommonPaginator :paginator="paginator">
```

It seems that `end-message` defaults to `true`. Omitting this property still has the same effect but can be simpler.

https://github.com/elk-zone/elk/blob/59dda09cd479b14d8ccff843296b048b651784f3/components/common/CommonPaginator.vue#L15

shuuji3 (changes requested):
Thank you, it works well 🙂

Screenshot from 2024-04-16 12-51-04

I just left one small suggestion.

shuuji3 (approved):
LGTM :+1:

components/list/Lists.vue
shuuji3: Thanks for the update!

PR #2842, opened 2024-04-22
fix(ui): fix publish dialog layout

### LINKED ISSUE #2834: Thread layout is off in compose dialog
Version: Canary (1eaaa6c@main)

There is no issue with the Compose page, but the Compose dialog (open by `c` key) has the wrong layout for threads. The initial layout has a small space on the right side. The cause might be the same.

## Screenshots

Screenshot from 2024-04-19 13-06-31

When click the thread button, new publish widget is appended to the right:

Screenshot from 2024-04-19 13-06-43

shuuji3:
fix #2834

The layout is off due to the change during thread implementation. This dialog was not assumed to have multiple widgets either.

Note: This publish dialog can be opened by typing <kbd>c</kbd> key.

## Before

Screenshot from 2024-04-22 18-22-19
Screenshot from 2024-04-22 18-22-12

## After
Screenshot from 2024-04-22 18-21-42

Screenshot from 2024-04-22 18-21-51

### DIFF
--- a/components/modal/ModalContainer.vue
+++ b/components/modal/ModalContainer.vue
@@ -63,7 +63,7 @@ function handleFavouritedBoostedByClose() {
     </ModalDialog>
     <ModalDialog
       v-model="isPublishDialogOpen"
-      max-w-180 flex
+      max-w-180 flex flex-col
       @close="handlePublishClose"
     >
       <!-- This `w-0` style is used to avoid overflow problems in flex layouts，so don't remove it unless you know what you're doing -->


patak-dev (approved):

patak-dev:
I wonder if `c` should go to the compose screen. I'm not using the shortcut, but even `g c` opens the dialog.

patak-dev:
Maybe `c` could open the Compose dialog, but `g c` the compose page 🤔

shuuji3:
That sounds good. We have `g c` for the conversation page, but for some issues, `c` is always prioritized and not working now 😅  https://github.com/elk-zone/elk/blob/7ac81d55e948e57b91f54ff132e8d23f5c9d7e59/plugins/magic-keys.client.ts#L43-L44

I think when we fix this issue, we can use `g c` for the compose page and assign another key for the conversation page. Parhaps, we can use `g d` (direct) or `g m` (message).

PR #2848, opened 2024-04-23
fix(ui): adjust z-index of Embedded Media Player and sticky header

### LINKED ISSUE #2847: Play button of Embedded Media Player is above header and bottom navbar
Version: canary (8ad05df@main)
OS: Android
Browser: Google Chrome (PWA)

Observed with this post: https://main.elk.zone/flipboard.social/@mike/112315842982477103

[]()

It's also over the manu panel.

shuuji3:
fix #2847 by adjusting z-index values

## After fix

Screenshot from 2024-04-23 22-56-02

Screenshot from 2024-04-23 22-56-44

### DIFF
--- a/components/search/SearchWidget.vue
+++ b/components/search/SearchWidget.vue
@@ -90,7 +90,7 @@ function activate() {
       </button>
     </div>
     <!-- Results -->
-    <div left-0 top-11 absolute w-full z10 group-focus-within="pointer-events-auto visible" invisible pointer-events-none>
+    <div left-0 top-11 absolute w-full z-10 group-focus-within="pointer-events-auto visible" invisible pointer-events-none>
       <div w-full bg-base border="~ base" rounded-3 max-h-100 overflow-auto py2>
         <span v-if="query.trim().length === 0" block text-center text-sm text-secondary>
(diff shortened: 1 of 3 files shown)


components/search/SearchWidget.vue
```diff
@@ -90,7 +90,7 @@ function activate() {
       </button>
     </div>
     <!-- Results -->
-    <div left-0 top-11 absolute w-full z10 group-focus-within="pointer-events-auto visible" invisible pointer-events-none>
+    <div left-0 top-11 absolute w-full z-10 group-focus-within="pointer-events-auto visible" invisible pointer-events-none>
```
shuuji3, line 93: This is not related to the fix but all other z-index have the prefix `z-` so use the same format for better code discoverability.

ayoayco:
Awesome 👏 thanks @shuuji3

ayoayco (approved):
Changes look good to me 👍

antfu:
I think later we could define more meaningful tokens like: https://github.com/nuxt/learn.nuxt.com/blob/1067f60883af002f8b08674faa79dfd4919fb302/uno.config.ts#L25-L29

PR #2873, opened 2024-05-27
chore: set package-manager-strict to false

Shinigami92:
like in vite, set package-manager-strict to false so pnpm v9 can locally be used

### DIFF
--- a/.npmrc
+++ b/.npmrc
@@ -1,3 +1,4 @@
 shamefully-hoist=true
 shell-emulator=true
 ignore-workspace-root-check=true
+package-manager-strict=false


antfu:
Any issue you encountered with it makes you want to disable it? For me it works fine

Shinigami92:
> Any issue you encountered with it makes you want to disable it? For me it works fine

It only happens if you use a pnpm version not matching the exact needed version (not using corepack by the way)

antfu:
So why not corepack? In a way I would encourage every contributor to enable corepack (I believe that's also why pnpm add this)

Shinigami92:
Why did we allowed it in Vite but not here?
https://github.com/vitejs/vite/pull/16565

This setting does not affect corepack users, but only makes life easier for non corepack users

antfu:
I don't hold a strong opinion on merging this PR - if the others agree, I think we could move on.

But in a way, I think there are true benefits to asking everyone to enable corepack. Otherwise, I'd say it's better to challenge ppm to revert that option with your counterpoints, instead of disabling it manually in every project. Would you agree?

Shinigami92:
> I don't hold a strong opinion on merging this PR - if the others agree, I think we could move on.
> 
> 
> 
> But in a way, I think there are true benefits to asking everyone to enable corepack. Otherwise, I'd say it's better to challenge ppm to revert that option with your counterpoints, instead of disabling it manually in every project. Would you agree?

I did not try to fight yet against pnpm with an issue, but I see nearly in all my related projects that this config was made and so I might report a list to pnpm-repo later on.
That config also just converts an error to a warning. Maybe that warning should be the default instead of hard-erroring and holding back normal developing.

patak-dev:
I don't have a strong opinion on this one either side.

Shinigami92:
Then lets merge it for now, as it does not have negative effects
If pnpm one day switches to be more non-destructive in this case, we can revert the setting

PR #2875, opened 2024-05-27
feat: generate alt-text with ai

Shinigami92:
This feature adds a Copilot-AI to the image-alt-text editor so it is possible to click a button that just generates an example text for the image.

https://github.com/elk-zone/elk/assets/7195563/47e99e73-7fe2-4234-bdca-672016be0636

Open TODOs:

- [ ] Style the button so it is inside the textarea in the upper right corner and looks like in VSCode commit title field
- [ ] Add a warning to inform the user that this feature will download ~250 MiB into the browsers cache
  _The download will only be done when the user confirms_
- [ ] Display error messages if something went wrong
- [ ] `props.attachment.url` might not be usable due to cors error?!

### DIFF
--- a/components/publish/PublishAttachment.vue
+++ b/components/publish/PublishAttachment.vue
@@ -23,6 +23,8 @@ const description = ref(props.attachment.description ?? '')
 
 const generationInProgress = ref(false)
 
+const userSettings = useUserSettings()
+
 async function generateAltText() {
   // eslint-disable-next-line no-console
   console.log(JSON.parse(JSON.stringify(props)))
@@ -35,7 +37,18 @@ async function generateAltText() {
   if (generationInProgress.value)
     return
 
-  // TODO @Shinigami92 2024-05-27: Show confirm dialog warning message that a model with ~250MiB will be downloaded
+  const experimentalAltTextGeneration = getPreferences(userSettings.value, 'experimentalAltTextGeneration')
+
+  if (!experimentalAltTextGeneration) {
+    // TODO @Shinigami92 2024-05-28: Use a fancy dialog instead of the browser's alert
+    // eslint-disable-next-line no-alert
+    const allow = confirm('This will download a model with ~250MiB. Do you want to continue? This is an experimental feature and might fail in several scenarios.')
+
+    if (!allow)
+      return
+
+    togglePreferences('experimentalAltTextGeneration')
+  }
 
   generationInProgress.value = true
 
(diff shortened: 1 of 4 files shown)


Shinigami92:
I run out of ideas how to get the image (either as dataUrl or blob) as every attempt results in cors issues 🙁

components/publish/PublishAttachment.vue
```diff
@@ -21,6 +21,74 @@ const maxDescriptionLength = 1500
...
+    imageElement.crossOrigin = 'Anonymous'
+
+    const dataUrl = new Promise<string>((resolve) => {
+      imageElement.onload = () => {
+        const canvas = document.createElement('canvas')
+        canvas.width = imageElement.width
+        canvas.height = imageElement.height
+
+        const ctx = canvas.getContext('2d')!
+        ctx.drawImage(imageElement, 0, 0)
```
userquin, line 58: can you try adding `await imgElement.decode();` before drawing image?

userquin: convert it to promise based 😄

components/publish/PublishAttachment.vue
```diff
@@ -21,6 +21,74 @@ const maxDescriptionLength = 1500
...
+
+        // TODO @Shinigami92 2024-05-28: Fix "Uncaught DOMException: Failed to execute 'toDataURL' on 'HTMLCanvasElement': Tainted canvases may not be exported."
+        const dataUrl = canvas.toDataURL(`image/${url.split('.').pop()!}`)
+
+        resolve(dataUrl)
+      }
+    })
+
+    imageElement.src = `${url}?request-with-cors`
```
userquin, line 67: move this logic before the promise and add the decode logic just after this line

userquin: I think you don't need that logic inside the promise, the function is async

Shinigami92:
It looks like the cors error is only on https://media.universeodon.com, but when I use my own account from mas.to I dont get a cors error on https://media.mas.to and it works 👀

userquin:
NOTE: maybe we can store the image temporarily in cache storage (similar to the model) instead of directly uploading the image to the server (we would also use the image from the cache in the preview and the dialog), this way, CORS problem will disappear (we'll need to change the publish logic)

Shinigami92:
Just as a note for FFR: I still have the idea in my head, however I observe the evolution of `window.ai` that is currently in experimental in chrome (e.g. https://syntackle.com/blog/window-ai-in-chrome/)
However, this is only for text and does not allow image as input... at least now...

Shinigami92:
Still interested in this feature some day, but maybe we need to wait some more earth rotations until browsers integrate AI into them self officially and stable

PR #2887, opened 2024-06-17
docs: add crab.bumscode.com to list of community elk deployments

maybeanerd:
this adds our deployment of elk to the list of community deployments

Not sure if it's okay, as it's a soft fork and not vanilla elk, but I thought It'd be nice to show that you can even deploy a slight variation of elk without much effort


patak-dev (approved):
Awesome! ❤️

Issue #2893, opened 2024-06-24
Aspect ratio avatars when a lot of people favorite your post is wrong


louwers:
<img width="698" alt="image" src="https://github.com/elk-zone/elk/assets/649392/d1730895-084a-42ed-98ac-0ae0e03d1109">

PR #2897, opened 2024-06-29
fix(ui): notification like avatar not shrink

cxroop:
Fixes Issue #2893 
After:
elk


ayoayco:
Thanks for the PR. Fos and refactor looks clean. @shuuji3

PR #2903, opened 2024-07-05
fix(ui): change apple app status bar style

### LINKED ISSUE #2384: iPadOS Safari displays status bar with light text in light theme
IMG_0088
IMG_0087

I'm using iPadOS 16.6 and using Safari browser. And I added elk.zone site to Home Screen.

As you can see from the first image, the status bar is white and the background is also white color. But in dark theme the text and background display correctly.

Another example is this image. 
IMG_0089

But it cannot change status bar color in dark theme

IMG_0090

I don't know if it's possible to fix this for Safari.

DasLixou:
image

On my iPhone, with a PWA, the status bar is always white, which makes it unreadable in white mode.

In this PR, I try to get this to work with the selected theme.

Because I don't know how to debug it differently, I'll use the hosted PR example to test.

Closes #2384

### DIFF
--- a/plugins/color-mode.ts
+++ b/plugins/color-mode.ts
@@ -12,7 +12,7 @@ export default defineNuxtPlugin(() => {
       {
         id: 'apple-mobile-web-app-status-bar-style',
         name: 'apple-mobile-web-app-status-bar-style',
-        content: () => 'default'
+        content: () => 'default',
       }],
   })
 })
--- a/plugins/color-mode.ts
+++ b/plugins/color-mode.ts
@@ -13,6 +13,7 @@ export default defineNuxtPlugin(() => {
         id: 'apple-mobile-web-app-status-bar-style',
         name: 'apple-mobile-web-app-status-bar-style',
         content: () => 'default',
-      }],
+      },
+    ],
   })
 })
--- a/plugins/color-mode.ts
+++ b/plugins/color-mode.ts
@@ -3,17 +3,10 @@ import { THEME_COLORS } from '~/constants'
 export default defineNuxtPlugin(() => {
   const colorMode = useColorMode()
   useHead({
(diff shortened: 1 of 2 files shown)


plugins/color-mode.ts
```diff
@@ -3,10 +3,16 @@ import { THEME_COLORS } from '~/constants'
...
+    meta: [
+      {
+        id: 'theme-color',
+        name: 'theme-color',
+        content: () => colorMode.value === 'dark' ? THEME_COLORS.themeDark : THEME_COLORS.themeLight,
+      },
+      {
+        id: 'apple-mobile-web-app-status-bar-style',
+        name: 'apple-mobile-web-app-status-bar-style',
+        content: () => 'default'
```
userquin, line 15: add a trailing comma here

userquin:
uhmm, we should just switch the flag here: https://github.com/elk-zone/elk/blob/main/nuxt.config.ts#L264

@DasLixou can you send a new PR changing to default previos line?

DasLixou:
> uhmm, we should just switch the flag here: https://github.com/elk-zone/elk/blob/main/nuxt.config.ts#L264
> 
> @DasLixou can you send a new PR changing to default previous line?

Oh interesting it was already there :0 yeah, can do that, but I'll do it in here to track everything

DasLixou:
I was wondering why an explicit default would change something but with the other line being overwritten that makes sense

userquin:
switching to dark theme  should also work, can you confirm that?

DasLixou:
> switching to dark theme should also work, can you confirm that?

I tested that with my old changes yes, and the new one set the same property, so it should work

DasLixou:
yep, works.

userquin (approved):
thx ❤️

PR #2913, opened 2024-08-08
docs(contributing): update node and pnpm version

Christopher-Hayes:
The contributing docs reference Node and PNPM versions that are no longer used by Elk:

- Node v16 should be Node v20
- PNPM v7 should be PNPM v9

Those versions were last updated on Elk by:

- Node v20: #2501
- PNPM v9: #2908


userquin (approved):

shuuji3:
Thanks!😃

Issue #2914, opened 2024-08-08
[UI] Profile name cut off by roles on `AccountInfo.vue`


Christopher-Hayes:
If you have a role, it truncates your profile name in the bottom-left `AccountInfo.vue`.

## Screenshots

![Profile snippet from Elk.zone showing a user with the display name "Chris ..." and a "Supporter" role badge. The username is truncated as "@chris_hayes@foss...".](https://github.com/user-attachments/assets/fac32d50-bc91-421e-92ac-a17bc68662d2)

PR #2915, opened 2024-08-08
fix(ui): profile name gets cut off by roles

Christopher-Hayes:
## Description

Fixes the UI issue with the profile name getting cut off when the user has roles.

On the bottom-left profile card, the roles UI was moved to be below the username.

It _could_ be in-between the name and username, but because this is a list of roles, for some users that list would look better as the last list item.

### Minor tweaks

For this taller profile info card, the avatar was centered. `AccountHoverCard.vue` originally was not using the `hoverCard` prop on `AccountInfo.vue`, this was added since the hover card has the space to show roles inline with the profile name.

## Screenshots

Before:

image

After:

image

The hover card remains untouched since it has the extra space:

image

Mobile:

image

Multiple users:

image

## Misc

- [X] Linted
- [X] Passed tests

fixes: #2914

### DIFF
--- a/components/account/AccountHeader.vue
+++ b/components/account/AccountHeader.vue
@@ -189,7 +189,6 @@ async function copyAccountName() {
         <div flex="~ col gap1" pt2>
           <div flex gap2 items-center flex-wrap>
             <AccountDisplayName :account="account" font-bold sm:text-2xl text-xl />
-            <AccountRolesIndicator v-if="account.roles?.length" :account="account" />
             <AccountLockIndicator v-if="account.locked" show-label />
             <AccountBotIndicator v-if="account.bot" show-label />
           </div>
@@ -201,6 +200,9 @@ async function copyAccountName() {
                 <span sr-only>{{ $t('account.copy_account_name') }}</span>
               </button>
             </CommonTooltip>
+          </div>
+          <div self-start mt-1>
+            <AccountRolesIndicator v-if="account.roles?.length" :account="account" />
           </div>
         </div>
       </div>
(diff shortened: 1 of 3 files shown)


Christopher-Hayes:
**Update** - I just realized the AccountInfo card only shows 1 role. So, my comment about this being a "list of roles" doesn't apply.

ayoayco:
If I remember correctly, you'd only see roles if the profile is viewed in the oroginal server, and maybe the preview deployment is in a different server than your account.

ayoayco:
I think the idea was for the profile to be consistent in different views. I'm not against moving the role chip down to prevent the name getting cut but maybe we should do that everywhere if we do it.

Christopher-Hayes:
Sure. So, if Elk went with this UI update, profile styling across all locations in the app would be:

```
<Display Name> <Lock Indicator> <Bot Indicator>
<Account Handle>
<Account Roles>
```

I pushed a commit to remove the code that shows the profile roles inline on the hover card. I also updated the profile page to make name + username + role order consistent with everywhere else that the profile is shown.

**Profile hover card:**
image

**Profile page:**
image

If we use this order, it gives the profile page room to show a full list of roles without the username getting pushed down. This is also the order that Mastodon uses.

image

Christopher-Hayes:
I'm not specifically against putting the roles inline, it does vibe with Elk try to make the UI feel cleaner and more space efficient.

For a UI comparison, GitHub has a pretty clean hover card for profiles, with display name and pronouns inline.

image

At the end of the day, I'm just interested in fixing the profile UI at the bottom-left. So, another alternative could be to just not show the roles at the bottom-left as part of the profile picker. It's not a public-facing part of the UI (compared to the profile page), and it's meant to tell the user what account their logged into, rather than educate the user about their account. It could be argued that omitted roles here would not confuse users too much, and on the consistency side, this would allow Elk to keep the inline roles everywhere else.

antfu (approved):
I am good with the change. We could figure out a better design later (if ever need), consider this is more like a fix, I think we could have it and move on first.

PR #2930, opened 2024-08-24
refactor: Remove redundant imports

itoao:
This PR removes redundant import statements. This will clean up the code and make maintenance easier.

This change is based on the previously merged PR, "[refactor: Enable auto-import for composables.](https://github.com/elk-zone/elk/pull/2929)" The redundant imports can be removed after auto-import has been enabled.


itoao:
Originally, I intended to split the changes into two separate PRs. However, this caused CI errors, so I will consolidate them into a single [PR](https://github.com/elk-zone/elk/pull/2929).

PR #2941, opened 2024-09-05
fix: fix invalid permissions-policy value for fullscreen

### LINKED ISSUE #2935: Elk white-outs in Chrome 129
On current Chrome beta (129), Elk is completely broken with an error `Error with Permissions-Policy header: Parse of permissions policy failed because of errors reported by structured header parser.`.

This is likely caused by a permission-policy based enforcement (I'm not familiar with details).

shuuji3:
fix #2935

The current Elk responds to the request with an invalid value in the `Permissions-Policy` HTTP Header.

This PR replaced the value with `*`, which is the correct value to allow the fullscreen feature for all domains:

```diff
-Permissions-Policy camera=(), display-capture=(), fullscreen='self' https: http:, geolocation=(), microphone=()
+Permissions-Policy camera=(), display-capture=(), fullscreen=*, geolocation=(), microphone=()
```

References:
- [Permissions-Policy - HTTP | MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Permissions-Policy#examples)
- [webappsec-permissions-policy/permissions-policy-explainer.md at main · w3c/webappsec-permissions-policy](https://github.com/w3c/webappsec-permissions-policy/blob/main/permissions-policy-explainer.md)
- [Controlling browser features with Permissions Policy  |  Privacy & Security  |  Chrome for Developers](https://developer.chrome.com/docs/privacy-security/permissions-policy)

### DIFF
--- a/nuxt.config.ts
+++ b/nuxt.config.ts
@@ -299,7 +299,7 @@ export default defineNuxtConfig({
         'upgrade-insecure-requests': true,
       },
       permissionsPolicy: {
-        fullscreen: ['\'self\'', 'https:', 'http:'],
+        fullscreen: '*',
       },
     },
     rateLimiter: false,


shuuji3:
It looks like Netlify preview doesn't return `Permissions-Policy` header:
image

I checked the response header value by building the app and running `pnpm run start:https` locally instead.

image

shuuji3:
Thanks, I can also confirm that only Elk logo is shown in the same version of Google Chrome Beta.

Screenshot from 2024-09-05 22-35-55

This might mean that there is some issue other than the Permissions-Policy header, and that's not compatible with Google Chrome v129.

Jawvig:
This PR still looks like a valid fix for the header value, even though it doesn't fix the Chrom(ium) 129 white out issue.

shuuji3:
Yes, there seems to be some unknown issue causing the white-out issue only on Chromium v129 other than this.

danielroe (approved):

PR #2945, opened 2024-09-06
perf: reduce timeline limit on slow connection

Shinigami92:
This reduces the limit for timeline post to 10 when used on slow connection

with this change:
image

without this change:
image


shuuji3 (approved):
Nice improvement👍🏻 Confirmed the post-number change by switching to 3G network on DevTools.

PR #2953, opened 2024-09-11
fix(ui): generate only one thread composer in `PublishWidgetList` and provide it to each widget

userquin:
This PR builds the composer in the list sfc and pass it to each widget: the widget will use the provided or build a new one composer depending on the new optional `thread-composer` prop.

(fixes the vue setup warning from this https://github.com/elk-zone/elk/pull/2946#pullrequestreview-2297350986)

@patak-dev @danielroe @shuuji3 I didn't did anything about the composer logic, maybe you can help me to test this PR properly

imagen

### DIFF
--- a/components/publish/PublishWidgetList.vue
+++ b/components/publish/PublishWidgetList.vue
@@ -20,9 +20,8 @@ const {
   dialogLabelledBy?: string
 }>()
 
-const threadItems = computed(() =>
-  useThreadComposer(draftKey, initial).threadItems.value,
-)
+const threadComposer = useThreadComposer(draftKey, initial)
+const threadItems = computed(() => threadComposer.threadItems.value)
 
 onDeactivated(() => {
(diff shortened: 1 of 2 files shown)


components/publish/PublishWidgetList.vue
```diff
@@ -20,9 +20,8 @@ const {
   dialogLabelledBy?: string
 }>()
 
-const threadItems = computed(() =>
-  useThreadComposer(draftKey, initial).threadItems.value,
-)
+const threadComposer = useThreadComposer(draftKey, initial)
+const threadItems = computed(() => threadComposer.threadItems.value)
```
userquin, line 24: I still don't understand why this should be a computed, Ig uess it can be used with `threadComposer.threadItems.value` in the template

danielroe (approved):

PR #2975, opened 2024-09-23
fix: change page reload and account switch logic

### LINKED ISSUE #2972: Losing my place when clicking the tab

userquin:
This PR includes:
- move page reload logic to the new plugin from the users module
- we need to reload the page (any client openened: tabs or browser instances): there is no way to make it working with Chrome 129 via `visibilty` listeners.
- ~~fix masto plugin, rn calling masto in the server~~
- fix hashtags and lists pages: add the folder and the index page using nuxt page on root
- removed `masto.ts` plugin and moved the logic to `0.setup-users.ts`  plugin: client logic will activate the watcher once user signed (✔️ we should use `finally` instead `then`: DONE)

**NOTE**: almost impossible to test 2 previous fixes from main, any page refresh and the current token broken (masto error)

<details>
<summary><i>Error building elk.zone (nitropack patched locally)</i></summary>

imagen
</details>

EDIT: no idea what's doing Google when releasing v129

closes #2972


composables/users.ts
```diff
@@ -19,7 +19,7 @@ const mock = process.mock
 
 const users: Ref<UserLogin[]> | RemovableRef<UserLogin[]> = import.meta.server ? ref<UserLogin[]>([]) : ref<UserLogin[]>([]) as RemovableRef<UserLogin[]>
 const nodes = useLocalStorage<Record<string, any>>(STORAGE_KEY_NODES, {}, { deep: true })
-const currentUserHandle = useLocalStorage<string>(STORAGE_KEY_CURRENT_USER_HANDLE, mock ? mock.user.account.id : '')
+export const currentUserHandle = useLocalStorage<string>(STORAGE_KEY_CURRENT_USER_HANDLE, mock ? mock.user.account.id : '')
```
userquin, line 22: maybe we don't need to export this

userquin:
Looks like we have some problem with sign-in logic (I added some await, FF and Chrome ara not able to load home page)

userquin:
I need to review the `useAsyncIDBKeyval` composable...

userquin:
When switching accounts on some `[[account]]` page:

imagen

composables/idb/index.ts
```diff
@@ -1,72 +1,75 @@
...
 
-const isIDBSupported = !process.test && typeof indexedDB !== 'undefined'
+export interface UseAsyncIDBKeyvalReturn<T> {
+  set: (value: T) => Promise<void>
+  readIDB: () => Promise<T | undefined>
+}
 
 export async function useAsyncIDBKeyval<T>(
-  key: IDBValidKey,
+  key: string,
```
userquin, line 11: revert this

userquin:
Cannot sign out in PR preview.

plugins/0.setup-users.ts
```diff
@@ -24,11 +25,87 @@ export default defineNuxtPlugin({
...
+    const user = (typeof query.server === 'string' && typeof query.token === 'string')
+      ? {
+          server: query.server,
+          token: query.token,
+          vapidKey: typeof query.vapid_key === 'string' ? query.vapid_key : undefined,
+        }
+      : (currentUser.value || { server: publicServer.value })
+
+    if (import.meta.client) {
+      loginTo(masto, user).then(callback)
```
userquin, line 102: use `finally` here

components/nav/NavSide.vue
```diff
@@ -8,6 +8,27 @@ const { notifications } = useNotifications()
...
+    return '/notifications'
+  }
+
+  return `/notifications/${lastAccessedNotificationRoute}`
+})
+const exploreLink = computed(() => {
+  const hydrated = isHydrated.value
+  const server = currentServer.value
+  const lastRoute = lastAccessedExploreRoute.value
+  if (!hydrated || !server || !lastRoute) {
```
userquin, line 26: upps, this is wrong, remove `!server`

components/nav/NavSide.vue
```diff
@@ -8,6 +8,27 @@ const { notifications } = useNotifications()
...
+
+const notificationsLink = computed(() => {
+  const hydrated = isHydrated.value
+  const user = currentUser.value
+  const lastRoute = lastAccessedNotificationRoute.value
+  if (!hydrated || !user || !lastRoute) {
+    return '/notifications'
+  }
+
+  return `/notifications/${lastAccessedNotificationRoute}`
```
userquin, line 20: shouldn't eslint fail here (computed without `.value`)?

userquin: `RemovableRef` is a type from VueUse using type instead interface (`type RemovableRef<T> = Omit<Ref, 'value'> & {...}`), maybe some check not being resolved: https://github.com/vuejs/eslint-plugin-vue/blob/2dc606ca589ff5cdcd197c3c64ace7e575a6347d/lib/utils/index.js#L1130

shuuji3 (approved):
Tested on Firefox 130.0.1 (64-bit) on Linux.

Although I didn't fully trace the code change, I confirmed that
- the reported reload bug was fixed
- list and hashtag pages are working as previously
- last selected notification and explorer tabs are remembered correctly
- there's reloading happening when switching accounts but I think that should not affect usability so significantly and acceptable for now

danielroe (approved):

mattcen:
Thanks for the quick resolution folks! Really appreciate it!

Issue #2974, opened 2024-09-23
dropdown menu is not keyboard accessible


jyn514:
posts in elk.zone have a dropdown menu that looks like this:

![image](https://github.com/user-attachments/assets/9872fdbc-e5b1-40a3-bb1c-65cf8b64a6e3)

3/4 of the links here cannot be selected with the keyboard (either with Tab, or with an extension like vimium). would it be possible to make them selectable? maybe with tabindex=0 or something like that?

PR #2976, opened 2024-09-24
feat(a11y): make menu buttons in 'More' dropdown selectable with the keyboard

jyn514:
previously, these were pure divs, without a `<button>` element, and could not be selected with `Tab`, rendering them keyboard-inaccessible.

see also https://github.com/elk-zone/elk/pull/2767 (cc @userquin). that uses a slightly different approach; it adds a disabled anchor instead of using buttons. i think this html is more semantically accurate, but i don't mind changing it if you so desire.

fixes https://github.com/elk-zone/elk/issues/2974


jyn514:
> 8:31:15 PM: Failed during stage 'checking build content for changes': Canceled build due to no content change

i am not sure why this thinks nothing changed; i tested it locally with `pnpm dev` and it's definitely selectable now where it wasn't before.

userquin:
can you update your main and merge changes in this PR?

userquin:
> i am not sure why this thinks nothing changed; i tested it locally with `pnpm dev` and it's definitely selectable now where it wasn't before.

What do you mean? The buttons are there:

<details>
<summary><i>Buttons on StatusActionsMore</i></summary>

imagen
</details>

userquin:
We should do the same also in `components/account/AccountMoreButton.vue`

jyn514:
> can you update your main and merge changes in this PR?
> do the same also in `components/account/AccountMoreButton.vue`

done :)

> What do you mean? The buttons are there (also in the PR preview):

i mean the comment the github bot posted on this PR, that says the build was cancelled. maybe i'm misinterpreting it somehow, clearly the preview is showing up.

> add type="button" when is is button in components/common/dropdown/DropdownItem.vue: iirc type by default is submit

can you say more? i'm looking at the rendered button and i don't see any `submit` attribute. does it affect the event type or something like that?

i glanced at DropdownItem but i don't see anything related to `submit`; the closest is `:is="is"` in a component. maybe i'm missing something, i've never used react before :sweat_smile: 

image

jyn514:
> i mean the comment the github bot posted on this PR, that says the build was cancelled. maybe i'm misinterpreting it somehow, clearly the preview is showing up.

ohhhh that's for the *docs*, not for the live preview of the site. that makes more sense.

userquin:
`const type = computed(() => is === 'button' ? 'button' : null)`

Then add `:type="type"` below `:is="is"` in the template

userquin:
We should use props destr. instead with defaults (enabled by default in vue 3.5)

jyn514:
> it broke at first (apparently NuxtJS can't handle `null`?) so i just had it fallback to `submit`

oh wait apparently this started working again?? the only thing i changed was the order `:type` appears in the element ...

anyway, i've reverted this change, so now `type=""` only appears for button elements, not other dropdown items.T

userquin:
Use undefined instesd null

shuuji3 (approved):
Sorry for taking a while to review 🙏🏻 This is a great accessibility improvement! Thank you so much.😀 

screenshot of context menu and accessibility tree view on devtool. it shows all menu items can be focusable and button elements

PR #3001, opened 2024-10-18
Correct some German translations, also make them more consistent

palant:
I’ve noticed that some German translations under https://elk.zone/settings/notifications/push-notifications are incorrect: these have been translated as an action to be performed by you rather than other people’s action that you are notified about. A similar mistake was made for notification filters, one translation also inconsistently using a plural form instead of singular.

There is also the issue of inconsistent translation of “boost”: sometimes it was translated as “reblog” which isn’t a Mastodon term, sometimes as “boost” which just isn’t German. I tried to make the translations here use “share” terminology consistently.

### DIFF
--- a/locales/de-DE.json
+++ b/locales/de-DE.json
@@ -230,7 +230,7 @@
     "dialog_header": "Tastaturkürzel",
     "groups": {
       "actions": {
-        "boost": "Boosten",
+        "boost": "Teilen",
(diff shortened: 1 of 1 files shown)


Shinigami92 (approved):

shuuji3:
Thanks! 😃

Issue #2889, opened 2024-06-19
Display of mentions and notifications mangled sometimes


audunmb:
This only occurs sometimes, and a reload of the page usually fixes the issue. What happens is that my notifications are mangled so that some posts are displayed as being part of just one post. 

Like this: 
![elk-mangled notifications](https://github.com/elk-zone/elk/assets/32357781/139b1a89-af39-405b-a654-3294523010fe)

The picture shows a Favorite notification with the text of three different posts displayed as one post.

johnsonm:
Looks to me a lot like #2216 

That had stopped happening to me some time ago, and honestly I forgot that it was still open. Looks like it might have been fixed at some point without reference to that issue and now started happening again?

audunmb:
Yes this is a duplicate, but didn't find the other issue since I searched for notification issues.

johnsonm:
Yup, I was just linking them together for more information for developers, not complaining about duplication. Issue search is hard. :)

Baralheia:
I'm also seeing this issue, and have for quite some time... at least the last several months. Like the OP, refreshing usually fixes it. I'm using Firefox 128 on Linux Mint 22, but this issue persists across browser versions and across OSs too.

![image](https://github.com/user-attachments/assets/697487cd-7b7f-47a1-abad-c574b41b015a)

piebob:
this is a weird example of the jumbling issue--the mentioned usernames from one notification item were replaced by the mentions from another notification item: 
![image](https://github.com/user-attachments/assets/252e5c08-b944-4328-a632-49d54d72b5cc)

the second notification item has the correct content and mentioned users. the first one has pulled tags from a different post i made. super weird.

PR #3014, opened 2024-10-22
fix(ui): Avoid duplicate IDs for grouped notifications

### LINKED ISSUE #2216: Content from multiple posts getting mixed together
Occasionally I'm seeing content from multiple posts mixed into one. For instance, I was on my notifications page, and saw "show 1 new item". I clicked the notification text, and saw this:

image

That content is mixed together from two posts:

https://social.makerforums.info/@mcdanlj/110678716327637629

https://social.makerforums.info/@mcdanlj/110684361403957499

I haven't yet figured out a way to reproduce this. I've seen something like this on Firefox 114; the screenshot is from Firefox 115 on Linux.

Here's another set of notifications:

image

That inserts the wrong hashtag, which it got from this post:

https://social.makerforums.info/@mcdanlj/110684291181195491

palant:
Steps to reproduce are in https://github.com/elk-zone/elk/issues/2216#issuecomment-2424229983. This fixes #2216 and #2889.

The issue here is function `groupItems` issuing identical IDs for different objects. In the first update it creates a `GroupedLikeNotifications` instance with ID `grouped-0`. Then we get a second update, and it again creates a `GroupedLikeNotifications` instance with ID `grouped-0`. Since both have the same key, Vue attempts to reuse the existing node for the new object. It replaces the existing `vnode` in `StatusBody` by one that belongs to a different status, and it fails miserably updating the display. So we get a correct internal structure but a completely messed up display. The solution is using a global ID counter, so that this function never re-issues the same ID.

### DIFF
--- a/components/notification/NotificationPaginator.vue
+++ b/components/notification/NotificationPaginator.vue
@@ -15,6 +15,8 @@ const groupCapacity = Number.MAX_VALUE // No limit
 
 const includeNotificationTypes: mastodon.v1.NotificationType[] = ['update', 'mention', 'poll', 'status']
 
+let id = 0
+
 function includeNotificationsForStatusCard({ type, status }: mastodon.v1.Notification) {
   // Exclude update, mention, pool and status notifications without the status entry:
   // no makes sense to include them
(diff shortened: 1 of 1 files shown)


antfu (approved):

shuuji3:
Thank you for the detailed analysis! I can also confirm the fix with the reproduction steps. 💯

PR #3015, opened 2024-10-22
fix: Do not require Git for development

palant:
Running `pnpm nuxt prepare` currently fails for me:

```
ERROR  fatal: not a git repository (or any parent up to mount point /)
Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
```

That’s because the directory in question isn’t a Git repository.

Running from a Git repository or even having Git installed shouldn’t be a requirement. With my change errors originating from Git are ignored. The potential impact is negligible.

### DIFF
--- a/config/env.ts
+++ b/config/env.ts
@@ -32,22 +32,25 @@ export async function getGitInfo() {
   let branch
   try {
     branch = gitBranch || await git.revparse(['--abbrev-ref', 'HEAD'])
-  } catch (e) {
-    branch = "unknown"
+  }
+  catch {
+    branch = 'unknown'
   }
 
   let commit
   try {
     commit = await git.revparse(['HEAD'])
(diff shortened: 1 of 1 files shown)


shuuji3 (approved):
I think reducing the requirement of `git` would be fine and this won't affect the behavior in dev environment or any other environment with `git`.

Just curious, What kind of environment are you working in? Is that something like a dev container with only minimal dependencies?

palant:
> Just curious, What kind of environment are you working in? Is that something like a dev container with only minimal dependencies?

I’m using Mercurial. Any git repo is checked out via hg-git, so git won’t be able to do anything with it of course.

shuuji3:
Ah, I see. That's totally makes sense😄

PR #3031, opened 2024-10-30
docs: Remove duplicate instruction in ReadMe

TangentFoxy:
The docker-compose rebuilds the image, so it's a waste of time to build it first.


shuuji3 (approved):
Thanks for the report. That's correct. `docker-compose up` should build the container so no need to build before that step. I adjusted the last step to include `--build` flag so that it can ensure a clean build.

PR #3034, opened 2024-10-31
feat: add new setting to select line height options

### LINKED ISSUE #3007: Make line height configurable
Text is sometimes hard to read because lines are so close to each other.

Please make the [line height](https://fonts.google.com/knowledge/glossary/line_height_leading) configurable in settings to improve [readability](https://fonts.google.com/knowledge/glossary/readability).

(I'd like to set it to 150%)

shuuji3:
resolve #3007 

This PR implements a new setting to select line height in the post body. Users can choose the narrower (1.2) or wider (1.5) line height options, in addition to the current normal option (4/3)

### Setting section for line height
Screenshot of Setting section for line height

### Narrow (1.2)
Screenshot of the post with narrow line height
### Normal (4/3 = 1.33...): current and default value
Screenshot of the post with normal line height
### Wide (1.5)
Screenshot  of the post with wide line height


styles/global.css
```diff
@@ -125,8 +125,16 @@ em-emoji-picker {
   }
 }
 
-.line-compact {
-  line-height: calc(4 / 3 * 1em);
```
shuuji3, line 129: Removed the unit `em` following the recommendation by MDN (https://developer.mozilla.org/en-US/docs/Web/CSS/line-height#values):

> In most cases, **this is the preferred way** to set line-height and avoid unexpected results due to inheritance.

Issue #3042, opened 2024-11-07
Fix documentation build


shuuji3:
The build of the documentation fails now. Probably we'll need to update some packages and adjust settings.

ref. https://app.netlify.com/sites/elk-docs/deploys/672ba29a3a7e9e0008ded650

## How to reproduce

The following commands will output the errors same as the above:

```shell
cd docs/
pnpm i
pnpm generate
```

PR #3045, opened 2024-11-08
docs: fix documentation build error

shuuji3:
fix https://github.com/elk-zone/elk/issues/3042


userquin:
can you update the docs readme file , installing dependencies should be done from root?

userquin:
upps, looks like we change the PR title at the same time

userquin (approved):

Issue #3044, opened 2024-11-08
Profile avatar image does not respect reduced motion setting


shuuji3:
When reduced motion setting is enabled, the avatar image with animation should be stopped but it has a movement now.

example: https://main.elk.zone/mastodon.social/@radiojammor@mastodon.scot

![Image](https://github.com/user-attachments/assets/53cdfc53-67d0-48bb-bb36-fd8b3d32cdb4)

AndrewCrescencio:
hi, I'll help

PR #3066, opened 2024-11-26
fix(ui): Profile avatar image motion setting #3044

AndrewCrescencio:


shuuji3 (approved):
Thanks for the contribution! I confirmed this change stopped gif animation by replacing static image like png when enabled prefers-reduced-motion setting.

AndrewCrescencio:
nice, just commenting on this, I was a little scared, I don't know if at some point it could happen that the profile doesn't have a photo or a static photo

shuuji3:
Good point.

Mastodon has ensured `avatar_static` (https://docs.joinmastodon.org/entities/Account/#avatar) exists since v1.1.2 (in 2017), so I think we don't have to worry about that case, at least for Mastodon.

But the same property might not exist on other fediverse servers so it's safer to fall back to `avatar` property when there is no `avatar_static`. It's not optimal, but I think it's better than an empty image.

shuuji3:
For example, something like this?
```diff
 const preferredMotion = usePreferredReducedMotion()
 const accountAvatarSrc = computed(() => {
-  return preferredMotion.value === 'reduce' ? account.avatarStatic : account.avatar
+  return preferredMotion.value === 'reduce' ? (account?.avatarStatic ?? account.avatar) : account.avatar
```

shuuji3:
ref. GoToSocial returns static avatar like Mastodon: Routes and Methods - GoToSocial Documentation - https://docs.gotosocial.org/en/latest/api/swagger/

Issue #2934, opened 2024-08-27
Fix errors during test script


shuuji3:
Although all tests seem to pass, a few errors during the test run look related to the WPA manifest file.

Here's the GitHub Actions log from `pnpm test:ci` command:

[chore(deps): update devdependencies · elk-zone/elk@e1a5d83](https://github.com/elk-zone/elk/actions/runs/10570912138/job/29286149498?pr=2931)

```shell
Run pnpm test:ci

> @elk-zone/elk@0.1[4](https://github.com/elk-zone/elk/actions/runs/10570912138/job/29286149498?pr=2931#step:7:5).0 test:ci /home/runner/work/elk/elk
> nr test:unit:ci


> @elk-zone/elk@0.14.0 test:unit:ci /home/runner/work/elk/elk
> stale-dep && vitest run

Using `2024-04-03` as fallback compatibility date.
[nuxt] Expected `@nuxt/schema` to be at least `3.13.0` but got `3.12.4`. This might lead to unexpected behavior. Check your package.json or refresh your lockfile.
[nuxt] Expected `@nuxt/kit` to be at least `3.13.0` but got `3.12.4`. This might lead to unexpected behavior. Check your package.json or refresh your lockfile.

 RUN  v2.0.[5](https://github.com/elk-zone/elk/actions/runs/10570912138/job/29286149498?pr=2931#step:7:6) /home/runner/work/elk/elk

stdout | tests/nuxt/html-to-text.test.ts
<Suspense> is an experimental feature and its API will likely change.

stderr | tests/nuxt/html-to-text.test.ts
[nuxt] Error fetching app manifest. FetchError: [GET] "/_nuxt/builds/meta/dcdd378e-543d-4c83-888d-c0fec58dacf2.json": 404 Cannot find any path matching /_nuxt/builds/meta/dcdd378e-543d-4c83-888d-c0fec58dacf2.json.
    at processTicksAndRejections (node:internal/process/task_queues:95:5)
    at $fetch2 (file:///home/runner/work/elk/elk/node_modules/.pnpm/ofetch@1.3.4/node_modules/ofetch/dist/shared/ofetch.3738[6](https://github.com/elk-zone/elk/actions/runs/10570912138/job/29286149498?pr=2931#step:7:7)b05.mjs:268:15)
[...]

PR #3069, opened 2024-11-29
refactor: code from profile avatar image reduced motion setting issue

AndrewCrescencio:
Hi, just refactoring the code to fix the issue: https://github.com/elk-zone/elk/issues/3044
I removed the props destructuring, because in the test log there is a warning about this, I also added the fallback to the avatar image as @shuuji3 told here https://github.com/elk-zone/elk/pull/3066#issuecomment-2505135268


components/account/AccountAvatar.vue
```diff
@@ -1,7 +1,7 @@
 <script setup lang="ts">
 import type { mastodon } from 'masto'
 
-const { account, square } = defineProps<{
+const props = defineProps<{
```
shuuji3, line 4: @AndrewCrescencio Thanks for the fix!

About the object destructuring, that warning was valid in Vue 3.4 (the previous result of `pnpm run test:ci`), but Vue 3.5+ allows us to use the same syntax again. So we can still use the original destructuring on `props`. Could you revert that destructuring since we already used it in many other places if you don't mind? You can also check by running `pnpm run test:ci` on your environment.

These links will help to understand the latest (Vue 3.5+) usage of props destructuring:
- Props | Vue.js - https://vuejs.org/guide/components/props#reactive-props-destructure
- Vue 3.5+: How to Destructure Props in 2024 (Composition API) - https://www.docs4.dev/posts/vue-3-how-to-destructure-props-in-2024

I've updated the test result with the latest commit, that doesn't contains the previous warning: https://github.com/elk-zone/elk/issues/2934

shuuji3 (changes requested):

shuuji3 (approved):
By the way, that refactoring is optional. I'm happy to merge this and refactor it later with other similar usages.

components/account/AccountAvatar.vue
AndrewCrescencio: Nice that we still can use props destructure, thanks for info

Issue #3068, opened 2024-11-28
Split Long Posts


larry77:
Mastodon prevents posts longer than 500 characters. It would be nice to have in Elk,  inbuilt or with a warning and a boolean, the capability to split a long post as in this tool https://mastodon.help/verbose/ .

ayoayco:
I think the copying Mastodon web client's behavior of showing a warning is a good start if we cannot fully implement splitting automatically into separate posts.

ayoayco:
Comparing elk & mastodon...

Currently Elk allows submitting even when post has more characters than the server limit, and shows an error message. 

![image](https://github.com/user-attachments/assets/9064e861-1060-4ee5-819c-89e8007548c7)

Mastodon behavior is disable the POST button -- we can do this first (simple start) then refine for a feature later that suggests splitting long texts

![image](https://github.com/user-attachments/assets/098bf8c8-4be1-45e2-b8bb-98e855a16794)

ayoayco:
addressed the publish button with #3085

larry77:
Thanks, that is already something. However, having the possibility of splitting and posting longer text would be really a pleasant feature.

ayoayco:
Let’s keep this open for the feature request

PR #3085, opened 2024-12-13
feat(ui): disable publish button

ayoayco:
This will add `disabled` attribute to actually disable the publish button when condition is met (e.g., post exceeds server char. limit). Previously we only had `aria-disabled`.

discovered with issue: #3068


shuuji3 (approved):
LGTM!

I never thought this button was clickable😄

PR #3090, opened 2024-12-15
fix: more menu is not opening on click

Cl0v1s:
The BottomNav more menu is not working on mobile view. Affecting the defineModel return value to a const allows to solve this issue

### DIFF
--- a/components/nav/button/MoreMenu.vue
+++ b/components/nav/button/MoreMenu.vue
@@ -1,10 +1,10 @@
 <script setup lang="ts">
-defineModel<boolean>()
+const model = defineModel<boolean>()
 </script>
 
 <template>
   <NavBottomMoreMenu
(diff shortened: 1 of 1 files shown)


shuuji3 (approved):
Thanks! I think the code change looks fine.

For reference, could you tell us more details on your environment such as OS, browser version etc.? I couldn't reproduce the same issue on my mobile browsers (Google Chrome and Firefox on Android 15).

Cl0v1s:
Thanks for your feedback ! 

I encountered that issue on both firefox (desktop, mobile dev tool) and Chrome 131.0.6778.135 on Android 12 :)


Le 16 décembre 2024 16:44:35 GMT+01:00, TAKAHASHI Shuuji ***@***.***> a écrit :

PR #3091, opened 2024-12-15
fix: MenuBottom settings are always empty

Cl0v1s:
When setting BottomNav menu entries, the space showing which NavButtons are currently shown is always blank. It is not showing the current saved selection neither the base defaults.


shuuji3 (approved):
Setting the empty list was the originally intended behavior, but keeping the current setting is more useful especially when users just want to tweak a few arrangements. Thanks!

PR #2938, opened 2024-09-01
fix(i18n): Fix redundant Japanese translation

m1sk9:
- 4d83a2061980f935caf859a1edc71e7e8f4cc499: Some redundancies in the Japanese translation have been corrected.

PR #3105, opened 2024-12-24
feat(i18n): Unify expression in Japanese translation

m1sk9:
Fixed expressions that could not be fixed in #2938


shuuji3 (approved):
Good catch! Thank you 🙏🏻

PR #3111, opened 2024-12-29
feat: update poll

Cl0v1s:
Add an "update poll" option on polls to retrieve the latest results 

update poll option (french)

Upodate poll option on hover (french)

The refresh icon is animated during loading

### DIFF
--- a/components/status/StatusPoll.vue
+++ b/components/status/StatusPoll.vue
@@ -87,19 +87,23 @@ const votersCount = computed(() => poll.votersCount ?? poll.votesCount ?? 0)
         </div>
       </div>
     </template>
-    <div text-sm text-secondary flex justify-between items-center>
-      <div flex="~ inline" gap-x-1>
-        <CommonLocalizedNumber
-          keypath="status.poll.count"
-          :count="poll.votesCount"
-        />
+    <div text-sm text-secondary flex justify-between items-center gap-3>
+      <div flex gap-x-1 flex-wrap>
+        <div inline-block>
+          <CommonLocalizedNumber
+            keypath="status.poll.count"
+            :count="poll.votesCount"
+          />
+        </div>
         &middot;
-        <CommonTooltip v-if="poll.expiresAt" :content="expiredTimeFormatted" class="inline-block" placement="right">
-          <time :datetime="poll.expiresAt!">{{ $t(poll.expired ? 'status.poll.finished' : 'status.poll.ends', [expiredTimeAgo]) }}</time>
-        </CommonTooltip>
+        <div inline-block>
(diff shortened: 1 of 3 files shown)


shuuji3 (commented):
Thanks, this is a nice addition.

I checked the change with this preview post: https://deploy-preview-3111--elk-zone.netlify.app/universeodon.com/@elkdev/112190894000305452

One comment: once the poll is finalized, it won't be updated anymore, so should we hide the update button after that?

Cl0v1s:
That could be great indeed. Do you think it would be better to disabled the button or to hide it completely ?

shuuji3:
I think hiding it would be better because
1. When the poll has already finished, users won't see the update button so we can avoid making them think it can be updated and they might think there's some issue they cannot click it until they realize the poll was finished at that time.
2. When the poll is still ongoing, users can click the update button to see the latest numbers until it finishes and can know it's finished clearly.

Cl0v1s:
Will do !

Cl0v1s:
On second though, there is still the following case: the poll is indeed closed but the local instance did not refetched the results yet (I saw that happening from time to time). Is this a case we should consider ?

shuuji3:
That sounds a bit tricky. So in my understanding, there are three phases about the poll state and returned poll data:
1. The poll is still ongoing on remote server, and the local server returns the unfinished poll data.
2. The poll was actually finished on remote server, but the local server returns the outdated poll data.
3. The poll was finished and the local server returns the latest finished poll data.

Since Elk can only send request via masto.js, we can only see the local poll data and we can use either `expired` (boolean) and `expiredAt` (time to finish) props in `Poll` (ref. https://neet.github.io/masto.js/interfaces/mastodon.v1.Poll.html).

So if we compared `expiredAt` with the current time and hide the update button, we shows the wrong state for case 2. That means we need to use `expired` and hide the update button only if its value is `true` (after local server synced the latest poll data).

Do you think my assumption is correct?

Cl0v1s:
Oh indeed ! That seems perfectly correct, I will change the code according to that last answer :)

Cl0v1s:
Done !

components/status/StatusPoll.vue
```diff
@@ -39,6 +40,21 @@ async function vote(e: Event) {
   await client.value.v1.polls.$select(poll.id).votes.create({ choices })
 }
 
+async function refresh() {
+  loading.value = true
```
shuuji3, line 44: Thanks for the updates!

Before merging, I noticed that "Update poll" button can be clicked multiple times in a short period.

The below change will be able to prevent a new request until Elk receives the response from the server.

```suggestion
  if (loading.value) {
    return
  }

  loading.value = true
```

shuuji3 (changes requested):

shuuji3 (approved):
Thanks, LGTM!

PR #3120, opened 2025-01-01
fix(a11y): show bottom nav button labels

### LINKED ISSUE #3119: Accessibility: add label under icons on mobile navbar
Accessibility best practices states that mobile navbar items should have a label.

We should at least offer an option to display them. 

Or, even better, display them by default with an option to hide them

shuuji3:
fix #3119

This PR includes
- showing labels below the bottom nav button icon
- a new option to hide them on the interface setting page (just below the nav button customization section)

**With labels**

Screenshot from 2025-01-01 23-46-09

**Without labels (current state)**

Screenshot from 2025-01-01 23-46-19

About the default setting, this change may surprise the existing users a bit if they like the current simpler interface. On the other hand, I also understand showing labels by default can provide better accessibility.


Dehelssey:
On deploy preview the checkbox for hiding labels does nothing 👀 (i run /e/OS (Android) with Vivaldi)

shuuji3:
@Dehelssey Unlike other settings, it seems like reloading is required to change to be effective.

shuuji3:
OK, I figured out how to make them reactive. Now checking should be effective immediately.

userquin:
uhmmm, why is `getPreferences` not reactive?

userquin:
I guess it will be better to move the check in the interface page to the top, between **Bottom Nvigation** and the `Choose...` paragraph:

imagen

userquin:
I guess you should also change the button layout in the interface page.

shuuji3:
OK, I'll add labels in that preview too, and move the option below **Bottom Navigation** header.

> uhmmm, why is `getPreferences` not reactive?

I'm not sure yet 😅

shuuji3:
- moved the checkbox
- show labels in the preview area
Screenshot of setting page without labels in preview
Screenshot of setting page with labels in preview

Also, fixed the reactivity issue with `compute()` and made it composable. (Thanks @userquin for the help!)

Other than that, I feel like we should add a small space between icon and label. I'll update that tomorrow.

userquin:
Don't merge yet, the preview buttons not being translated, I'm going to fix it.

userquin:
https://m3.material.io/components/navigation-bar/guidelines

imagen

Dehelssey:
@userquin indeed, ellipsis is not advised. If that label can't be replaced by something shorter I'm not sure how to deal with this. French isn't as effective as english but we don't deal with that long words anyway. In my opinion, two lines label remains the best way to deal with it, but icons should be remain aligned in such a case.

userquin:
> @userquin indeed, ellipsis is not advised. If that label can't be replaced by something shorter I'm not sure how to deal with this. French isn't as effective as english but we don't deal with that long words anyway. In my opinion, two lines label remains the best way to deal with it, but icons should be remain aligned in such a case.

We cannot do that, the bottom nav should have ~67px height, we can ask if the lang has some abbr. to use in that case otherwise there is no solution for this.

shuuji3:
It seems that we need to find a short string for this new button label in some languages. For English, I think we can shorten "More menu" to "Menu" or something.

In various apps, developers themselves somehow come up with good short labels and if no localization is available, they fall back to English.

So how about adding new localization keys like `nav_short_label` next to `nav` keys here?
https://github.com/elk-zone/elk/blob/411fcc8/locales/en.json#L305

For example:

```diff
 "nav": {
     ....
     "conversations": "Conversations",
     ...
     "more_menu": "More menu",
     ...
+"nav_short_label": {
+    ....
+    "conversations": "Conversations",
+    ...
+    "more_menu": "Menu",
```

We can just copy the existing `nav` strings for languages that already have short labels and fall back to English while waiting for the update of other languages.

## Nav bars examples

I checked nav bars in various Android apps (in English, French, and German).

### Slack
Screenshot_20250102-140357
Screenshot_20250102-140328
Screenshot_20250102-140305

### Play Store

Screenshot_20250102-140728
Screenshot_20250102-135909

### Contacts
Screenshot_20250102-135645
Screenshot_20250102-135641
Screenshot_20250102-140653

### Others
Screenshot_20250102-135534
Screenshot_20250102-140246
Screenshot_20250102-135756


### Google Fit and GitHub
They failed showing a good short label in German 😅 
Screenshot_20250102-141111
Screenshot_20250102-141144

shuuji3:
It seems that we need to find a short string for this new button label in some languages. For English, I think we can shorten "More menu" to "More" or something.

In various apps, developers themselves somehow come up with good short labels and if no localization is available, they fall back to English.

So how about adding new localization keys like `nav_short_label` next to `nav` keys here?
https://github.com/elk-zone/elk/blob/411fcc8/locales/en.json#L305

For example:

```diff
 "nav": {
     ....
     "conversations": "Conversations",
     ...
     "more_menu": "More menu",
     ...
+"nav_short_label": {
+    ....
+    "conversations": "Conversations",
+    ...
+    "more_menu": "More",
```

We can just copy the existing `nav` strings for languages that already have short labels and fall back to English while waiting for the update of other languages. What do you think?

## Nav bars examples

I checked nav bars in various Android apps (in English, French, and German).

### Slack
Screenshot_20250102-140357
Screenshot_20250102-140328
Screenshot_20250102-140305

### Play Store

Screenshot_20250102-140728
Screenshot_20250102-135909

### Contacts

Screenshot_20250102-135641
Screenshot_20250102-140653

### Others
Screenshot_20250102-135534
Screenshot_20250102-140246
Screenshot_20250102-135756


### Google Fit and GitHub
They failed showing a good short label in German 😅 
Screenshot_20250102-141111
399644292-54b5c1a2-adba-4ad9-afb9-72351f383e70

shuuji3:
^ A bit of style refinement (only for preview area)

Screenshot of nav button settings page

Screenshot of nav button settings page after adjusting spaces

shuuji3:
Showing ellipsis is suboptimal but maybe we need to use ellipsis for now to avoid overflow or two lines of label. For some languages having ellipsis labels, users already have choices either reducing to four buttons or hiding labels when they don't want to see ellipsis. Later, we can encourage translators to shorten those labels to avoid ellipsis as possible.

I'd like to keep the maximum number to five since it's better to avoid reducing functionality just for this label issue and the Material guide does not recommend more than 5 buttons[^1].

(Aside from the guideline, I don't know the current state but X app may not be reviewed by accessibility expert team so it may not be a good example in terms of accessibility[^2])

[^1]: [Navigation bar – Material Design 3](https://m3.material.io/components/navigation-bar/guidelines#8d570c9d-3822-4011-b974-1c7f00e02a5a)
[^2]: [Elon Musk just axed key Twitter teams like human rights, accessibility, AI ethics and curation | TechCrunch](https://techcrunch.com/2022/11/04/elon-musk-twitter-layoffs/)

shuuji3:
I'm trying to implement the ellipsis but the current CSS does not work as expected yet. But the two-line label issue was fixed.

screenshot of nav bottom labels with narrow window width, labels are overflowing without ellipsis

I referred this method: Truncate String With Ellipsis | CSS-Tricks - https://css-tricks.com/snippets/css/truncate-string-with-ellipsis/

ayo-run:
Hi @shuuji3 will you be able to resolve the conflicts? I want to check/test this next :) Thank you!

shuuji3:
@ayo-run Just merged the latest main branch (there was 785 changed files)! The screenshot is the mobile viewport simulation for iPhone SE (375x667).

screenshot of mobile view; no text overwrap

ayo-run (changes requested):
Tested on `Galaxy Z Fold 5` -- which is the thinnest screen / dimension on chrome dev tools -- and choosing navigation options with long labels. I think it is okay to completely hide the label if the screen is not wide enough, but would it be possible to hide all labels if this is the case? It looks inconsistent if only one is hidden. See screenshot:

Screenshot from 2026-06-16 22-04-01

PR #3123, opened 2025-01-02
fix(rtl): replace ellipsis span anchor content with bdi

### LINKED ISSUE #3122: Elk places ellipses on the wrong side of a truncated URL in RTL posts
[Here's](https://mastodon.social/@eff/113658239670063298) a post from the EFF in in Arabic. If you view the post in Mastodon's UI, the post's embedded URL is truncated with the ellipses appearing on the right side of the URL.

- Semaphore displays the ellipses on the right as well.
- Tusky actually mangles this post entirely...

Viewing this post in Elk, the ellipses are on the left side, giving the mistaken appearance that the URL is right-side complete and the truncation is occurring on the left side incorrectly.

I'm not certain what the *right* way to do this is, but this is at least inconsistent with other apps and conveys incorrect information to a user about what side of the URL is clipped.

For the record I'm viewing this through a Gotosocial v0.17.1 instance but it looks correct in the database; the ellipses class is being applied to the entire visible URI segment; I think Elk's CSS is just not correctly accounting for the mixed text ordering.

userquin:
closes #3122

### DIFF
--- a/composables/content-render.ts
+++ b/composables/content-render.ts
@@ -60,16 +60,40 @@ export function nodeToVNode(node: Node): VNode | string | null {
   }
 
   if ('children' in node) {
-    if (node.name === 'a' && (node.attributes.href?.startsWith('/') || node.attributes.href?.startsWith('.'))) {
-      node.attributes.to = node.attributes.href
+    if (node.name === 'a') {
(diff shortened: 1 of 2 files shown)


shuuji3 (approved):
LGTM!

image

image

PR #3124, opened 2025-01-03
chore(utils): remove unnecessary `await`

iiio2:
I think here `await` is unnecessary though it will work. But in the previous function `deleteApp` the loop is done without `await`. So I think for consistency, we may remove `await` here.


shuuji3 (approved):
Yes, `keys` has already been resolved and is typed `string[]` here, so we don't have to wait. Thanks!

iiio2:
Thanks @shuuji3 .

PR #3133, opened 2025-01-13
fix: disable publish when be bypassed by Ctrl + Enter

### LINKED ISSUE #3107: Disabled publish button can be bypassed by Ctrl + Enter
## Problem
This is a similar issue to https://github.com/elk-zone/elk/pull/3085

When a post has more characters than the server limit, the "Publish" button is disabled and not clickable. But users can still try publishing the post by typing <kbd>Ctrl</kbd> + <kbd>Enter</kbd>, which always results in the error.

screenshot of compose widget with the error message "Publish failed 1. Validation failed: Text character limit of 500 exceeded"

## Expected
Elk should prevent sending a new post with too many characters when typing the keyboard shortcut and no error should be shown.

ippatev:
fix #3107

### DIFF
--- a/components/publish/PublishWidget.vue
+++ b/components/publish/PublishWidget.vue
@@ -196,6 +196,9 @@ async function toggleSensitive() {
 }
 
 async function publish() {
+  if (isPublishDisabled.value || isExceedingCharacterLimit.value)
+    return
+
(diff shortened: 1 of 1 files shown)


shuuji3 (approved):
Thanks! LGTM

PR #3145, opened 2025-01-20
feat(i18n): Update it-IT locale

katullo11:
Corrected one string and used more standard terminology for the "boost" action.


shuuji3:
Hi @katullo11, thanks for the update! About replacing "Boost" with "Reblog", as we still use "Boost" in other languages, I'd like to avoid changes only in Italy translation. Other fixes look good to me. Could you only include non-reblog updates in this PR for now?

I guess Elk sticks to the word "Boost" because the official Mastodon still uses that word in their UI and keeps Elk aligned with them.

katullo11:
> I don't know the past discussion but I guess Elk sticks to the word "Boost" because the official Mastodon still uses that word in their UI, and to keep Elk aligned with them.

I totally hear you. I had actually translated "Boost" more faithfully up until now, but I've recently noticed that the official stance for the Italian Mastodon community (i.e. the wording you will find on the default Mastodon website) is to translate that as "reblog". I'm gonna be honest, I hate that translation, but as a translator I can't ignore the user's expectations when it comes to standard terminology. At the end of the day it's your call: if you are fine with having a "less standard" but more "bespoke" wording I'll keep using the original translation.

shuuji3:
Ah, sorry, I didn't know such Mastodon translation specific to the Italian community, and yes, certainly [https://mastodon.social?lang=it](https://mastodon.social?lang=it) uses the term "Reblog". In that case, I also think your approach is right, it's less confusing when new users migrate from the official app to Elk. I'm going to merge as is. Thanks for explanation 🙏🏻

shuuji3 (approved):

PR #3153, opened 2025-01-24
fix: prevent infinite account hover card

### LINKED ISSUE #2986: recursive profile pic popup
Sorry, I don't really know how to put it better in the title :)

See attached picture. Basically, if you hover over someone's profile pic to get a profile preview, that is a very helpful feature! But if you hover over the profile pic on the popup, it opens a new profile preview, on which you can hover over the picture, which in turn opens a new profile preview and so on and so forth. 

Image

Windows 10
Firefox 130.0.1
Elk 0.15.1

### LINKED ISSUE #3148: Bug: infinite loop for profile hover card
When hovering a profile picture or username on any feed, Elk displays a card to summarize the account's informations.

When hovering the profile picture (not the username) **_on that card_** another card appears... And so on^^

Image

Only one card should appear.

shuuji3:
fix #2986 (and fix #3148)


shuuji3:
I confirmed that this won't prevent the existing hovercard, so it should be fine. I'm merging this to include in the next release as I saw similar reports mentioning the elk account a few times.

PR #3159, opened 2025-01-27
fix: account profile page title

Cl0v1s:
When visiting someone's account page, the page's title shows "Posts" (here "messages" in french) 
image

Using `nav.profile`  instead of  `account.posts` to show "Profile"


shuuji3 (approved):
LGTM :+1:

PR #3165, opened 2025-01-30
fix: account avatar should not be streched

Cl0v1s:
Account avatar are stretched instead of being centered and cropped both on profile page and avatar in status 

Before
image

After
image

### DIFF
--- a/components/account/AccountAvatar.vue
+++ b/components/account/AccountAvatar.vue
@@ -24,7 +24,7 @@ const accountAvatarSrc = computed(() => {
     :src="(error || !loaded) ? 'data:image/gif;base64,R0lGODlhAQABAIAAAAAAAP///yH5BAEAAAAALAAAAAABAAEAAAIBRAA7' : accountAvatarSrc"
     :alt="$t('account.avatar_description', [props.account.username])"
     loading="lazy"
-    class="account-avatar"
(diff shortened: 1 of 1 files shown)


shuuji3:
@Cl0v1s I think your fix looks correct and I can merge this. But when I tried to replicate the issue like "before" screenshot, I couldn't find the same issue with the same account on my side.

For example, this preview (from PR 3164) doesn't have this fix but it shows the profile icon correctly: https://deploy-preview-3164--elk-zone.netlify.app/universeodon.com/@Soket@bdx.town

I also checked main.elk.zone but the same result: https://main.elk.zone/m.webtoo.ls/@Soket@bdx.town

Do you know any other profile with the same issue I can check?

shuuji3:
It seems that Mastodon server crops and caches the original profile image on their server, and the Mastodon API returns that cache data to Elk if the account is logged in to the Mastodon server.

- Original image: https://media.bdx.town/media/761a924ef64b90bf215932ac634b37749ce6ce76e61a9a733978c40f608e69e5.jpg
- Image from Mastodon API: https://media.universeodon.com/cache/accounts/avatars/113/865/648/347/107/400/original/85f4c968fb80648e.jpg

shuuji3 (approved):

Issue #3150, opened 2025-01-23
Support for push notifications in GotoSocial 0.18


VyrCossont:
I'm the GotoSocial developer implementing Web Push notifications for the upcoming GtS 0.18 release and wanted to give you a heads up. Right now Elk properly displays that GtS doesn't support them yet:

![Image](https://github.com/user-attachments/assets/f3b9981f-c504-4a4a-b561-272f283300a8)

But we'll be adding them with superseriousbusiness/gotosocial#3587 fairly soon. They work almost identically to [Mastodon's](https://docs.joinmastodon.org/methods/push/), with the following caveats:
- We have a few additional notification types for GotoSocial [interaction policies](https://docs.gotosocial.org/en/latest/user_guide/settings/#default-interaction-policies): `pending.favourite`, `pending.reply`, and `pending.reblog`
- `policy` for controlling who you get PNs from is not _yet_ implemented: you can set it to anything but GtS will ignore it and always return `policy: "all"` (this may change before we ship 0.18)

Please let me know if you have questions.

shuuji3:
Thanks for sharing the update. 😃 

I'm still not familiar with push notification-related codes but checked some of the related codes.

## Message
In the current implementation, the second paragraph (`It seems that your server does not...`) is shown when `currentUser` object has `vapidKey` value. If the new version has the same property, it will be hidden. Otherwise, we'll need to adjust the feature detection.

ref. 
- https://github.com/elk-zone/elk/blob/main/components/notification/NotificationEnablePushNotification.client.vue#L18
- https://github.com/elk-zone/elk/blob/main/components/notification/NotificationEnablePushNotification.client.vue#L59-L61

## New notification type

> We have a few additional notification types for GotoSocial [interaction policies](https://docs.gotosocial.org/en/latest/user_guide/settings/#default-interaction-policies): pending.favourite, pending.reply, and pending.reblog

Current Canary Elk (https://main.elk.zone) won't show notifications with unknown types and just print out warnings to the console: https://github.com/elk-zone/elk/blob/main/components/notification/NotificationCard.vue#L34
Once Goto Social updates, we could add those types with new strings if needed.

VyrCossont:
Do you happen to know where the VAPID key is coming from?

VAPID public keys are per _instance_, but Mastodon used to also send them when app credentials were created; this was [deprecated in Mastodon 4.3.0](https://docs.joinmastodon.org/entities/Application/#vapid_key) and may stop working someday.

[...]

PR #3193, opened 2025-02-07
feat: adopt to `/api/v2/instance` to fetch app `vapid_key`

shuuji3:
Currently, the vapid key in AppInfo (needed for push notification) is retrieved from the response of `POST /api/v1/apps` and stored permanently to the server storage, when someone first tries to log in to the server from the Elk. But it was deprecated since Mastodon v4.3.0 (ref. https://docs.joinmastodon.org/methods/apps/#create). This change uses the recommended `GET /api/v2/instance` value.

This also updates the key of the storage for AppInfo (`servers:v3:${server}:` -> `servers:v4:${server}:`) because some (non-Mastodon) servers may return different values with this change. As a side effect, this will partially solve the existing lock-out issue (ref. #3150).

Mastodon server returns the exact same vapid key value from `POST /api/v1/apps` and `GET /api/v2/instance`, so there should not be any difference for Mastodon.

### DIFF
--- a/server/cache-driver.ts
+++ b/server/cache-driver.ts
@@ -16,10 +16,10 @@ export default defineDriver((driver: Driver = memory()) => {
 
       return driver.hasItem(key, {})
     },
-    async setItem(key: string, value: any) {
+    async setItem(key: string, value: any, opts: any = {}) {
       await Promise.all([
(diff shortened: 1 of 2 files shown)


shuuji3:
I also added `ttl` option, which is used for some storage drivers.

ref. #3193

PR #3196, opened 2025-02-08
feat: add option to use explore page for search

### LINKED ISSUE #2102: Allow users to add Explore to Search view
I see that a recent PR (https://github.com/elk-zone/elk/pull/2075) has removed the Explore functionality from the Search view. I'm requesting an option in the settings to revert that change, so that this decision can be determined by the user.

Whilst I understand the reasoning given (mainly expanded on here: https://github.com/elk-zone/elk/pull/1301#issuecomment-1416490216) this was/is a core feature of Mastodon. Explore is the primary method on the platform for keeping up with breaking news, engaging in community discussions, and facilitating content discovery. This is particularly the case if you're on a mid-to-large-sized instance, where the Local and Federated timelines are often too busy to provide meaningful value.

Locking that functionality away behind a multi-step menu – particularly on mobile – is problematic.

I see that there is already ongoing discussion around making the mobile navigation (and perhaps all Homepage navigation) more customisable in the future (https://github.com/elk-zone/elk/issues/641), but as this is a much larger change (with little interaction since January), I'd suggest a toggle would be a more immediate, beneficial solution.

This can be defaulted to "off" to preserve the original concerns around doomscrolling and cluttering up the Search interface, and would make sense to nest within the Wellbeing section of the Preferences panel. But for those of us who actively use, and benefit from, this feature, simply removing it feels like a regressive step to take.

shuuji3:
resolve #2102

New preference option (enabled in this screenshot but disabled by default)

Screenshot of preference page. new item "Use Explore page for search" exists with checked

Screenshot of Explore page. search input is shown the above of header tab

Screenshot of Explore page searching "#css" shows hashtags and accounts

### DIFF
--- a/composables/settings/definition.ts
+++ b/composables/settings/definition.ts
@@ -27,6 +27,7 @@ export interface PreferencesSettings {
   enableDataSaving: boolean
   enablePinchToZoom: boolean
   useStarFavoriteIcon: boolean
+  useExplorePageForSearch: boolean
   zenMode: boolean
   experimentalVirtualScroller: boolean
   experimentalGitHubCards: boolean
@@ -83,6 +84,7 @@ export const DEFAULT__PREFERENCES_SETTINGS: PreferencesSettings = {
   enableDataSaving: false,
   enablePinchToZoom: false,
   useStarFavoriteIcon: false,
+  useExplorePageForSearch: false,
   zenMode: false,
   experimentalVirtualScroller: true,
   experimentalGitHubCards: true,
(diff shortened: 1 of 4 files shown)


shuuji3:
I'm not sure if this option should be either in "Preference" or "Interface". So just used Preference which is easier to implement. If it should be in "Interface", I feel like other options like "Use star icon" should also be "Interface". 🤔

patak-dev:
I think "Preference" is good for this one 👍🏼

ayo-run:
I think this is a good change. I'm happy to approve if you can resolve the conflicts @shuuji3 -- also a question for later, should we hide the explore nav option if this is enabled?

PR #3213, opened 2025-02-25
feat: always force login to Akkoma or Pleroma servers

### LINKED ISSUE #934: Login doesn't work with Pleroma
I can't login into my Pleroma instance just after Elk was updated to `v0.3.0`. It was working fine in earlier versions. The backend is Pleroma version `2.4.55-921-g20790c1d-develop`.

While being redirected back to `elk.zone` for the OAuth callback (URL `https://elk.zone/signin/callback?server=gabi.is&token=...&vapid_key=...`), this is what the Console is showing:

image

shuuji3:
fix #934

WIP: not tested and `isAkkoma` and `isPleroma` may not work before login


Cl0v1s:
Thanks for proposing this change !

PR #3217, opened 2025-02-27
refactor: remove `withDefaults` macro and clean up reactive props destructuring

shuuji3:
I saw the inconsistent usages of withDefaults in code base. So I did global search and made its usages as consistent as possible.

- Removed `withDefaults` macro and instead use default assignment in the object destructuring
- Removed some unused props assignment in the object restructuring when only used in `<template>` section
- other clean up for consist usage of `defineProps` macro

### DIFF
--- a/components/account/AccountAvatar.vue
+++ b/components/account/AccountAvatar.vue
@@ -1,7 +1,7 @@
 <script setup lang="ts">
 import type { mastodon } from 'masto'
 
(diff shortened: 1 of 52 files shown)


patak-dev (approved):
looks great! 💚

shuuji3:
Thanks! 😀

PR #3218, opened 2025-02-27
feat: add new preference to unmute videos by default

### LINKED ISSUE #3216: Add preference toggle for whether audio and video should be muted by default or not
Similar to the "Enable Autoplay" toggle in Settings > Preferences, it would be nice to have a toggle for whether audio and video should be muted by default. I have autoplay off and find it a bit cumbersome having to press both play _and_ unmute on everything I want to watch/listen to.

Having a default mute state that follows the autoplay preference could also work for me. I.e. autoplay on -> mute on, autoplay off -> mute off.

shuuji3:
resolve #3216

Currently, Elk always mutes all videos. This PR adds a new preference to allow users to choose to unmute videos by default.

screenshot of preference page. it has "Unmute videos by default" checkbox item

### DIFF
--- a/components/status/StatusAttachment.vue
+++ b/components/status/StatusAttachment.vue
@@ -71,6 +71,7 @@ const isVideo = computed(() => attachment.type === 'video')
 const isGif = computed(() => attachment.type === 'gifv')
 
 const enableAutoplay = usePreferences('enableAutoplay')
+const unmuteVideos = usePreferences('unmuteVideos')
 
 useIntersectionObserver(video, (entries) => {
   const ready = video.value?.dataset.ready === 'true'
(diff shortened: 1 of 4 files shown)


patak-dev (approved):

patak-dev:
The way these preferences currently work is that they are all off by default and then you opt-in into them. I also thought it was weird to have "unmute" at first but I think it is worth to have consistency here. Instead of "Unmute" it could also be called "Sound on"

decibyte:
All good! It wasn't just a detail that slipped trough. Thanks again :)

shuuji3:
Ah, yes, this was intended behavior as Elk focuses on a calmer social media experience. That's why Elk has Zen mode🙂 

But I think my English word choice was not so good. How about "Video sound on by default" instead?

Mock UI:

screenshot of preferences page with item: "Video sound on by default

decibyte:
Absolutely works for me :+1:

shuuji3:
This label was updated by #3225! 😀

PR #3227, opened 2025-03-05
chore: bump node version

iiio2:
Since node version 22 is stable, i think we may use node 22 in .nvmrc.


shuuji3:
Yes, let's version up now 👍🏻

iiio2:
Thanks @shuuji3

PR #3228, opened 2025-03-06
ci: refer `.npmrc` to ensure using consistent node version

shuuji3:
follow up of #3227

- This makes sure the same version of Node.js in Workflows
- Update docs


shuuji3:
I don't see no error on CI build etc.

PR #3247, opened 2025-03-25
chore: use `navigator.userAgent` instead of `navigator.platform`

iiio2:
`navigator.platform` is deprecated. (https://developer.mozilla.org/en-US/docs/Web/API/Navigator/platform)


shuuji3 (approved):
Thanks! I also confirmed this works as previously on macOS.

iiio2:
@shuuji3 Thank you. 😊

PR #3249, opened 2025-03-26
feat: Update nl-NL.json uppercase key ALT

haayman:
modify key status.img_alt.alt to status.img_alt.ALT

### DIFF
--- a/locales/nl-NL.json
+++ b/locales/nl-NL.json
@@ -617,7 +617,7 @@
     "filter_show_anyway": "Laat toch zien",
     "gif": "GIF",
     "img_alt": {
-      "alt": "ALT",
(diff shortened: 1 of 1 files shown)


shuuji3 (approved):
Good catch!👍🏻

shuuji3 (approved):

haayman:
> Good catch!👍🏻

well, when the key is staring you in the face on every image it's really not that hard. The fact that English no longer is the fallback language makes it pretty bad for any language that has low coverage in the translations

shuuji3:
Sorry, I'm looking into a solution for this regression, but haven't found one yet. I appreciate your improvement in translation in any case.

PR #3265, opened 2025-04-14
fix: Right-click causes image drag on ModalMediaPreviewCarousel

### LINKED ISSUE #3264: Right-click on image preview starts dragging image
## Description

Every time I right click the image on a post from inside the image preview (where the image is enlargened), the my cursor starts trying to drag the image. I'm not holding down any mouse buttons and my cursor is dragging the image.

## Expected Result

When I right-click on an image in the image preview it should not start dragging the image.

## Video

https://github.com/user-attachments/assets/0b649c43-0a17-4961-83e9-9505cc5a1e80

## Solution

I'm guessing this is related to the image preview being a slider component where you can drag images to move from one slide to the next. This might be fixable just by ignoring right-click events.

Christopher-Hayes:
## Description

When a drag on the image preview (the "modal media preview carousel") begins, check if the mouse right-click button is down. If it is down, prevent the carousel drag behavior.

This is to fix the issue where you right-click on an image to show the browser right-click menu, but the carousel thinks you're now dragging the carousel.

## Videos

The video below first shows the bug on the current stable build. And then it shows the behavior on the dev build with the new code. The image is no longer dragged on right-click.

https://github.com/user-attachments/assets/5d3716a4-4eff-4685-b22c-64422fa7ca22

- [X] Tested
- [X] Linted

Fixes #3264 

## Misc Notes

Vertical dragging should probably be supressed, as well as dragging when there is only a single media component. But, to keep this PR focused, I did not try to fix those UX quirks.

### DIFF
--- a/components/modal/ModalMediaPreviewCarousel.vue
+++ b/components/modal/ModalMediaPreviewCarousel.vue
@@ -32,6 +32,7 @@ const y = ref(0)
 
 const isDragging = ref(false)
 const isPinching = ref(false)
+const isRightClicking = ref(false)
 
 const maxZoomOut = ref(1)
 const isZoomedIn = computed(() => scale.value > 1)
@@ -78,6 +79,10 @@ useGesture({
       goToFocusedSlide()
   },
   onDrag({ movement, delta, pinching, tap, last, swipe, event, xy }) {
+    // Ignore right-click events
+    if (isRightClicking.value)
+      return
+
     event.preventDefault()
 
     if (pinching)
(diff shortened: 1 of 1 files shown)


Christopher-Hayes:
@howwohmm Thanks for the feedback. I pushed a new commit to just use left-click, added some flexibility to avoid breaking mobile.

**Double-set:** I don't see this creating new bugs. Ultimately, `onDrag()` is repeatedly resetting `isDragging` while dragging, and that's pre-existing functionality. So, if render performance is the concern, this PR wouldn't have much impact in comparison.

I could remove the set from the `onDrag()` since it shouldn't be necessary anymore, but I'm hesitant to remove code under the assumption it's no longer needed.

https://github.com/user-attachments/assets/77511098-4c44-4903-8770-e7a0e6da0e99

PR #3271, opened 2025-04-25
feat: set media player preference as enabled by default

ayoayco:
Proposal to promote the media player as not experimental 😄 

This PR will
- set the preference to show embedded media player as true by default
- promote embedded media player as not experimental

### DIFF
--- a/components/status/StatusContent.vue
+++ b/components/status/StatusContent.vue
@@ -29,7 +29,7 @@ const hideAllMedia = computed(
     return currentUser.value ? (getHideMediaByDefault(currentUser.value.account) && (!!status.mediaAttachments.length || !!status.card?.html)) : false
   },
 )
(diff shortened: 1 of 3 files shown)


shuuji3:
Since setting this as default means we think it's stable and recommend using it, I think we could let this one time change and merge it as is? (For me, I've been using this nice player for a while and haven't seen any issues.)

Some people might have tried it and turned it off, but I think there are many more people who haven't even tried it yet. Keeping the existing value will let those people miss this useful feature after that.

shuuji3:
The current default value is `false` (I think it was automatically stored after update) so we cannot distinguish if the value `false` means that it was turned off manually.

ayo-run:
Thanks @shuuji3 for checking! I agree that this feature is now stable and should be a default. I need to re-check the settings stored, but overall I feel it should be fine for a the users who enabled this to see some unexpected changes on their experience as that's the reason we marked this as experimental.

Issue #1586, opened 2023-02-03
Crazy List Ideas


boehs:
Follow up on #372

Ideas:

- [ ] blank space to the right of the timeline ~~and under the search box~~ could be used to display the user's pinned lists or followed hashtags. @vwchu https://github.com/elk-zone/elk/issues/372#issuecomment-1363484385
- [ ] Lists without following user @prohr https://github.com/elk-zone/elk/issues/372#issuecomment-1365982170
  - [ ] Alternatively/And/Or, fine grained filtering @sarah11918 (hi) https://github.com/elk-zone/elk/issues/372#issuecomment-1365992548
- [ ] hashtags in lists (without firehose?) https://github.com/elk-zone/elk/issues/372#issuecomment-1381336026 (no privacy concern here, unlike users without following) (mentioned in @sarah11918 quote, explicitly first mentioned by @maegul
- [ ] Timeline tabs https://github.com/elk-zone/elk/issues/372#issuecomment-1387480181

prohr:
To clarify, my proposal wasn't ever to bulk unfollow everyone on a list.
That'd be icky.

In the current Mastodon model, the fact that I read posts from people I
follow via named lists is private to me.  How those lists are organized,
what I call them, how often they're used -- only my client + I know any of
that.  Nothing about lists federates.

None of that would change with this proposal.
As far as ActivityPub is concerned, I'd still continue to **follow**
everyone on these lists.  Their posts would get delivered to both:

- the inbox (my primary/main/home feed) and
- one or more lists (which only I see)

The only proposed change is that, on the client side, any posts from
accounts on filtered lists (or exclusive lists or whatever you call it)
wouldn't be shown on the home feed.  They'd only be shown on the list
(rather than in both places).

This user-friendly, privacy-protecting behavior is just a UX enhancement to
help me segregate which posts I read where.  Think of it as yet another
"don't show me this twice" feature + nothing more.

prohr:
To clarify, my proposal wasn't ever to bulk unfollow everyone on a list.  That'd be icky.

In the current Mastodon model, the fact that I read posts from people I follow via named lists is private to me.  How those lists are organized, what I call them, how often they're used -- only my client + I know any of that.  List information is stored on the server, but nothing about that feature federates.   None of that would change with this proposal.  

As far as ActivityPub is concerned, I'd still continue to **follow** everyone on these lists.  Their posts would get delivered to both: 

- the inbox (my primary/main/home feed) and
[...]

PR #3272, opened 2025-04-25
feat: search for following when viewing accounts in a list

### LINKED ISSUE #1600: feat: Pick from Follows to add to list
## Problem

Adding people to lists is tedious/non intuitive

## Proposed solution:

Somewhere (ideally inside the lists interface), allow to scroll through the people you follow, and add people to the list.

I propose adding an extra tab similar to the Accounts tab when clicking on a list, but either with all followed accounts, or the accounts not in the list:
list interface with an extra followed tab

ayoayco:
fix #1600 (included in list improvements #1586)

This implements a "Search among people you follow" field when viewing the Accounts in a list to easily add new accounts in the list being viewed.

## how it will be in elk:

https://github.com/user-attachments/assets/41afc7bb-9ded-45c7-9368-823e5b8a9c94

## how it looks in the mastodon web app:

Screenshot 2025-04-25 at 18 01 10

### DIFF
--- a/pages/[[server]]/list/[list]/index/accounts.vue
+++ b/pages/[[server]]/list/[list]/index/accounts.vue
@@ -110,7 +110,7 @@ function removeAccount(account: mastodon.v1.Account) {
     <div left-0 top-18 absolute w-full z-10 group-focus-within="pointer-events-auto visible" invisible pointer-events-none>
       <div w-full bg-base border="~ dark" rounded-3 max-h-100 overflow-auto :class="results.length === 0 ? 'py2' : null">
         <span v-if="query.trim().length === 0" block text-center text-sm text-secondary>
-          {{ $t('search.search_desc') }}
+          {{ $t('list.search_following_desc') }}
         </span>
         <template v-else-if="!loading">
           <template v-if="results.length > 0">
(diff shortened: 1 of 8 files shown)


pages/[[server]]/list/[list]/index/accounts.vue
```diff
@@ -1,16 +1,153 @@
...
+  inputRef,
+})
+
 const params = useRoute().params
 const listId = computed(() => params.list as string)
 
-const paginator = useMastoClient().v1.lists.$select(listId.value).accounts.list()
+const $mastoList = useMastoClient().v1.lists.$select(listId.value).accounts
+const paginator = $mastoList.list()
+const accountsInList = ref((await useMastoClient().v1.lists.$select(listId.value).accounts.list({ limit: 0 })).map(account => account.id))
```
Comment (ayoayco) [inline pages/[[server]]/list/[list]/index/accounts.vue:18]: There is a bug in masto.js -- setting the `limit` to 0 should return all accounts but it doesn't work atm. Opened a ticket: https://github.com/neet/masto.js/issues/1282

Comment (ayoayco) [reply pages/[[server]]/list/[list]/index/accounts.vue]: I made this limit to 1000 while the masto.js ticket is still open :)

shuuji3 (approved):
Great, this is a good addition!

PR #3273, opened 2025-04-25
feat: show followed hashtag badge

### LINKED ISSUE #2636: Add "following tag" badge on the header of the post
This is the feature proposal to add a "following tag" status header similar to the one https://phanpy.social/ has:

Screenshot from 2024-02-28 23-39-31

The boost status header code could be reused for this.

Screenshot from 2024-02-29 00-08-09

Occasionally, I'm still a bit confused when I see a post from a not-following user in the home timeline.

To implement this, I assume there are at least two options:
1. If the masto.js's status object has metadata that tells us that the status is included in the home timeline because it includes one of followed tags, then we can use that information to add a tag header.
2. If not, maybe we need to fetch and store the current user's followed tags beforehand and check if one of the tags in each post is included in that list.

related issue
- #1833 (request to highlight the hashtag link in a post)
- #1832 (request to show indicator from following users to distinguish from posts from following tag)

ayoayco:
fix #2636 

This will show a followed tag badge on a status

## Example (#privacy is followed)
Screenshot 2025-04-25 at 22 10 24

### DIFF
--- a/components/status/StatusCard.vue
+++ b/components/status/StatusCard.vue
@@ -61,6 +61,10 @@ const collapseRebloggedBy = computed(() => rebloggedBy.value?.id === status.valu
 const isDM = computed(() => status.value.visibility === 'direct')
 const isPinned = computed(() => status.value.pinned)
 
+// limit 0 should get all w/o pagination, but might need checking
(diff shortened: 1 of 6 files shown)


ayoayco:
Thank you @shuuji3 🧡

PR #3288, opened 2025-05-05
[WIP] fix: temporary mitigation for locale number format issue

### LINKED ISSUE #3241: statistics missing
I might be mistaken, but it appears that the number of likes, boosts, etc. is missing here

https://main.elk.zone/todon.nl/@haayman/114198842472996320

Image

shuuji3:
fix #3241

My findings so far:
- In [`LocalizedNumber.vue`](https://github.com/elk-zone/elk/blob/60b1d0224c2b2d371c6280cc2d477e669944e852/components/common/LocalizedNumber.vue#L13-L15), `count` has the right number but `formatNumber(count)` and `formatHumanReadableNumber(count)` returns `''` causing no number presented.
- Custom formats like `smallCounting` are defined in [`config/i18n.ts`](https://github.com/elk-zone/elk/blob/55ad2438ed7703331d38b26888e128ee824bd6e6/config/i18n.ts#L326-L343). I guess those are not used properly now. This might be caused by `vue-i18n`'s upgrade (now v10.0.5) and the possible `n()` function's behavior change. But I don't understand how `vue-i18n` works yet.

### DIFF
--- a/composables/i18n.ts
+++ b/composables/i18n.ts
@@ -22,9 +22,18 @@ export function useHumanReadableNumber() {
   }
 
   return {
-    formatHumanReadableNumber: (num: MaybeRef<number>) => fn(unref(num)),
-    formatNumber: (num: MaybeRef<number>) => n(unref(num), 'smallCounting', locale.value),
-    formatPercentage: (num: MaybeRef<number>) => n(unref(num), 'percentage', locale.value),
+    formatHumanReadableNumber: (num: MaybeRef<number>) => {
+      return String(unref(num))
+      return fn(unref(num))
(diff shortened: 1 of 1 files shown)


shuuji3:
Maybe we can merge this mitigation first so that main.elk.zone will get usable again. Though a large number is not looking so good with this: 
screenshot of at mastodon account with "840204 followers", not "840k followers"

shuuji3:
#3298 solved the issue completely, so this PR is no longer needed.

I guess there might be some incompatibility between the newer i18n library and non-v4 compatible Nuxt setup.

PR #3295, opened 2025-05-18
chore: use `vite-pwa/nuxt`

iiio2:
Actually it was said in the comment, "@vite-pwa/nuxt' once released and remove pwa module" in nuxt config.


userquin:
we need some extra work to allow the migration since we're using i18n to build the webmanifest  files... we cannot just remove the files in this PR

PR #3299, opened 2025-05-20
feat: pass origin as website when loging into server

nexus-uw:
- helps users know which elk server(s) they have active login sessions with
- currently all login sessions are 'elk.zone'

# testing
- npm run dev
-  logged into my gotosocial server, inspected tokens under /settings/user/tokens, confirmed "App website" pointed to my local elk dev instance


shuuji3 (approved):
LGTM! I confirmed this change use the correct origin value via `/api/login`. (Sorry, it took a while to review 🙏🏻 )

PR #3314, opened 2025-06-16
feat: add support for the Translation API

tomayac:
Currently, translations happen on the server by making a `POST` request to `https://translate.elk.zone/translate` as can be seen in Elk's production version and by inspecting the **Network** tab, filtering for `"translate"`:

Screenshot 2025-06-16 at 19 18 22

This PR adds support for the experimental [Translation API](https://developer.chrome.com/docs/ai/translator-api) in Chrome, which consists of the Language Detector and the Translator API. These two APIa are currently in origin trials ([Translator API](https://developer.chrome.com/origintrials/#/view_trial/4445615782168100865), [Language Detector API](https://developer.chrome.com/origintrials/#/view_trial/662592095176884225)), so they can be tested already by real users.

Screenshot 2025-06-16 at 19 20 02

The advantage over the current approach is reduced server costs for Elk, and full offline support, which, for example, would allow for using this API to translate DMs in a privacy-preserving way.

### DIFF
--- a/app/components/status/StatusBody.vue
+++ b/app/components/status/StatusBody.vue
@@ -11,7 +11,7 @@ const {
   withAction?: boolean
 }>()
 
-const { translation } = useTranslation(status, getLanguageCode())
+const { translation } = await useTranslation(status, getLanguageCode())
 
 const emojisObject = useEmojisFallback(() => status.emojis)
 const vnode = computed(() => {
--- a/app/components/status/StatusTranslation.vue
+++ b/app/components/status/StatusTranslation.vue
@@ -9,7 +9,7 @@ const {
   toggle: _toggleTranslation,
   translation,
   enabled: isTranslationEnabled,
-} = useTranslation(status, getLanguageCode())
+} = await useTranslation(status, getLanguageCode())
 const preferenceHideTranslation = usePreferences('hideTranslation')
 
 const showButton = computed(() =>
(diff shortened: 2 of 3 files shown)


tomayac:
Not sure why TypeScript is angry, there's [`Status.language`](https://docs.joinmastodon.org/entities/Status/#language).

danielroe:
Yes, the issue is that it doesn't exist on `StatusEdit`. We likely need to check if `'language' in status` ...

tomayac:
> Yes, the issue is that it doesn't exist on `StatusEdit`. We likely need to check if `'language' in status` ...

Alright, I'll add this check.

tomayac:
Wait, TypeScript is happy now, but the `shouldTranslate` logic is messed up.

tomayac:
Alright, I think TypeScript is happy now, and the logic seems to be working as well. Here's a [test link to a toot](https://deploy-preview-3314--elk-zone.netlify.app/universeodon.com/@dominikg@webtoo.ls/114658482222285696) (German, translate to English), or just take `/universeodon.com/@dominikg@webtoo.ls/114658482222285696` to test on `localhost`. Try it on regular Chrome and Chrome Dev (or Canary) with the `chrome://flags/#translation-api` flag set to `"Enabled without language pack limit"`. Once this is merged, you can join the origin trials linked in https://github.com/elk-zone/elk/pull/3314#issue-3150739347 and get this out to real users.

danielroe:
currently trying to figure out why LanguageDetector is not available on Chrome 137... (I have restarted the browser.)

CleanShot 2025-06-17 at 15 26 15@2x

The [article you wrote here](https://developer.chrome.com/docs/ai/language-detection) seems to indicate it might require Chrome Beta (138+)?

tomayac:
Duh, sorry. I forgot to mention the `chrome://flags/#language-detection-api` flag. You also need to set that.

danielroe (approved):

PR #3316, opened 2025-06-20
feat: add support for the Language Detector API

tomayac:
This PR adds support for automatically detecting the composition language based on the [Language Detector API](https://developer.chrome.com/docs/ai/language-detection). The detected language is updated as the user types, and the more text (I went with a minimum number of 5 graphemes), the more reliable the detection. This should greatly improve the lives of people who compose toots in different languages and always forget to update the language picker. On browsers that don't support the API, simply nothing happens.

Screenshot 2025-06-20 at 16 31 15
Screenshot 2025-06-20 at 16 33 31
Screenshot 2025-06-20 at 16 33 46
Screenshot 2025-06-20 at 16 34 20

PR #3317, opened 2025-06-21
feat: do not use language detection features on low end devices

Cl0v1s:
Also add debounce and aborting to prevent spamming the API / non-conherent UI changes

### DIFF
--- a/app/components/publish/PublishWidget.vue
+++ b/app/components/publish/PublishWidget.vue
@@ -237,9 +237,13 @@ function stopQuestionMarkPropagation(e: KeyboardEvent) {
     e.stopImmediatePropagation()
 }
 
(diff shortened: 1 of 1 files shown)


tomayac (commented):
This (debouncing) is a good addition that I should have thought of in the context of https://github.com/elk-zone/elk/pull/3316. Thanks for the refinement of the implementation! 

Non-owner LGTM

shuuji3:
@Cl0v1s @tomayac Thanks for the PR & review! 🙂 

Once we're confident there's no need for additional changes, we could probably release a new stable version for broader delivery (since it's a while ago when I relased the previous version).

Cl0v1s:
Thanks for the merge ! 🔥

PR #3325, opened 2025-07-07
fix(ui): increase target size of back button

### LINKED ISSUE #1836: Back button is too small to be usable (18x18)
The back button used in detail views was surprisingly hard to click. I ended up missing it half the time. Turns out it's small: 18x18 pixels.

Screenshot of a toot view with an element overlay from Chrome DevTools showing that the back arrow button is a link element with a 18x18 size

Recommendations for bigger target areas include:

[WCAG 2.1 Success Criterion 2.5.5 Target Size (AAA)](https://www.w3.org/WAI/WCAG21/Understanding/target-size.html):

> The size of the target for pointer inputs is at least 44 by 44 CSS pixels except when: [series of possible exceptions]

[Material Design 2](https://m2.material.io/design/usability/accessibility.html#layout-and-typography):

> ### Touch targets
>
> Touch targets are the parts of the screen that respond to user input. They extend beyond the visual bounds of an element. For example, an icon may appear to be 24 x 24 dp, but the padding surrounding it comprises the full 48 x 48 dp touch target.
>
> For most platforms, consider making touch targets at least 48 x 48 dp. […]
>
> ### Pointer targets
>
> Pointer targets are similar to touch targets, but apply to the use of motion-tracking pointer devices such as a mouse or a stylus. Consider making pointer targets at least 44 x 44 dp.

[Apple Human Interface Guidelines](https://developer.apple.com/design/human-interface-guidelines/components/menus-and-actions/buttons#best-practices):

> Make buttons easy for people to choose. On a touchscreen, buttons need a hit target of at least 44x44 points to accommodate a fingertip. On all screens, it’s essential to include enough space around a button so that people can visually distinguish it from surrounding components and content, whether people use touch, a pointer, or a system that expands a button when it’s in focus.

If Elk tries to be a touch-compatible UI, it should probably try to achieve touch targets of at least 44x44 CSS pixels. Currently the main sidebar buttons are a bit shy of that at 42.5x37.5, which is not ideal but might still work.

But buttons that are below ~30px of size, and especially as small as 18x18, are a clear UX issue and lead to a lot of missed clicks & taps.

fvsch:
Fixes #1836, which has links to the relevant WCAG success criterion, as well as Material Design guidelines and Apple HIG.

This small UI problem has been bothering me as an Elk user for a couple years. I still routinely miss clicks on the header's back button because its target area is so small. So here's a fix at last ^^

This PR:

- Increases the size of the sidebar (large screens) back button's target area from 30x22.5 to 45x45.
- Increases the size of the header (small & medium screens) back button's target area from 18x18 to 42.75x42.75.
- Sets a minimum height on the header, so that its height does not depend only on a combination of nested padding values and the presence or absence or specific buttons or content elements.
- Changes the buttons from `<a>` without a `href` (the result of using `<NuxtLink>` without a `to` or `href` prop) to `<button>` instead, so that the buttons are recognized as, well, buttons, and can be focused. Anchors without a `href` are not considered interactive elements and cannot receive focus.

Elements in Elk's UI are sized using increments of `0.25rem` as well as text sizes also expressed in fractions of `rem`, which can result in fractional pixel values like 42.75x42.75. I’ve kept this style instead of trying to hit specific sizes like 44x44 (Apple HIG) or 48x48 (Material Design). The only exception is the header's min-height, set to 53px (52px for content and 1px for the border).

Here are before-and-after comparisons, with the last image in each group showing the button's clickable area in purple.

elk-back-button-small-screen

elk-back-button-large-screen

### DIFF
--- a/app/components/main/MainContent.vue
+++ b/app/components/main/MainContent.vue
@@ -35,15 +35,16 @@ const containerClass = computed(() => {
         'backdrop-blur': !getPreferences(userSettings, 'optimizeForLowPerformanceDevice'),
       }"
     >
-      <div flex justify-between px5 py2 :class="{ 'xl:hidden': $route.name !== 'tag' }" class="native:xl:flex" border="b base">
-        <div flex gap-3 items-center :overflow-hidden="!noOverflowHidden ? '' : false" py2 w-full>
-          <NuxtLink
-            v-if="backOnSmallScreen || back" flex="~ gap1" items-center btn-text p-0 xl:hidden
+      <div flex justify-between gap-2 min-h-53px px5 py1 :class="{ 'xl:hidden': $route.name !== 'tag' }" class="native:xl:flex" border="b base">
+        <div flex gap-2 items-center :overflow-hidden="!noOverflowHidden ? '' : false" w-full>
+          <button
+            v-if="backOnSmallScreen || back"
+            btn-text flex items-center ms="-3" p-3 xl:hidden
             :aria-label="$t('nav.back')"
             @click="$router.go(-1)"
           >
-            <div i-ri:arrow-left-line class="rtl-flip" />
-          </NuxtLink>
+            <div text-lg i-ri:arrow-left-line class="rtl-flip" />
+          </button>
           <div :truncate="!noOverflowHidden ? '' : false" flex w-full data-tauri-drag-region class="native-mac:justify-start native-mac:text-center">
             <slot name="title" />
           </div>
--- a/app/components/nav/NavTitle.vue
+++ b/app/components/nav/NavTitle.vue
@@ -33,17 +33,16 @@ router.afterEach(() => {
         {{ $t('app_name') }} <sup text-sm italic mt-1>{{ env === 'release' ? 'alpha' : env }}</sup>
       </div>
     </NuxtLink>
-    <div
-      hidden xl:flex items-center me-8 mt-2 gap-1
-    >
-      <CommonTooltip :content="$t('nav.back')">
-        <NuxtLink
+    <div hidden xl:flex items-center me-6 mt-2 gap-1>
+      <CommonTooltip :content="$t('nav.back')" :distance="0">
+        <button
+          type="button"
           :aria-label="$t('nav.back')"
-          :class="{ 'pointer-events-none op0': !back || back === '/', 'xl:flex': $route.name !== 'tag' }"
+          btn-text p-3 :class="{ 'pointer-events-none op0': !back || back === '/', 'xl:flex': $route.name !== 'tag' }"
           @click="$router.go(-1)"
         >
-          <div text-xl i-ri:arrow-left-line class="rtl-flip" btn-text />
-        </NuxtLink>
+          <div text-xl i-ri:arrow-left-line class="rtl-flip" />
+        </button>
       </CommonTooltip>
     </div>
   </div>


app/components/main/MainContent.vue
```diff
@@ -35,15 +35,16 @@ const containerClass = computed(() => {
...
+        <div flex gap-2 items-center :overflow-hidden="!noOverflowHidden ? '' : false" w-full>
+          <button
+            v-if="backOnSmallScreen || back"
+            btn-text flex items-center ms="-3" p-3 xl:hidden
             :aria-label="$t('nav.back')"
             @click="$router.go(-1)"
           >
-            <div i-ri:arrow-left-line class="rtl-flip" />
-          </NuxtLink>
+            <div text-lg i-ri:arrow-left-line class="rtl-flip" />
```
fvsch, line 46: Using `text-lg` increases the size of the icon slightly, which makes the back button a bit more visible and the icon size more consistent with the icons shown alongside titles.

Note that the other instance of this back button (in the sidebar in `app/components/nav/NavTitle.vue`) uses the next size, `text-xl`.

app/components/main/MainContent.vue
```diff
@@ -35,15 +35,16 @@ const containerClass = computed(() => {
...
     >
-      <div flex justify-between px5 py2 :class="{ 'xl:hidden': $route.name !== 'tag' }" class="native:xl:flex" border="b base">
-        <div flex gap-3 items-center :overflow-hidden="!noOverflowHidden ? '' : false" py2 w-full>
-          <NuxtLink
-            v-if="backOnSmallScreen || back" flex="~ gap1" items-center btn-text p-0 xl:hidden
+      <div flex justify-between gap-2 min-h-53px px5 py1 :class="{ 'xl:hidden': $route.name !== 'tag' }" class="native:xl:flex" border="b base">
+        <div flex gap-2 items-center :overflow-hidden="!noOverflowHidden ? '' : false" w-full>
+          <button
+            v-if="backOnSmallScreen || back"
+            btn-text flex items-center ms="-3" p-3 xl:hidden
```
fvsch, line 42: When placing buttons with a large target area (e.g. 44x44 or 48x48) in a header or toolbar, one common concern is how to handle spacing and especially padding in said header or toolbar.

If we use some padding on the parent container itself, we ensure that content (e.g. a title) is nicely spaced and does not touch the edges of the container. But then bigger elements like buttons may not have enough room, or the padding of the container and the padding of the button are added and we get too much space.

Another option is to avoid padding on the container, and then making sure that each child comes with its own padding: buttons or other clickable elements should have generous padding, but plain text elements like titles need similar padding too.

In this case I didn't want to overhaul the styling of many elements, so I opted for:

- Setting a min-height on the container, so that it appears to have a stable height whether its content is some large buttons, small buttons, or a title.
- Removing most of the vertical padding on the container (keeping just one step of padding, i.e. `0.25rem`, as a safety net).
- Keeping most of the horizontal padding on the container, and using a negative margin (`ms="-3"`) on the button equal to its horizontal padding to visually offset the button.

So on the vertical axis, height is primarily managed by a min-height and elements (big or small) are vertically centered using flexbox. And on the horizontal axis, spacing is primarily managed by padding on the container, and we use a negative margin to offset one button.

I’m happy to change this approach for something different, but this one looked like the least invasive change I could make.

Issue #3266, opened 2025-04-15
Asterisk pairs are removed in code block


sieste:
This code block

```
1 * 2 * 3
```

gets converted to

```
1 2 3
```

nove-b:
I’ve submitted a pull request to fix this issue: #3326 
Any feedback would be greatly appreciated. Don't hesitate to let me know if any adjustments are required.

shuuji3:
fixed by #3326 ✔️

PR #3326, opened 2025-07-08
fix(ui): Asterisk pairs are removed in code block

nove-b:
### Fix for #3266

Fix: Asterisk Rendering in Code Blocks
I've fixed the problem where pairs of asterisks were disappearing inside code blocks. My solution involves converting these asterisk pairs into HTML entities, which makes sure they now render correctly.

image

### DIFF
--- a/app/composables/content-parse.ts
+++ b/app/composables/content-parse.ts
@@ -104,11 +104,12 @@ export function parseMastodonHTML(
           .replace(/</g, '&lt;')
           .replace(/>/g, '&gt;')
           .replace(/`/g, '&#96;')
+          .replace(/\*/g, '&ast;')
         const classes = lang ? ` class="language-${lang}"` : ''
         return `><pre><code${classes}>${code}</code></pre>`
       })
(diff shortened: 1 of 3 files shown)


shuuji3:
For your information, this is just a GitHub tips but if you follow the specific format like "close #3266" (without header), the GitHub links the PR to the original issue properly. Then, it automatically closes the original issue when we merge the PR. This is useful to manage Issues, not just for this project 🙂 

ref. Linking a pull request to an issue - GitHub Docs - https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue

nove-b:
@shuuji3 
Thanks for the tip!
I didn't know the format needs to be in plain text without a header—I'll make sure to follow that next time.
Really helpful for keeping issue management smooth! 😊
Also, thanks for adding the test cases as well — much appreciated!

shuuji3:
You're welcome! 😀

PR #3347, opened 2025-08-13
feat: add new well-being feature - disable timeline autoloading 

### LINKED ISSUE #3291: Wellbeing option to add "Load More" button or turn off infinite scrolling
### Problem
Infinite scrolling continuously loads new posts as the user scrolls down. Because there are no interruptions or cues to stop, the user is at risk of falling into a state where they're not conscious of their own scrolling and will not stop until they're interrupted or they get tired. Infinite scrolling may contribute to rising rates of internet addiction and its related symptoms of depression and anxiety.

### Proposed solution
Create an option in Preferences > Wellbeing that adds a "Load more" button to feeds.

When the "Load more" option is active, feed posts are loaded in batches and the next batch is hidden. When the user scrolls to the bottom of the first batch, they must click "Load more" to unhide them and continue reading until they reach the bottom of the next batch, and so on...

Or, the option could copy Tumblr's wording and call it "Endless scrolling," and deactivating this mode would add the "Load more" button.

The purpose of the "Load more" button is to create a "natural stopping cue" so that the user can assess whether they want to continue scrolling or switch to another task.

### Alternatives
The user could use browser extensions that limit their access to certain websites after a period of time. In my personal experience, this does not work. Often I'll be in the middle of writing something when I get the notice that time's almost up. This is annoying and makes me feel like I can't control my own time, even if I had set it up myself, so I usually end up disabling it.

An option to add something that interrupts the flow and implicitly reminds you of what you're doing is better for retaining (and regaining) a sense of control. I think one button and one click would be enough to briefly shake someone out of zombie scrolling. 

### More context
Elk has a Grayscale mode in Wellbeing options and this discussion about it caught my eye (https://github.com/elk-zone/elk/issues/1122) and actually inspired me to make this request. An option for a "Load more" button would fulfill the same purpose: to inhibit excessive use of social media. I figure that implementation may be more involved but I strongly feel it's still worth considering adding this feature to Elk's Wellbeing options!

**Further reading**
Ethics of the Attention Economy: The Problem of Social Media Addiction. DOI: [https://doi.org/10.1017/beq.2020.32](https://doi.org/10.1017/beq.2020.32)
I’m so sorry, says inventor of endless online scrolling. [The Times (2019)](https://www.thetimes.com/business-money/technology/article/i-m-so-sorry-says-inventor-of-endless-online-scrolling-9lrv59mdk)

I also made a mockup of what the button could look like. I hope this is helpful in any way. Thank you so much for reading and for y'all's consideration.

Image

shuuji3:
resolve #3291

Clicking the button loads the next single page.

Screenshot from 2025-08-14 00-23-44

### Remaining tasks
- [x] Add a new preference in the well-being section, as written in the original issue #3291
- [x] Allow showing the first page without clicking it
- [x] Fix issue where each load showing weird spacing between the button and the end edge of the timeline.

### DIFF
--- a/app/components/common/CommonPaginator.vue
+++ b/app/components/common/CommonPaginator.vue
@@ -126,7 +126,7 @@ defineExpose({ createEntry, removeEntry, updateEntry })
       {{ t('common.error') }}: {{ error }}
     </div>
     <button
-      v-if="state !== 'loading'"
+      v-if="state !== 'loading' && state !== 'done'"
       flex="~ gap-1 center" w-full my-6 py-6
       btn-text rounded-lg bg="base"
       filter-saturate-0 hover:filter-saturate-100


shuuji3:
Refined the button style, which would be less attractive for users to click the button.

screenshot of 'load more' button at the bottom of timeline

shuuji3:
The loading button is still shown even after finished the loading 😅

Screenshot of the end of timeline. there is still loading button

haayman:
I'm seeing the 'load more posts' link even though I've not enabled this option

image

image

shuuji3:
Hi, @haayman #3540 should fix the issue. Can you try updateing to the latest version (https://github.com/elk-zone/elk/commit/7fd8b4d78d829305a5c97b352a8b85630d52de1b)? Thanks always for reporting issues 🙏🏻

haayman:
> Hi, @haayman #3540 should fix the issue. Can you try updateing to the latest version ([7fd8b4d](https://github.com/elk-zone/elk/commit/7fd8b4d78d829305a5c97b352a8b85630d52de1b))? Thanks always for reporting issues 🙏🏻

yes, it's fixed. Of course you've got it all mixed up about who should be grateful to whom :-) 
Thanks always for fixing issues  🙏🏻

PR #3355, opened 2025-08-20
docs: change docker compose command

iiio2:
From July 2023, Compose v1 stopped receiving updates. It’s also no longer available in new releases of Docker Desktop.

https://docs.docker.com/compose/releases/migrate/#how-do-i-switch-to-compose-v2


shuuji3 (approved):
LGTM :+1:

iiio2:
Thanks @shuuji3 . :blush:

Issue #3385, opened 2025-09-22
Canary build won't load anymore


haayman:
Since today the canary build won't load anymore. This happens both at home as at work. 
I'm only seeing the elk logo on an otherwise blank page.

quite a few errors in the network tab.

```js
Failed to find a valid digest in the 'integrity' attribute for resource 'https://main.elk.zone/_nuxt/B2740BIo.js' with computed SHA-384 integrity 'SmVtc0vhqCnVCKh18WtOkrYpJItWXItyxh4wgV4YZePJfHapPxmjdYJ3sxbekWpm'. The resource has been blocked.Understand this error
home:1 Failed to find a valid digest in the 'integrity' attribute for resource 'https://main.elk.zone/_nuxt/BPdISNAt.js' with computed SHA-384 integrity 'kLy5VmQBScBjJYks6qWRvp44INDwISK2OEyhexib7MSOYGbfi+rCByKtx53IVUYA'. The resource has been blocked.Understand this error
[...]

PR #3387, opened 2025-09-26
chore: temporarily disable `entryImportMap` feature to avoid hash mismatch issue

shuuji3:
related: #3385


shuuji3:
In the PR preview, I cannot find any difference from #3381 (https://deploy-preview-3381--elk-zone.netlify.app/ is still has no issue before merging to `main` and being accessed at https://main.elk.zone). Can we test this on `main` branch?

PR #3406, opened 2025-10-28
chore: bump node version to lts in `.nvmrc`

iiio2:
i think node v24 is not lts now. that is why i have changed it.


iiio2:
Thanks @danielroe .

PR #3420, opened 2025-11-17
docs: remove `@antfu/ni` example

iiio2:
Since we have removed `@antfu/ni` (https://github.com/elk-zone/elk/pull/3417), maybe we can also remove from the readme file.


danielroe (commented):
it's still useful - ni is something that should be installed globally - and you then use it to install deps.

these commands would not work if you didn't have it installed globally, even if it were a dependency of elk

iiio2:
Ok. Thanks.

PR #3434, opened 2025-11-23
feat: check user session before subscribing to streaming

ayo-run:
This will check if there is a user session before creating a streaming client to make sure there is a valid token for subscribing to the instance's web sockets endpoint.

### Inspecting mastodon web behavior

Mastodon does not allow streaming on the client when no user is logged in, which can be seen, for example, by inspecting the dev tools Network tab -> WS on this page: https://m.webtoo.ls/public/local

...then comparing when there is a logged in user (top window) and not logged in (bottom window):

Screenshot from 2025-11-23 11-40-42

### Preventing errors for elk.zone

This will prevent the following errors when opening https://elk.zone without a user session

Screenshot from 2025-11-23 11-42-29
Screenshot from 2025-11-23 11-42-52

### DIFF
--- a/app/composables/masto/masto.ts
+++ b/app/composables/masto/masto.ts
@@ -25,7 +25,7 @@ export function mastoLogin(masto: ElkMasto, user: Pick<UserLogin, 'server' | 'to
   const url = `https://${server}`
   const instance: ElkInstance = reactive(getInstanceCache(server) || { uri: server, accountDomain: server })
   const accessToken = user.token
-  const streamingApiUrl = instance?.configuration?.urls?.streaming
+
   const createStreamingClient = (streamingApiUrl: string | undefined) => {
     // Only create the streaming client when there is a user session
     return streamingApiUrl && currentUser.value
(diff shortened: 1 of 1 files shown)


app/composables/masto/masto.ts
```diff
@@ -25,14 +25,19 @@ export function mastoLogin(masto: ElkMasto, user: Pick<UserLogin, 'server' | 'to
...
   masto.client.value = createRestAPIClient({ url, accessToken })
-  masto.streamingClient.value = createStreamingClient(streamingApiUrl)
+
+  if (currentUser.value !== undefined) {
+    createStreamingClient = (streamingApiUrl: string | undefined) => {
+      return streamingApiUrl ? createStreamingAPIClient({ streamingApiUrl, accessToken, implementation: globalThis.WebSocket }) : undefined
+    }
+
+    masto.streamingClient.value = createStreamingClient(streamingApiUrl)
+  }
```
ayoayco: I added a check for `currentUser` before creating the streaming client to prevent the errors we get when elk.zone tries to subscribe to the web socket without a valid token.

PR #3447, opened 2025-11-26
chore: update masto.js to latest v7.6.0

shuuji3:
ref. https://github.com/neet/masto.js/releases/tag/v7.6.0


shuuji3:
@SocketSecurity ignore npm/safer-buffer@2.1.2

This is false positive alert.

PR #3433, opened 2025-11-22
fix: paginator values being cloned

ayo-run:
fixes #3431 
Prevent paginator values being cloned

Please test the deployed preview: https://deploy-preview-3433--elk-zone.netlify.app/home

## recording
[Screencast from 11-22-2025 06:53:30 PM.webm](https://github.com/user-attachments/assets/540ca28d-e78a-4eb1-b7ad-d1925c703f17)

PR #3429, opened 2025-11-21
fix: new toots doesn't load anymore

ayo-run:
This should upgrade masto.js to the latest version, and fix some breaking changes.

fixes #3301

PR #3462, opened 2025-12-08
chore: pin vue scroller to 2.0.0 beta 3

### LINKED ISSUE #3457: All thread replies clustered at bottom of thread; difficult to see structure of conversation
All threads so far have this behavior: replies to any post in the thread appear at the very bottom of the thread. It is possible, with careful scrolling and good eyesight, to see which post a reply refers to, but it is time-consuming and frustrating. I am thinking the intended behavior would be to have each reply appear right below the post to which it replies, as in other threading apps/pages. If this is a feature instead of a bug, I will change this to a feature request (or make a new feature request).

Tested on Firefox 145.0.2, Linux Mint.

userquin:
@darrinlrogers can you test with the pr preview here

Looks like beta 5 has wrong logic: https://github.com/Akryum/vue-virtual-scroller/issues/795

We should add vue scroller to ignoreDeps at renovate.json file until fixed

resolves #3457

### DIFF
--- a/package.json
+++ b/package.json
@@ -123,7 +123,7 @@
     "vue": "^3.5.4",
     "vue-advanced-cropper": "^2.8.9",
     "vue-i18n": "^11.1.12",
-    "vue-virtual-scroller": "2.0.0-beta.4",
+    "vue-virtual-scroller": "2.0.0-beta.3",
     "workbox-build": "^7.1.1",
     "workbox-cacheable-response": "^7.1.0",
     "workbox-expiration": "^7.1.0",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -285,8 +285,8 @@ importers:
         specifier: ^11.1.12
         version: 11.1.12(vue@3.5.22(typescript@5.9.2))
       vue-virtual-scroller:
-        specifier: 2.0.0-beta.4
-        version: 2.0.0-beta.4(vue@3.5.22(typescript@5.9.2))
+        specifier: 2.0.0-beta.3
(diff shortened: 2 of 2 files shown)


userquin:
weird :thinking:, on pr preview the jump still present, maybe we need to pin virtual scroller to previous beta version 3 (only when virtual scrolling enabled => by default it is enabled); no idea why on my local working (maybe I need to dedupe dependencies on my local)

shuuji3:
I see. Then, let's try the beta 3 😅 I recently updated only masto.js version but I'm still not sure it's related to this issue.

userquin:
argggggg, it is even wrose, the issue maybe not in the scroller, this is weird, I will try to check it again later on my local deduping deps, no idea what's wrong (the deploy failed arggggg) :unamused:

shuuji3:
Oh, sorry for that 😅 I'm going to investigate the issue too.

There was the major version up of masto.js (ref. https://github.com/elk-zone/elk/pull/3429) and a bit different empty timeline issue (ref. https://github.com/elk-zone/elk/pull/3433) recently. Those might be related to this.

shuuji3:
I tried a few times but not yet found how to fix this 😓 The current `vue-virtual-scroller` has not been published for a few years and maybe it's not compatible with something in newer versions of vue/nuxt.

I think I can instead try [virtua](https://github.com/inokawa/virtua/), which is used by npmx to implement infinit virtual scroller in the search result page. It looks very good and supports wide varieties of features including dynamic sizing:

> It also handles common hard things in the real world (dynamic size measurement, scroll position adjustment while reverse scrolling and imperative scrolling, iOS support, etc).

PR #3465, opened 2025-12-11
feat: scale up emoji image on hover

### LINKED ISSUE #3452: Allow to zoom in custom emoji image
Some custom emojis have small details that are difficult to see at the default zoom level. Some kind of zoom feature will make the details more visible.

shuuji3:
resolve #3452

Another approch is only scale up when an emoji is clicked/tapped but it seems to require a more complex change on the logic of the emoji component construction.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -8,25 +8,34 @@ on:
   pull_request:
     branches:
       - main
+  workflow_dispatch:
+
 jobs:
-  docker:
-    runs-on: ubuntu-latest
+  build:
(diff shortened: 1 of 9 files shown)


danielroe:
this looks good - do we need to handle the case of emojis in user names as well?


https://github.com/user-attachments/assets/344503e5-7ee4-4573-9818-472a52013df3

shuuji3:
Yeah, I think it's better to just zoom in on the post body. I'll limit the scope of the effect.

danielroe (approved):

Issue #3471, opened 2025-12-20
`Unknown file extension ".ts"` on build


blacklight:
This seems like a regression somewhere: on `docker compose build`, the prepare step is now failing with:

```
12.51 . prepare$ node scripts/prepare-translation-status.ts && node scripts/prepare.ts
12.55 . prepare: node:internal/modules/esm/get_format:219
12.55 . prepare:   throw new ERR_UNKNOWN_FILE_EXTENSION(ext, filepath);
12.55 . prepare:         ^
12.55 . prepare: TypeError [ERR_UNKNOWN_FILE_EXTENSION]: Unknown file extension ".ts" for /elk/scripts/prepare-translation-status.ts
12.55 . prepare:     at Object.getFileProtocolModuleFormat [as file:] (node:internal/modules/esm/get_format:219:9)
12.55 . prepare:     at defaultGetFormat (node:internal/modules/esm/get_format:245:36)
12.55 . prepare:     at defaultLoad (node:internal/modules/esm/load:120:22)
12.55 . prepare:     at async ModuleLoader.loadAndTranslate (node:internal/modules/esm/loader:483:32)
12.55 . prepare:     at async ModuleJob._link (node:internal/modules/esm/module_job:115:19) {
12.55 . prepare:   code: 'ERR_UNKNOWN_FILE_EXTENSION'
12.55 . prepare: }
12.55 . prepare: Node.js v22.13.0
12.55 . prepare: Failed
```

shuuji3:
@blacklight Hi, probably that error is due to the older version of Node.js. v22.18.0+ can execute `.ts` file natively (ref. Node.js — Running TypeScript Natively - https://nodejs.org/en/learn/typescript/run-natively).

The error shows the v22.13.0 but the latest node container (Elk's base container uses `node:lts-alpine`) should be v24.12.0 at this moment. 

```shell
> head Dockerfile 
FROM docker.io/library/node:lts-alpine AS base

# Prepare work directory
WORKDIR /elk

FROM base AS builder

> docker run --rm -it node:lts-alpine
Welcome to Node.js v24.12.0.
Type ".help" for more information.
```

Could you try clear out the local image cache and rebuild again?

PR #3472, opened 2025-12-20
Fix Typescript build error

blacklight:
Context: #3471


shuuji3:
Thank you for the PR! We'd like to avoid adding extra dependencies just for handling .ts file execution. Please refer to my comment on the original issue. That should fix the issue.

PR #2643, opened 2024-02-29
feat: add schedule post button to publish widget

shuuji3:
resolve #511

This is still WIP PR.

PR #3476, opened 2025-12-23
feat: add schedule posts page

shuuji3:
follow up for #2643

screenshot of new 'scheduled post' page. sidebar icon is under compose icon


shuuji3:
Now it shows the scheduled posts!

Screenshot of scheduled posts page

Screenshot of scheduled posts page with another account

PR #3479, opened 2025-12-25
feat(i18n): Update pt-PT translation

emanuelpina:
Remove 1 key, change another and Add and Translate 24 new keys.


shuuji3 (approved):
Thanks! 🙂

Issue #673, opened 2022-12-31
Option to hide replies and boosts from the Timeline


draga79:
On Mastodon, you can choose to hide replies from the timeline and it also works with apps or other web clients. On Pleroma and forks (Akkoma, etc) this is strictly controlled by the frontend. Pleroma-FE, Soapbox (and its fork Mangane) allow it, elk doesn't seem to support it at the moment. 
This could make it useful for all the people who just do not want to read replies on Pleroma and/or people who do not want to change the global Mastodon option but have a less "noisy" timeline on mobile.

edimitchel:
I don't if it's accurate, but the zen-mode is hiding action button.
Your issue is applicable also for normal mode?

aadilayub:
+1! Really need this feature for when the timeline feels overwhelming

windhamdavid:
just wanted to show my support for this feature 👊🏼

juandarr:
Would love for this feature to be implemented. Sometimes I just want to check out what the people I follow is up to without all the noise "replies" and "boosts" brings.

skullbunnygalaxy:
I would love to have this feature as well.

phocks:
+1 for this feature. It's the only thing keeping me going back to native Mastodon.

ayoayco:
Hi all, the Mastodon Settings PUT endpoint allows sending a "shows" prop in the "home" setting in the payload. I don't see it documented though. 
![Screenshot 2023-08-12 at 12 44 49](https://github.com/elk-zone/elk/assets/4262489/23015907-72ec-4a4a-abb0-9d23219d358c)

ayoayco:
Hey guys sorry I haven't been able to move this forward. I still have this on my list to get back to, but am removing my name on this just so someone else can pick it up if i don't get around to it soon enough

Karmalakas:
Any chance this is happening?

PR #3482, opened 2025-12-28
feat(timeline): add filters to hide replies and boosts

draga79:
Add user preference settings to hide replies and boosts from home and public timelines. Self-replies are preserved when reply filter is enabled. Settings are persisted in user preferences under the Wellbeing section.

This closes the issue I opened in 2022 - #673

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -23,7 +23,7 @@ jobs:
       packages: write
     steps:
       - name: Checkout
-        uses: actions/checkout@v4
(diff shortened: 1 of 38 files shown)


shuuji3:
OK, I managed to fix the test error. Thank you for implementing this! 😀

shuuji3 (approved):

shuuji3:
It would be nice if we can have more granular filterings like #1931. Some user may want to have per timeline configuration (for example, some want to show see replies in their custom list) but we can enhance those later.

PR #3504, opened 2026-01-08
fix: handle p.quote-inline class in content parsing

shuuji3:
Mastodon prepends the "RE: https://..." string in quote post for the compatibility purpose. This PR adjust the existing parser to allow `p.quote-inline` element and hides it.

## Before
Screenshot of quote post, showing a text 'RE: https://...' by mastodon's compat purpose

## After
Screenshot of quote post without RE text

### DIFF
--- a/app/components/status/StatusBody.vue
+++ b/app/components/status/StatusBody.vue
@@ -55,4 +55,8 @@ const vnode = computed(() => {
 .status-body.with-action p {
   cursor: pointer;
 }
+
+.status-body .quote-inline {
+  display: none;
+}
 </style>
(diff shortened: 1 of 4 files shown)


shuuji3:
confirmed the fix!

cheeaun:
I think it should only be hidden *if* there's a native quote.

shuuji3:
Good point.😅  My assumption was `class="quote-inline"` is only added to `<p>RE: ...</p>` element, just based on my observation that only Matsodon v4.5 add `class="quote-inline"` but Mastodon v4.4 doesn't (not sure if there's documentation). But I think checking that can prevent hiding it unintentionally.

PR #3520, opened 2026-01-23
feat: Add extra confirmation step to "block domain"

Jonty:
Blocking a domain is a highly destructive action that results in all follows and followers on a domain being permanently lost.

The current block domain dialog is almost identical to the block user dialog, and the buttons to select each function are right next to each other making it quite easy to accidentally block a domain when you intended to block a user.

This implements a github-repo-deletion style "type in the name of the server to confirm" step to domain blocking, as well as calling out that this action is permanent.


shuuji3:
Thanks for implementing this!🙂 I think github-repo-deletion style is good fit for this, given this is very destructive operation when there's mistake.

PR #3553, opened 2026-02-19
fix: switch color mode icons

ttntm:
This PR switches the color mode icons.

When Elk opens in dark mode, it will now show the "sun" icon that represents light mode, instead of showing the "moon" icon that represents dark mode, which is already active.

### DIFF
--- a/app/components/nav/NavFooter.vue
+++ b/app/components/nav/NavFooter.vue
@@ -17,7 +17,7 @@ function toggleDark() {
   <footer p4 text-sm text-secondary-light flex="~ col">
     <div flex="~ gap2" items-center mb4>
       <CommonTooltip :content="$t('nav.toggle_theme')">
-        <button flex i-ri:sun-line dark-i-ri:moon-line text-lg :aria-label="$t('nav.toggle_theme')" @click="toggleDark()" />
+        <button flex dark-i-ri:sun-line i-ri:moon-line text-lg :aria-label="$t('nav.toggle_theme')" @click="toggleDark()" />
       </CommonTooltip>
       <CommonTooltip :content="$t('nav.zen_mode')">
         <button


shuuji3:
Thanks for the PR. But this would change the current semantics of the state showing the selected color mode. I think the root cause is that the app failed to load the initialize the correct color mode in some case. Could you adjust the PR so that it can detect the dark theme correctly?

ttntm:
Thanks for having a look at this, @shuuji3.

`useColorMode()` is a composable provided by the package `@nuxtjs/color-mode`. As such, we rely on it to function as documented, which means that a failure "to initialize the correct color mode in some case" would mean opening a bug ticket in the package's repo, I think.

But this PR is about something else: as you wrote, it "would change the current semantics of the state showing the selected color mode". The PR aims to change these exact semantics, because - from an UX perspective - they are confusing at best, if not simply wrong.

Theme switching uses a `<button>` element, which should be labelled in a way that clearly indicates what pressing it is doing. Using said button to show the _currently selected_ color mode is the exact opposite of that.

Take a simple form, for example - it loads, then gets rendered, and can be filled. The button used to submit the form is usually labelled in a way that indicates its functionality, i.e. "Submit form". We'd all agree that labeling this button "Form loaded" (the empty form's current state), would be wrong, I'd assume. The same principles and UX patterns should apply here, IMO.

shuuji3:
Thanks for the explanation.🙂 I think I now understand the difference. I was thinking the button expresses the current state (the sun icon represents that the current theme is the light mode and it's fair given this is a `<button>`) but if it is seen as the action to change as you explained (the same sun icon now means that when users click the button, it will switch to the light mode!), it makes sense. I guess the issue is the button could be recognized either way. Maybe we could change the UI like toggle switch, but it'd be better to keep the current simple button. I think I'm ok now, just need to adjust my understanding of the UI.

ttntm:
Thanks for thinking about it again :)

A tiny change, but it might make a difference.

PR #3557, opened 2026-02-23
fix: border typo

MatthiasPortzel:
The border on quote retweets is defaulting to white instead of base. The relevant elements have `b="base 1"`, but the correct Tailwind syntax for border seems to be `border="base 1"`. This is relevant/noticeable in dark mode.

Before (first QRT in my timeline of course, post isn't relevant):
Screenshot 2026-02-23 at 8 26 33 AM

After:
Screenshot 2026-02-23 at 8 29 46 AM

### DIFF
--- a/app/components/publish/PublishWidget.vue
+++ b/app/components/publish/PublishWidget.vue
@@ -587,17 +587,17 @@ const detectLanguage = useDebounceFn(async () => {
                 {{ $t('action.remove_quote') }}
               </button>
             </div>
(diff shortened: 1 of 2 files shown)


shuuji3 (approved):
Thank you for taking time to create a PR :pray: That seems to be my mistake and probably I was confused with other shortcut notation.

(By the way, we are actually using [UnoCSS](https://unocss.dev) built by Anthony on Elk, which was inspired by Tailwind CSS and has quote similar namings 🙂)

PR #3575, opened 2026-03-22
chore: replace deprecated `process.test` with `import.meta.test`

iiio2:
`process.test` is showing as deprecated.


iiio2:
Thanks.

PR #3576, opened 2026-03-22
fix(ui): deduplicate profile link icons

### LINKED ISSUE #1957: Duplicate features profile link icons

howwohmm:
Fixes #1957

Noticed this on a profile with both a "Website" and "Portfolio" field — they both map to the same link icon so you end up with two identical icons in a row. Same thing happens with "GPG" / "OpenPGP" (both key icons).

Fix: deduplicate by icon name before rendering. First field wins, any subsequent fields with the same icon still show up in the named section so the values aren't lost.

Tested on profiles with duplicate icon mappings and on profiles with no duplicates (no change).

### DIFF
--- a/app/components/account/AccountHeader.vue
+++ b/app/components/account/AccountHeader.vue
@@ -65,13 +65,17 @@ async function toggleNotifications() {
 watchEffect(() => {
   const named: mastodon.v1.AccountField[] = []
   const icons: mastodon.v1.AccountField[] = []
+  const seenIcons = new Set<string>()
 
   account.fields?.forEach((field) => {
     const icon = getAccountFieldIcon(field.name)
-    if (icon)
+    if (icon && !seenIcons.has(icon)) {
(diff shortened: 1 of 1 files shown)


howwohmm:
gentle bump — let me know if there's anything I should change or if this needs a different approach.

ayo-run:
I think the problem is when a users choose an icon via the UI (e.g., click the website icon) but means to add the other label for it (e.g, wants the "portfolio" label), which will show up in other client apps.

The current solution removes the subsequent labels in this flow, and keeps only the first.

I'm thinking if it is possible to somehow separate the flow for choosing the icon & label. For example, I choose the chain icon and it shows a dropdown to select possible labels (e.g., 'website' or 'portfolio).

ayo-run (commented):
Thanks for the contribution! Would you be willing to explore another approach as mentioned above?

PR #3577, opened 2026-03-22
chore: use `page-lifecycle` as dev deps

iiio2:


iiio2:
Thanks @shuuji3 . :blush:

PR #3582, opened 2026-03-24
fix: apply client-side content filters for case-insensitive matching

### LINKED ISSUE #3523: Post filtering is not case insensitive
Problem: the elk web client, is not honoring [mastodon user filtering rules](https://docs.joinmastodon.org/user/moderating/) (which are case insensitive)

eg. Does not match "Trump", when rule is case insensitive for "trump' 

Filtering rule:

Image

Post on elk:

Image

Post on native mastodon server:

Image

howwohmm:
Fixes #3523

## Problem

Elk relies solely on the server-populated `status.filtered` field for content filtering. The Mastodon streaming API does not include filter results on statuses, so posts received via real-time streaming bypass all user-defined content filters. This means a filter rule for "trump" won't match "Trump" in streamed posts — or any streamed post at all — because the `filtered` field is simply absent.

## Fix

- Added `app/composables/masto/filters.ts` — fetches the user's v2 filters on login and caches them
- `applyClientFilters()` matches status content against filter keywords using case-insensitive regex, respecting `wholeWord` boundaries per [Mastodon implementation guidelines](https://docs.joinmastodon.org/api/guidelines/#filters)
- Only applies to statuses that don't already have server-provided `filtered` results (so REST API responses are unaffected)
- Integrated into timeline preprocessing (`filterAndReorderTimeline`), notification filtering, and conversation filtering

## What it doesn't change

- No changes to the filter UI or settings
- Statuses with existing server-side filter results are left as-is
- No new dependencies

### DIFF
--- a/app/components/conversation/ConversationPaginator.vue
+++ b/app/components/conversation/ConversationPaginator.vue
@@ -7,9 +7,15 @@ defineProps<{
 
 function preprocess(items: mastodon.v1.Conversation[]): mastodon.v1.Conversation[] {
   const isAuthored = (conversation: mastodon.v1.Conversation) => conversation.lastStatus ? conversation.lastStatus.account.id === currentUser.value?.account.id : false
-  return items.filter(item => isAuthored(item) || !item.lastStatus?.filtered?.find(
-    filter => filter.filter.filterAction === 'hide' && filter.filter.context.includes('thread'),
-  ))
+  return items
+    .map((item) => {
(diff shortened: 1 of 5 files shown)


ayo-run:
> The Mastodon streaming API does not include filter results on statuses

I'm surprised this is the case, I didn't know they don't apply filters to streamed posts. This seems an important fix to me. I will just check the implementation & also compare with other clients.

PR #3601, opened 2026-04-14
fix: prevent empty home timeline in qoto-mastodon server

### LINKED ISSUE #1472: No support for qoto.org
I am unable to log on using elk zone 

Steps to reproduce :
Connect with an account from qoto.org

image

shuuji3:
fix #1472

The `followed_tags` endpoint is not supported by qoto instance, causing the crash before rendering the home timeline.

This fix ensure rendering home timeline when the followed tags requests fails.


ayo-run (approved):
Looks good, thanks!

shuuji3:
Thank you! 🙂

PR #3609, opened 2026-04-28
fix: convert ContentRich to a plain Vue SFC

danielroe:
currently throwing an error when upgrading to vite 8

cc: @sxzz in case it's an issue to be fixed in vue-macros


sxzz:
Hmmm, Vue Macros is not actively maintained now.

Maybe we can just remove it

PR #3643, opened 2026-06-14
chore: delete provenance workflow

shuuji3:
We've already migrated to pnpm no-downgrade policy (ref. https://github.com/elk-zone/elk/blob/main/pnpm-workspace.yaml#L5-L6) and the provenance action README explains:
> pnpm users: As of [pnpm v10.21](https://github.com/pnpm/pnpm/releases/tag/v10.21.0), pnpm now has built-in support for trustPolicy in .npmrc, which provides native enforcement of provenance checks. If you're using pnpm v10.21 or later, you may not need this action. See the [pnpm documentation](https://pnpm.io/blog/releases/10.21#trust-policy) for more details.

So I think we can remove the additional provenance workflow now. 

I still sometimes encounters a false positive version downgrade detection: https://github.com/elk-zone/elk/actions/runs/27469200692/job/81196917116?pr=3440


ayo-run (approved):
It looks good to me. Thanks!

PR #3649, opened 2026-06-16
chore: convert issue templates to yaml

### LINKED ISSUE #3491: Create GitHub Issue template
- Elk version
- Mastodon version (or other fediverse server)
- Browser version
- How to reproduce issue
- etc.

ayo-run:
This will implement github issue forms for bug reports & feature requests

Closes #3491


ayo-run:
I'm not sure if we should lessen the form fields for the bug report, but most of them are optional 😄

.github/ISSUE_TEMPLATE/bug_report.yaml
```diff
@@ -0,0 +1,90 @@
...
+  - type: input
+    id: server-version
+    attributes:
+      label: Mastodon Server version
+      placeholder: v4.6.0
+  - type: input
+    id: server-version
+    attributes:
+      label: Mastodon Server version
+      placeholder: v4.6.0
```
shuuji3:
```suggestion
```

There's a duplication.

shuuji3 (changes requested):
I think that's fine if it's optional. We can rather encourage reporters to give more information that may need to solve the issue.

shuuji3 (approved):
Thank you 🙂

shuuji3:
@ayo-run There seems to be an syntax error in the issue template 😅 Could you check the YAML syntax?

> Critical
> There is a problem with this template
> body[1]: validation is not a permitted key. [Learn more about this error.](https://docs.github.com/communities/using-templates-to-encourage-useful-issues-and-pull-requests/common-validation-errors-when-creating-issue-forms#bodyi-x-is-not-a-permitted-key)

https://github.com/elk-zone/elk/blob/ba28e3c39e48da109c29d8cf83f7831c55528ab0/.github/ISSUE_TEMPLATE/bug_report.yaml

PR #3653, opened 2026-06-17
feat: add jxl as a supported image format

lil5:
Added the jxl attachment in the list of supported images

### DIFF
--- a/app/components/status/StatusAttachment.vue
+++ b/app/components/status/StatusAttachment.vue
@@ -49,7 +49,7 @@ const objectPosition = computed(() => {
 const typeExtsMap = {
   video: ['mp4', 'webm', 'mov', 'avi', 'mkv', 'flv', 'wmv', 'mpg', 'mpeg'],
   audio: ['mp3', 'wav', 'ogg', 'flac', 'aac', 'm4a', 'wma'],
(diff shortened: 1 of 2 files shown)


shuuji3:
I'm in favor of Jpeg XL format itself, but it seems that the official Mastodon doesn't accept it yet (the latest main branch: https://github.com/mastodon/mastodon/blob/c4ea89dfd952e05da6edd0d47473fd4bd4834dae/app/models/media_attachment.rb#L53).

I wonder if this is for some compatible fediverse server. Have you already tested the upload? Could you share more information?

shuuji3:
Tested with a sample jxl file but it was failed with Mastodon server as expected.

screenshot of compose with error message: Upload failed Validation failed: File content type is invalid, File is invalid

Sample file (red trafic light, opened in image editing software that supports jxl file): https://github.com/libjxl/testdata/blob/73695d303670c90e4d506ea89d9901b081385089/jxl/blending/cropped_traffic_light.jxl

red trafic light illustration

shuuji3:
I saw Phanpy's PR https://github.com/cheeaun/phanpy/pull/1532, so I guess this is for GoToSocial support.

PR #3656, opened 2026-06-18
feat: update icons for adding / removing accounts to lists

### LINKED ISSUE #3654: Add to list and remove from list icons are hard to distinguish
The current icon the add/remove list menu on profile page ("+" and "x") are a bit too small and hard to distinguish.

screenshot of elk opening the add/remove list menu on profile page

Maybe we can replace them with simpler icons such as "+" and "✓", for example.

ayo-run:
This should make the icons for adding / removing an account to lists easier to distinguish

Closes #3654

### DIFF
--- a/app/components/list/Account.vue
+++ b/app/components/list/Account.vue
@@ -48,7 +48,7 @@ async function edit() {
           btn-action-icon
           @click="edit"
         >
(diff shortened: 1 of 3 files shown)


shuuji3 (approved):

shuuji3:
Thanks for this. I believe the new icons are much easier to understand the current state.

PR #3661, opened 2026-07-01
feat(i18n): update id-ID locale

adydetra:
Added new keywords 606/639 (94.7%) to 639/639 (100%).


shuuji3 (approved):
Thanks for the updates!

PR #3677, opened 2026-07-11
chore: welcome AI assistance, but with guidance

ayo-run:
This PR adds a few guidelines with regards to using "AI" in contributing to Elk:
- a new `AGENTS.md` to instruct AI agents of both the repository structure and our expectations from the user of the agentic tool
- a new `PULL_REQUEST_TEMPLATE.md` for a short checklist that includes our AI usage recommendations
- updates to the `CONTRIBUTING.md` for AI-assisted contributions

We have had some PRs that were obviously vibe-coded. This is not a problem in itself, and in fact some where honest and took ownership of the review themselves and therefore got in.

The intention of the updates in this PR is to make our expectations explicit: that showing ownership in your contribution makes the review process easier to trust.

### How I used AI here :)

The AGENTS.md was generated with the assistance of CLAUDE and manually updated by me to enforce so the agent reading it will remind its user of the contributor responsibility

The PR template and Contributing guides, were initially updated by CLAUDE to reflect the new additions and then manually updated by me to make them more human-readable (without repeated unnecessary details)


ayo-run:
👋 Hello @elk-zone/team -- If we agree on the changes here, I want to add agentscan workflow next

Shinigami92:
So we need a CLAUDE.md which points to AGENDS.md like in oxc repo?

Shinigami92:
I can throw my claude later onto this PR and request a review to see if it is good enough with a fresh session.

Shinigami92:
🤖 _the following comment is written fully with Claude Opus 4.8_ 🤖 

Thanks for putting this together — the intent (making contributor ownership explicit rather than banning AI) reads really well, and the "never let an LLM think/speak for you" framing from antfu's guide fits Elk nicely. I ran a fresh Claude Code review over the diff and the repo. A few things worth addressing before merge:

### 1. We need a `CLAUDE.md` pointing at `AGENTS.md`

`AGENTS.md` is the emerging cross-tool convention, but **Claude Code doesn't read `AGENTS.md` by default — it looks for `CLAUDE.md`.** So ironically the tool this file was generated with won't auto-load it. The [oxc](https://github.com/oxc-project/oxc) pattern is the clean fix — a one-line pointer that keeps `AGENTS.md` as the single source of truth:

```markdown
# CLAUDE.md

See [AGENTS.md](./AGENTS.md).
```

(A `CLAUDE.md -> AGENTS.md` symlink also works but is messier cross-platform and in Git, so the pointer file is usually preferred.)

### 2. Typo in the PR template callout

In `.github/PULL_REQUEST_TEMPLATE.md`, the `> [!IMPORTANT]` block reads:

> See **the our** guidance for [AI-assistance](...)

Doubled article — should be "See **our** guidance on [AI assistance]". It's the most-read line for the exact audience this PR targets, so worth fixing.

### 3. `AGENTS.md` documents the wrong test command

The quality-gates block says:

```bash
pnpm test:unit                # vitest run (unit + nuxt-env tests)
```

But `package.json` has `test:unit = "vitest"` (watch mode in a TTY) and a separate `test:unit:ci = "vitest run"`, which is what CI actually enforces. Since the whole point of the file is to give agents/contributors *accurate* commands, this should point at `pnpm test:unit:ci` for the "before a PR / CI enforces" gate. The same `pnpm test:unit` reference in the PR-template checklist has the same issue.

### Minor / non-blocking

- `AGENTS.md` lists the auto-import dir as `app/composables/tiptap`, but `nuxt.config.ts` registers the file `./composables/tiptap/index.ts`, not the directory.
- The by-name example `pnpm vitest -t "reorders timeline"` doesn't match any existing test — fine as illustrative syntax, but a literal run returns nothing.

Everything else I checked holds up (port `5314`, the other auto-import dirs, the `test:typecheck` vs `nuxt typecheck` distinction, the nano-staged pre-commit hook, and the anchor links all resolve correctly). 👍

<details>
<summary>🤖 Claude Code review session transcript</summary>

**User:** `/review 3677`

**Claude:** Pulled the PR context and diff via `gh pr view` / `gh pr diff`. This is a documentation-only PR (139 additions, 0 deletions) adding `AGENTS.md`, `.github/PULL_REQUEST_TEMPLATE.md`, and an "AI assistance" + "signing off" section to `CONTRIBUTING.md`. Since there's no executable code, I verified the factual claims in the docs against the actual repo rather than hunting for runtime bugs.

Verification steps run:
- `package.json` scripts → confirmed port `5314`, and found `test:unit = "vitest"` (watch) vs `test:unit:ci = "vitest run"`.
- `nuxt.config.ts` → confirmed auto-import dirs (`composables/masto`, `push-notifications`, `settings`, `tiptap/index.ts`) and the `useI18n` override.
- `.nvmrc` (`lts/*`), `simple-git-hooks`/`nano-staged` pre-commit config, existence of `tests/unit/language.test.ts`, `locales/en.json`, `config/i18n.ts`.
- Checked the PR-template anchor links (`#signing-off-your-work-recommended`, `#using-ai-assistance`) resolve, and that `../blob/main/...` relative links resolve correctly from a PR page.

Findings surfaced:
1. `.github/PULL_REQUEST_TEMPLATE.md:20` — "See the our guidance" grammar typo.
2. `AGENTS.md:28` — `pnpm test:unit` documented as `vitest run` but is actually `vitest` (watch mode); CI uses `test:unit:ci`.
3. Minor: `app/composables/tiptap` vs registered `./composables/tiptap/index.ts`; the `-t "reorders timeline"` example matches no test.

**User:** Have you read the comments in the PR as well?

**Claude:** Fetched issue comments, inline review threads, and reviews. No inline threads or submitted reviews yet. Human comments: @ayo-run wants to add an agentscan workflow next; @Shinigami92 asked whether a `CLAUDE.md` pointing to `AGENTS.md` (like the oxc repo) is needed, and offered to run Claude on the PR. Flagged the `CLAUDE.md` question as a real gap: Claude Code reads `CLAUDE.md`, not `AGENTS.md`, so without a pointer file the guidance won't auto-load.

**User:** Asked for this review to be written to a local markdown file (not pushed to the PR, since they aren't the author) for manual copy-paste, with the session transcript appended as a `<details>` block.

**User:** How would you actually rate the AGENTS.md?

**Claude:** Rated it ~8/10 — well above average for an AGENTS.md and clearly human-curated, not raw LLM output.

- **Strengths:** leads with the non-obvious (module-level refs instead of Pinia despite Pinia being installed, the `useI18n` override, SSR mocking of client-only deps, UnoCSS-not-Tailwind, "only edit `en.json`"); the RTL section is the standout with a concrete physical→logical substitution table and the `rtl-flip`/`dir="auto"` caveat; honest about footguns (`test:typecheck` vs `nuxt typecheck`, the v2→v1 instance-API fallback); good length discipline (~93 lines, doesn't re-document the framework).
- **Weaknesses:** the `pnpm test:unit` command inaccuracy is the worst category of error for a file agents trust literally (main thing keeping it from a 9); thin on a "how to self-check your work" validation loop; minor staleness (`composables/tiptap` vs registered `tiptap/index.ts`, the `-t "reorders timeline"` example matching no test); the DCO/sign-off policy rule is clever but slightly muddies an otherwise operational file.
- **Bottom line:** fix the `test:unit` command and the two staleness nits and it's a 9/10. The missing `CLAUDE.md` pointer is orthogonal to quality — the file is good, it just won't be auto-loaded by Claude Code without it.

</details>

ayo-run:
Thanks @Shinigami92 

I updated the files and confirmed the example for named vitest run :)

CLAUDE.md
shuuji3, line 1: Common pattern I saw is to create symlink to `AGENT.md`. I don't have Claude subscription so cannot test it on my side, but this may cause a delay or a bit more extra token usage by reading this file and then rereading the `AGENT.md`.

Example: https://github.com/refined-github/refined-github#:~:text=CLAUDE%2Emd

ayo-run: Made CLAUDE.md a symlink to AGENTS.md :)

ayo-run:
@shuuji3 I did the symlink, but Today I Learned ™️ that on Windows they *might\** need an elevated permission to create symlinks, so if non-admin the git checkout might not create the symlink correctly and instead contributors will see a normal text file with a single line "AGENTS.md" -- should be harmless, and I hope agents will know to follow it 😆 

\* I could not test this though... so we can keep the symlink as it is...

Other's solution is to put a oneliner: "See [AGENTS.md](#)" or the claude syntax '@' for file paths (see https://github.com/nitrojs/nitro/blob/main/CLAUDE.md)

shuuji3:
I see, today I learned! 😅 In that case, I think either option should work fine (and the root cause is that Anthropic still hasn't adopted the `AGENTS.md` filename). 

In terms of the content, how granular `AGENTS.md` needs to be would depend on the coding agents and models (which will change over time), but I think this is a good starting point. I fully agree with Anthony's rules and Daniel's blog post.

ayo-run:
I was mostly curious if the signed-off-by recommendation might turn off contributors, so I loosened it to a recommendation rather than a requirement. I think I saw that first in the Linux kernel repo as a requirement.

And yeah we took heavy inspiration from Anthony and Daniel's posts here! :)

I will do the CLAUDE.md to AGENTS.md redirection as they have in Nitro when I get to my laptop!

shuuji3:
Making it mandatory would create too many hurdles (even if we could enforce it via GitHub Actions), especially for people who aren't familiar with signing. I think just recommending it is better for now. If it helps discourage fully automated PRs or helps reflect contributors on their commits, I'm good with that.

I've seen some CNCF projects do this, probably because they need more legal protection. In that sense, it might be less necessary for Elk.

ayo-run:
Okay I'm happy with the changes here now too if we're good. I think we should get this in as a start so the AGENTS.md also evolves.

shuuji3 (approved):

PR #3678, opened 2026-07-11
feat: Update from upstream v1

maybeanerd:


maybeanerd:
ugh wrong target sorry

PR #3681, opened 2026-07-18
test: run reorder-timeline test in nuxt environment

danielroe:
spotted a [regression in nuxt's ecosystem-ci](https://github.com/nuxt/ecosystem-ci/actions/runs/29616277122/job/88007194577)

the issue is that we are opting out of nuxt's environment (=> happy-dom) but _not_ opting out of a lot of nuxt's build plugins because this is still in a nuxt runtime _project_

the other option would be to create an entirely separate vitest project that doesn't use nuxt's runtime environment at all


ayo-run (approved):
Thanks! I think let's get this in first to fix the ecosystem-ci

PR #3682, opened 2026-07-18
fix(a11y): server thumbnail alt descriptions

ayo-run:
This adds an `alt` text to the server thumbnails shown when no user is logged in. I also added a new translation string for the alt. The default English alt text, for example, will be "Thumbnail for webtoo.ls server"


shuuji3 (approved):
LGTM :+1: ♿

